PR #18, opened 2020-05-01
trigger the update of the plotly color scheme

### LINKED ISSUE #17: map color range is not updated when the text boxes are edited
Seems that changing the color range manually does not trigger an update of the plotly map

ceriottm:
Fixes #17 

Not sure this is the right way to fix it, but it works


Luthaf:
Should be live in ~10min

PR #19, opened 2020-05-01
Implement a method to specify style presets in the JSON input

ceriottm:
This should allow deciding at the dataset level how the map and the structure viewer look like. This allows e.g. to show the cell, and replicate it, or decide which variables should be displayed by default.
ATM this is just a stub.

### DIFF
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -49,6 +49,8 @@ export interface Metadata {
     references?: string[];
     /** description of the dataset */
     description?: string;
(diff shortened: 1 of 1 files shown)


rosecers:
I'd suggest these additions pre-merge (in addition to fleshing out the stub):

- specification of which parameters are available to pre-set
- example file

Luthaf:
Closing this, since there is no code in this PR. I'll implement it once #24 is merged =)

Issue #11, opened 2020-02-11
Use Plotly GroupBy instead of separate traces for symbols


Luthaf:
See: https://plot.ly/javascript/group-by/

We currently use one additional empty trace for each symbol to be displayed, this will make the code simpler and might improve performances.

Luthaf:
Related: there seems to be multiple `transform` implemented, which would also allow us to do things like #3, but they are not very well documented =/

Luthaf:
This is not very actionable, and I tried a couple of times without managing to get anything working using `groupBy`. Let's open a new issue when we have a clearer idea of what we want.

PR #25, opened 2020-05-29
Feature/pinning

rosecers:
#16 
This is a feature addition of the widget "grid" to the right side of the chemiscope, allowing for viewing of multiple widgets at once. The grid allows for anywhere from 1-12 widgets.

There are three new buttons on each widget instance, an "active" flag, a "duplicate" button, and a "close" button:

- active - sets the widget as that which will receive any data passing between the map, info, and structure. Should resize upon click. Corresponds/synchronizes to the points highlighted in the left map.
- duplicate - add a new widget to the grid with the same structure as the given widget and resize the grid.
- close - close the given widget and resize the grid.

Overview of code changes:
- overarching change: `onselect` now takes indexes and a guid, so that the map/info/grid know which guid to operate on/set active. `src/index.ts` has been adapted for proper communication
- `static/chemiscope.css` contains the css objects for all buttons
- `structure/structure.ts` (large changes): now contains a grid instance rather than a single widget. Contains all infrastructure for adding/removing widgets, including the 
    - new variables:
        - `_widgetMap` (Map): mapping from guids to widget, color, and current structure
        - `_guids` (string[], used for indexing): guids of the widgets in the grid
        - `_active` (string): guid of the active widget
    - and functions:
        - `bestArrangement(number)`: hacky function for determining the best arrangement of the widget grid
        - `get active()`: returns active guid
        - `set active()` calls _setActive
        -  `_setActive(string)`: sets active guid
        - `_resizeGridInc(number)`: resizes the grid by a given increment
        - `_setupCell(cell#, row#, column#)`: sets up grid cell
        - `_setupGrid()`: sets up the grid
        - `_removeWidget`: kills the given widget
    - and changes to existing functions

- `structure/widget.ts` (minimal changes): allows for optional passing of a guid for the widget. This allows for keeping the same size graph upon changing the dataset. Not critical, and can be changed if needed later on.
- `map/map.ts` (medium changes), including 
    - new variables:
        - `_active (string)`: guid of the active widget
        - `_markerMap (Map)`: mapping from guids to color, current structure and marker
        - `_guids (string[], used for indexing)`: guids of the widgets in the grid
    -  new functions
        - `set active()` calls _setActive
        -  `_setActive(string)`: sets active guid
        - `_removeMarker(guid)`: kills the given marker

    - and changes to existing functions
        - `constructor()`, `select()`, and `updateSelectedMarker()` take a guid to synchronize with widgets
        - all functions referring to the selected marker either loop through all markers or refer to the active marker, as appropriate

Requested Changes:
- [x] change the "+" to a "duplicate current panel" button
- [x] make the "colored dot" of non-selected structures smaller, and become large upon mouseover to hint at clickability <-- done in theory, but smaller dots on widgets are distorted for unknown reason
- [x] make sure the whole area around icons is clickable, not only the black "+" or "x" area. I think this should be easy by using a png/svg for the icon, which would also give us the possibility of choosing something that has the same look&feel of the plotly commands buttons
- [x] limit the number of panels to 9 or 12
- [x] make sure that colors are contrasting <-- in progress
- [x] add tooltips to the buttons. <-- in progress
- [x] Also, BUG: selected points are not highlighted in the 3D map

Outstanding Issues:
- [ ] JSMol throws a width error when resizing the grid, we should set a minimum width in some aspect of the grid, although the options I have tried have not worked
- [ ] Global "Add" button, which is an open discussion


src/map/map.ts
```diff
@@ -1200,53 +1258,133 @@ export class PropertiesMap {
...
-            // overflow (for points just on the border)
-            const minX = computeX(xaxis.range[0]) - 10;
-            const maxX = computeX(xaxis.range[1]) + 10;
-            const minY = computeY(yaxis.range[1]) - 10;
-            const maxY = computeY(yaxis.range[0]) + 10;
-            if (!isFinite(x) || !isFinite(y) || x < minX || x > maxX || y < minY || y > maxY) {
-                this._selectedMarker.style.display = 'none';
-            } else {
-                this._selectedMarker.style.display = 'block';
+            if (removeGUID) {
```
Luthaf, line 1374: In which case would we want to remove the marker but not the GUID?

src/static/chemiscope.css
```diff
@@ -37,14 +37,120 @@
     z-index: 10;
 }
 
-.chsp-open-viewer-settings {
+.chsp-open-viewer-settings{
```
Luthaf, line 40: that's a weird change 😄

src/static/chemiscope.css
```diff
@@ -37,14 +37,120 @@
...
+  border: solid thin black;
+  position: absolute;
+  top: 10px;
+  height: 20px;
+  width: 20px;
+  /* Place on top of jsmol applet */
+  z-index: 10;
+}
+
+.vertical-rect {
```
Luthaf, line 59: where are these used? Since CSS has no concept of scope/namespace, the class names need to be prefixed by `chsp-<use-case>-`

src/static/chemiscope.css
```diff
@@ -37,14 +37,120 @@
...
-.chsp-open-viewer-settings {
+.chsp-open-viewer-settings{
     position: absolute;
     top: 5px;
     right: 5px;
     /* Place on top of jsmol applet */
     z-index: 10;
 }
 
+.chsp-circle-button {
```
Luthaf, line 48: Is this used for the color circle on the viewer side? If so, how about naming it `.chsp-structure-selected-marker`?

src/static/chemiscope.css
```diff
@@ -37,14 +37,120 @@
...
+.vertical-rect {
+  height:20px;
+  width:10px
+}
+.horizontal-rect {
+  height:10px;
+  width:20px
+}
+
+.chsp-active {
```
Luthaf, line 68: `.chsp-active` and `.chsp-inactive` need a bit more context in their name. They are used with the viewer, right?

src/static/chemiscope.css
```diff
@@ -37,14 +37,120 @@
...
+   top: 7.5px;
+   right: 80px;
+   height:25px;
+   width:25px;
+
+   /* Place on top of jsmol applet */
+   z-index: 10;
+  }
+
+  .chsp-duplicate {
```
Luthaf, line 123: Strange indentation here

src/static/chemiscope.css
```diff
@@ -37,14 +37,120 @@
...
+  position: absolute;
+  top: 7.5px;
+  right: 50px;
+  height:25px;
+  width:25px;
+
+  /* Place on top of jsmol applet */
+  z-index: 10;
+ }
+ .chsp-add {
```
Luthaf, line 110: `.chsp-add/.chsp-close/.chsp-duplicate` need more context =)

src/static/chemiscope.css
```diff
@@ -37,14 +37,120 @@
...
+    position: absolute;
+    top: 7.5px;
+    right: 80px;
+    height:25px;
+    width:25px;
+
+    /* Place on top of jsmol applet */
+    z-index: 10;
+   }
+.chsp-widget-grid {
```
Luthaf, line 135: how about `.chsp-structure-grid`?

src/structure/widget.ts
```diff
@@ -809,6 +819,7 @@ export class JSmolWidget {
         }
     }
 
+    /// Background *atoms*
```
Luthaf, line 822: Feel free to rename the functions (since it is a private function) for something more descriptive!

src/utils/index.ts
```diff
@@ -7,6 +7,40 @@ import {makeDraggable} from './draggable';
...
+    const standardColors = ['White', 'Silver', 'Gray', 'Black',
+                            'Red', 'Maroon', 'Yellow', 'Olive',
+                            'Lime', 'Green', 'Aqua', 'Teal',
+                            'Blue', 'Navy', 'Fuchsia', 'Purple'];
+    if (i === undefined) {
+      return standardColors[Math.floor(standardColors.length * Math.random())];
+    } else {
+      return standardColors[i % standardColors.length];
+    }
+    }
```
Luthaf, line 20: strange indentation

src/utils/index.ts
```diff
@@ -7,6 +7,40 @@ import {makeDraggable} from './draggable';
...
+                         'Seashell', 'Beige', 'OldLace', 'FloralWhite',
+                         'Ivory', 'AntiqueWhite', 'Linen', 'LavenderBlush',
+                         'MistyRose'];
+    if (i === undefined) {
+      return lightColors[Math.floor(lightColors.length * Math.random())];
+    } else {
+      return lightColors[i % lightColors.length];
+    }
+
+  }
```
Luthaf, line 33: strange indentation

src/structure/widget.ts
rosecers: This was more of a note to self, because I found the name confusing. I can remove it from this PR and make a separate, small PR to change the name.

src/map/map.ts
```diff
@@ -1143,8 +1201,8 @@ export class PropertiesMap {
         this._settings.z.max.disabled = true;
 
         const symbols = this._symbols();
-        for (let i = 0; i < this._data.maxSymbols; i++) {
-            symbols.push([i]);
+        for (let sym = 0; sym < this._data.maxSymbols; sym++) {
```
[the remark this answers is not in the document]
rosecers: I received a "shadowed" variable warning and though we should be more descriptive. If that's not a problem I'm fine to revert.

src/map/map.ts
```diff
@@ -1200,53 +1258,133 @@ export class PropertiesMap {
...
 
-            this._updateSelectedMarker();
+            for (const guid of this._guids) {
+                this._updateSelectedMarker(guid);
+            }
         }
     }
+    private _setActive(guid: string) {
+
+      if (guid !== this._active) {
```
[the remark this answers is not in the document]
rosecers: This toggles the appearances of the old and new active buttons (the small circles in the upper right of the widget). Old refers to that found on the previous active widget.

src/map/map.ts
```diff
@@ -1200,53 +1258,133 @@ export class PropertiesMap {
...
+          }
+        } else {this._root.appendChild(marker); }
+      }
+      if (! this._guids.includes(guid)) {
+        this._guids.push(guid);
+      }
+      this._currentList[guid] = Math.max(0, index);
+    }
+    private _removeMarker(guid: string, removeGUID: boolean = true) {
+        const marker = document.getElementById(`chsp-selected-${guid}`);
```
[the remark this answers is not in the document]
rosecers: Is there a need (i.e. is getting a time-consuming operation)?

src/static/chemiscope.css
rosecers: Lol I'll revert. Didn't catch this one.

src/static/chemiscope.css
rosecers: These are an artifact and will be removed.

src/map/map.ts
```diff
@@ -206,7 +214,7 @@ export class PropertiesMap {
     /// Using such div is much faster than trying to restyle the full plot,
     /// especially with more than 100k points. In 3D mode, a separate trace is
     /// used instead.
-    private _selectedMarker: HTMLElement;
+    // private _selectedMarker: HTMLElement;
```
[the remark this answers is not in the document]
rosecers: This is from the old code. I'll omit and commit.

src/map/map.ts
```diff
@@ -1088,6 +1141,10 @@ export class PropertiesMap {
 
     /** Is the the current plot a 3D plot? */
     private _is3D(): boolean {
+        if (this._current === undefined) {
```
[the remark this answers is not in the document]
rosecers: It is not needed. Omitted/Committed.

src/map/map.ts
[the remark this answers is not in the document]
Luthaf: Oh, ok then. It is fine like this too 😃

src/map/map.ts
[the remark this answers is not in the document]
Luthaf: You mean the button on each JSmolWidget? Why is this done in the map code then?

src/map/map.ts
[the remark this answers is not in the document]
Luthaf: It is not so much time consuming (although it might be, I'll have to look into this), but rather making sure there is a single point of knowledge for each element in the system. In turn, I find this makes the code easier to understand and modify later (we have less places to remember to modify).

src/structure/widget.ts
Luthaf: Sounds good!

src/structure/structure.ts
```diff
@@ -267,4 +357,258 @@ export class StructureViewer {
...
+        cell = document.createElement('div');
+        cell.id = cellId;
+        cell.classList.add('chsp-widget-cell', 'grid-item');
+        cell.style.gridColumn = `${colNum}`;
+        cell.style.gridRow = `${rowNum}`;
+
+        // this was removed because it conflicted with the buttons
+        // cell.onclick = () => {this._setActive(guid)}
+
+        // use HTML5 template to generate a DOM object from an HTML string
```
Luthaf, line 480: The comment do not correspond to the code below =)

src/structure/structure.ts
```diff
@@ -267,4 +357,258 @@ export class StructureViewer {
...
+        const activeFlag = document.createElement('div');
+        activeFlag.classList.add('chsp-inactive', 'chsp-circle-button');
+        activeFlag.id = `chsp-activate-${guid}`;
+        this._colors.push(getFarColor(this._colors));
+        activeFlag.style.backgroundColor = this._colors[this._colors.length - 1];
+        activeFlag.onclick = () => { this._setActive(guid); };
+        activeFlag.innerHTML = `<div><span class="tooltiptext">Choose as active</span></div>`;
```
Luthaf, line 487: You can use the `title` attribute for simple tooltips, as used in the structure viewer settings for example

src/structure/structure.ts
```diff
@@ -267,4 +357,258 @@ export class StructureViewer {
...
+        cell.appendChild(activeFlag);
+
+        // use HTML5 template to generate a DOM object from an HTML string
+        const close = document.createElement('div');
```
Luthaf, line 491: I would use a `button` element instead of a `div`, since it already have most of the right semantics (clickable everywhere, changes pointer styling, ...)

src/structure/structure.ts
```diff
@@ -267,4 +357,258 @@ export class StructureViewer {
...
+        this._root.appendChild(cell);
+        } else {
+        cell.style.gridColumn = `${colNum}`;
+        cell.style.gridRow = `${rowNum}`;
+      }
+
+    }
+    /*
+     * Function to initialize the grid instance for nwidgets cells and place
+     * onto the DOM element mapped to `id`
```
Luthaf, line 527:
```suggestion
     * Function to initialize the grid instance for `this._nwidgets` cells and place
     * onto the DOM element mapped to `this._id`
```

src/structure/structure.ts
```diff
@@ -267,4 +357,258 @@ export class StructureViewer {
...
+     */
+    private _setupGrid() {
+
+      // Determine best arrangement for nwidgets
+      const arrangement = bestArrangement(this._nwidgets);
+
+      if (this._widgets.length > this._nwidgets) {
+        sendWarning(`Warning: Eliminating last ${this._widgets.length - this._nwidgets} widgets.`);
```
Luthaf, line 535: This is a bit strange from the user perspective. When would this occurs?

src/structure/structure.ts
```diff
@@ -267,4 +357,258 @@ export class StructureViewer {
...
+        }
+      }
+
+      const root = document.getElementById(this._id);
+      if (root === null) {
+          throw Error(`Could not get the '#${this._id}' element`);
+      }
+
+      // Initialize the grid root if necessary
+      if (this._root === undefined) {
```
Luthaf, line 548: I would do this in the constructor instead, having `this._root` undefined is not a valid state for this class; and the whole point of a constructor is to create an object in a valid state.

src/structure/structure.ts
```diff
@@ -267,4 +357,258 @@ export class StructureViewer {
...
+        this._root = document.createElement('div');
+        this._root.id = 'grid-root';
```
Luthaf, line 550: Side note, this will fail if we have multiple StructureViewer in the page since the id will no longer be unique

src/structure/structure.ts
```diff
@@ -267,4 +357,258 @@ export class StructureViewer {
...
+        const gi = this._guids.indexOf(guid);
+        if ( gi >= 0 ) {
+          this.onselect(this._currentList[gi], `${guid}-nuke`);
+        }
+
+        if ( widgetRoot !== null) {
+            const oldRoot = widgetRoot.parentNode;
+            if ( oldRoot !== null) {oldRoot.removeChild(widgetRoot); }
+        }
+        const deadCell = document.getElementById(`gi-${guid}`);
```
Luthaf, line 601: Most of the uses of `document.getElementById` should be `getById` instead to catch the case where the element id do not exists in the document

ceriottm:
Re the color scheme, since it doesn't make sense to have an unlimited number of panels, we can just take one of the default plotly palettes and use it. Those are already designed to be contrasting, and using a dynamically generated palette is really overkill.

src/structure/structure.ts
rosecers: I wanted to leave it open if we want to ad-hoc resize the grid with universal buttons. If there are more widgets than the new grid size, we'll eliminate the last few.

src/structure/structure.ts
```diff
@@ -123,10 +128,10 @@ export class StructureViewer {
 
     // GUID of the Active Widget
     private _active: string;
+    // Map of Widgets GUIDS to their color, widget, and current structure
+    private _widgetMap: {[GUID: string]: GriddedWidget; };
```
Luthaf, line 132: An explicit `Map` might be better here, since we also want to delete elements from it. What you've written here is a JavaScript object with unknown keys, which is slightly different (see Object vs Map in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

src/map/map.ts
```diff
@@ -286,13 +300,26 @@ export class PropertiesMap {
...
-    public select(indexes: Indexes) {
+    public select(indexes: Indexes, guid: string = this._active) {
         // Plotly.restyle fires the plotly_click event, so ensure we only run
         // the update once.
         // https://github.com/plotly/plotly.js/issues/1025
-        if (indexes.environment !== this._selected) {
-            this._selected = indexes.environment;
-            this._updateSelectedMarker();
+        if ((this._active !== undefined || guid !== undefined)) {
+          if (guid.includes('nuke')) {
```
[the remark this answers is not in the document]
rosecers: I've changed this to note when the structure/environment indexes passed are negative and nuke then, instead of guid-nuke

src/map/map.ts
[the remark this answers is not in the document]
rosecers: Sorry, this is referring to toggling the buttons (or circles) on the map. Misread the code included.

src/map/map.ts
rosecers: If you want to go to 3D, you need to remove the marker from the scene, but you want it cached in case you go back from 3D to 2D. There might be a better way to do this, I'll think about it.

src/structure/structure.ts
```diff
@@ -267,4 +357,258 @@ export class StructureViewer {
         }
         return this._cachedStructures[index];
     }
+
+    /*
+     * Function to set the active widget for communicating with the map
+     */
+    private _setActive(guid: string) {
+
+      /// this is here to prevent infinite loops
```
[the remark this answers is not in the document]
rosecers: when the map sets active, it points to the structure grid to set active, which points to the map to set active, etc.

src/structure/structure.ts
```diff
@@ -578,15 +564,15 @@ export class StructureViewer {
...
 
-        if ( widgetRoot !== null) {
-            const oldRoot = widgetRoot.parentNode;
-            if ( oldRoot !== null) {oldRoot.removeChild(widgetRoot); }
+        if ( widgetRoot.parentNode !== null) {
+          widgetRoot.parentNode.removeChild(widgetRoot);
         }
-        const deadCell = document.getElementById(`gi-${guid}`);
+
+        const deadCell = getByID(`gi-${guid}`) as HTMLElement;
```
Luthaf, line 574: FYI `getByID` returns an HTMLElement by default, and it can return something else by calling it as `getById<HTMLInputElement>(...)`

src/map/map.ts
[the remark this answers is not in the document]
Luthaf: oh, toggling style/size?

src/map/map.ts
```diff
@@ -876,7 +913,11 @@ export class PropertiesMap {
      */
     private _xValues(trace?: number): number[][] {
         const values = this._properties()[this._current.x].values;
-        const selected = [values[this._selected]];
+        const selected = [];
+        for (const guid of this._guids) {
+          selected.push(values[this._markerMap[guid].current]);
+        }
+
         if (!this._is3D()) {
```
Luthaf, line 921: Just realised this needs to be adapted with the code above, setting all selected markers positions to NaN in 2D mode (and not only the first one), since in 2D mode we are manually dealing with the markers instead of leaving it to plotly for performance reasons.

src/map/map.ts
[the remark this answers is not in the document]
rosecers: yep!

src/structure/structure.ts
```diff
@@ -267,4 +357,258 @@ export class StructureViewer {
...
+
+      if (cell === null) {
+        cell = document.createElement('div');
+        cell.id = cellId;
+        cell.classList.add('chsp-widget-cell', 'grid-item');
+        cell.style.gridColumn = `${colNum}`;
+        cell.style.gridRow = `${rowNum}`;
+
+        // this was removed because it conflicted with the buttons
+        // cell.onclick = () => {this._setActive(guid)}
```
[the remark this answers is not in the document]
rosecers: Okay, I'm going to hold off on this for now.

rosecers:
I’ve handled this by removing them from the root but keeping their data stored

src/map/map.ts
Luthaf: The initial bug this was fixing occurred when going from 2D to 3D and then back to 2D, markers from the 3D mode where not correctly removed

src/map/map.ts
```diff
@@ -1200,53 +1258,133 @@ export class PropertiesMap {
...
+          if (this._is3D()) {
+              let symbol;
+              if (this._current.symbols !== undefined) {
+                  const symbols = this._properties()[this._current.symbols!].values;
+                  symbol = get3DSymbol(symbols[selected]);
+              } else {
+                  symbol = get3DSymbol(0);
+              }
+
+              this._restyle({
```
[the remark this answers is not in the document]
rosecers: Can you elaborate?

src/map/map.ts
```diff
@@ -1200,53 +1258,133 @@ export class PropertiesMap {
...
+                  'x': this._xValues(1),
+                  'y': this._yValues(1),
+                  'z': this._zValues(1),
+
+                  'marker.symbol': symbol,
```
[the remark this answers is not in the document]
rosecers: Can you elaborate? It did not fail.

src/structure/structure.ts
rosecers: Can you elaborate? I can't find the corresponding code.

src/map/map.ts
rosecers: Can you elaborate? Not sure of the changes needed.

Luthaf:
A few other things to be done:

- [ ] Remove duplication of GUID list in structure.ts, merge it with the map as in fd2e2a8
- [ ] Get rid of the `this._active === ''` checkes, there should always be an active widget/marker. The checks should be replaced by assertions
- [ ] Get rid of the `const foo = Map.get(guid), if (foo === undefined)` conditions, either by using assert if `guid === this._active` or with the first point of this list
- [ ] check that we can select envs in 3D mode (I currently get an error `Error: unable to get element with id chsp-selected-3e123c9e-01ed-451d-b296-a6bbcab29599`, which I assume is linked to my refactoring)
    - [ ] Check that switching back and forth between 2D and 3D works fine

src/structure/structure.ts
```diff
@@ -130,39 +182,49 @@ export class StructureViewer {
...
 
     /**
      * Show a new structure, as identified by `indexes`. This will switch to
      * the structure at index `indexes.structure`, and if environments where
-     * passed to the constructor and the current display mode is `"atom"`,
+     * passed to the constructor and the current display mode is `'atom'`,
      * highlight the atom-centered environment corresponding to `indexes.atom`.
      *
      * @param  indexes         structure / atom pair to display
+     * @param  wi              GUID of the widget to show
```
Luthaf, line 215:
```suggestion
     * @param  selectedGUID   GUID of the widget to show
```

src/structure/structure.ts
```diff
@@ -254,4 +338,239 @@ export class StructureViewer {
...
+
+            if (this._selected.has(activeGUID) ) {
+                if (this._selected.has(this._active)) {
+                    const oldButton = getByID(`chsp-activate-${this._active}`);
+                    oldButton.classList.toggle('chsp-inactive-structure-marker', true);
+                    oldButton.classList.toggle('chsp-active-structure-marker', false);
+                    oldButton.innerHTML = `<span class="tooltiptext">Choose as active.</span>`;
+                }
+                this._active = activeGUID;
+                const newButton = getByID(`chsp-activate-${this._active}`);
```
Luthaf, line 365: Same as in the map, the class toggling logic could be extracted to a separate `GridCell` class, with appropriate functions.

This can be refactored later

src/map/map.ts
```diff
@@ -236,12 +239,18 @@ export class PropertiesMap {
...
-    constructor(id: string, indexer: EnvironmentIndexer, properties: {[name: string]: Property}) {
+    constructor(id: string,
+                indexer: EnvironmentIndexer,
+                properties: {[name: string]: Property},
+                starterGuid: string, /// fix me later
+                ) {
         this._indexer = indexer;
-        this._selected = -1;
         this.onselect = () => {};
+        this._active = '';
```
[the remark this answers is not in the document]
rosecers: Yes, otherwise npm doesn't register that this._active has been set in the constructor, unfortunately

ceriottm:
100!

Luthaf:
Remaining TODO items:

- [ ] We are unable to remove a widget from the grid in Safari
- [ ] The Azaphenacenes example fails to load, this is because the Structure constructor tries to load a structure when setting up the grid, when the loading callback is not yet set.

src/structure/structure.ts
```diff
@@ -439,9 +446,9 @@ export class StructureViewer {
             close.classList.add('chsp-close-widget-button', 'btn', 'btn-light', 'btn-sm');
             close.id = `chsp-close-widget-button-${cellGUID}`;
             close.onclick = () => {this._removeWidget(cellGUID); this._setupGrid(this._selected.size); };
-            close.innerHTML = `
+            close.innerHTML = `<i class="fa fa-times"></i>
```
Luthaf, line 449: I would rather not do that, font awesome is available on the default viewer but not a dependency of the core library. We could use svg of the corresponding fa glyphs though.

src/structure/structure.ts
```diff
@@ -463,9 +470,8 @@ export class StructureViewer {
 
             };
 
-            duplicate.innerHTML = `
-                <span class="chsp-tooltip">duplicate viewer</span>
-                <object>${DUPLICATE_SVG}</object>`;
+            duplicate.innerHTML = `<i class="fa fa-clone"></i>
```
Luthaf, line 473: Same as above

src/structure/structure.ts
ceriottm: OK! Will try and fix these as soon as I have the time, but if you want to push a fix in the meantime, feel free

Luthaf:
The Azaphenacenes/on demand loading issue was not too hard to fix, and I also found & fixed an issue with initial setting placement. I've deployed a new beta at http://luthaf.fr/chemiscope/beta/pinning/, please play with it and try to break it! 

Except for bugs found in the beta, this should be good to go!

rosecers:
> The Azaphenacenes/on demand loading issue was not too hard to fix, and I also found & fixed an issue with initial setting placement. I've deployed a new beta at http://luthaf.fr/chemiscope/beta/pinning/, please play with it and try to break it!
> 
> Except for bugs found in the beta, this should be good to go!

Is this error related to those in beta? I got it from switching from Arginine to CSD.
Screen Shot 2020-06-11 at 11 01 48 AM

Luthaf:
> Is this error related to those in beta? 

By "bug found in beta" I mean bugs that will be found in beta, this is one of them! The error is weird, looks like a malformed input JSON. I have a slightly different error, I'll look into this if no-one beats me!

Luthaf:
New blocker issue: 
- [ ] In 3D mode, the map markers are missing

rosecers:
> New blocker issue:
> 
> * [ ]  In 3D mode, the map markers are missing

I thought we didn't want them* in 3D?

Assuming the `selected` markers

Luthaf:
Yeah, I am talking about the 'selected' markers. Why would we not want them? We had one marker before, I would rather keep it =)

rosecers:
> Yeah, I am talking about the 'selected' markers. Why would we not want them? We had one marker before, I would rather keep it =)

Can we make this part of a future PR? Whenever I didn't code them out it would freeze the 3D map

ceriottm:
> > Yeah, I am talking about the 'selected' markers. Why would we not want them? We had one marker before, I would rather keep it =)
> 
> Can we make this part of a future PR? Whenever I didn't code them out it would freeze the 3D map

I think they should be there, I also noticed they were missing, and it's a major regression as one can't see who is who, breaking the viewer/map relationship

ceriottm:
Badass bug: whenever the 2D (but not the 3D!!) plot is redrawn, the last structure in the grid becomes the active one.

ceriottm:
> Badass bug: whenever the 2D (but not the 3D!!) plot is redrawn, the last structure in the grid becomes the active one. Even just zooming in or out will transfer the focus to the last structure.

This seems related to the update_AllMarkers() call on line 1295. which incidentally I guess has to do with the fact that highlighted symbols do not automatically scale when mousewheel-zooming.

@Luthaf  do you think this will be fixed by addressing #11 ?

rosecers:
> > Badass bug: whenever the 2D (but not the 3D!!) plot is redrawn, the last structure in the grid becomes the active one. Even just zooming in or out will transfer the focus to the last structure.
> 
> This seems related to the update_AllMarkers() call on line 1295. which incidentally I guess has to do with the fact that highlighted symbols do not automatically scale when mousewheel-zooming.
> 
> @Luthaf do you think this will be fixed by addressing #11 ?

I know what this is from -- when switching from 3D to 2D, the map adds all of the markers back, and the `add` function makes the new marker the active one. I can fix this pretty easily I think.

ceriottm:
Was actually a separate bug in which just zooming in 2D would switch the active point. I pushed a fix but it's ugly, the problem is that updateMarkers function also changes the active point. 
Also, related to the bug you fixed, there is still the reverse problem, when switching from 2D to 3D changes the active point.

ceriottm:
as far as I can tell, this works well and I can't find any additional bugs
would be in favor of merging, as this has got quite big and it risks becoming a pain to merge it with other developments.

Luthaf:
Ok, this looks good to me. I rebased & force pushed since we had duplicated commits. If you want to check it out locally, make sure to use `git fetch rosecers` `git reset --hard rosecers/feature/pinning`, and **not** `git pull`.

Issue #1, opened 2020-01-22
Switch between 'atom' and 'structure' display mode


Luthaf:
If the dataset contains both atom and structure properties, we currently default to showing the 'atom' ones, but the user should have a way to switch between the two modes.

The display mode is already centralized in the EnvironmentIndexer, this mainly needs to be a setting somewhere.

rosecers:
would potentially live within the map code and be changeable within the map settings

- [] visually add option
- [] propagate the change to all elements

PR #26, opened 2020-05-30
Command-line version of chemiscope_input

ceriottm:
Just added options to have chemiscope_input double up as command-line tool. 
If one has say an extended xyz file with structures and data, it can be converted to chemiscope.


utils/chemiscope_input.py
```diff
@@ -210,3 +210,54 @@ def write_chemiscope_input(filename, frames, meta=None, extra=None, cutoff=None)
...
+
+    # command-line execution. requires ASE IO module
+    import argparse
+    try:
+        import ase.io as ase_io
+    except:
+        raise ImportError("chemiscope_input needs ASE modules to parse structure inputs")
+    
+    parser = argparse.ArgumentParser(description=helpstring)
+    parser.add_argument('-i', '--input', type=str, required=True,
```
Luthaf, line 234: I would use a positional argument for the input, so that the simplest way of using this is `chemiscope_input.py file.xyz` instead of `chemiscope_input.py -i file.xyz`

utils/chemiscope_input.py
```diff
@@ -210,3 +210,54 @@ def write_chemiscope_input(filename, frames, meta=None, extra=None, cutoff=None)
...
+                        help='List of references for the dataset.')                 
+    args = parser.parse_args()
+
+    # reads filen with ASE    
+    frames = ase_io.read(args.input, ':')
+    
+    # determines output file name automatically if missing
+    output = args.output
+    if args.output is None:
+        output = args.input + "_chemiscope.json"
```
Luthaf, line 257: As a default, I would rather use `.json.gz` as the default extension, since there are no drawback and size advantages

rosecers: Also, shouldn't we remove the input extension? Otherwise, it can become (minorly) cumbersome with "file.xyz.json.gz" when one might prefer "file.json.gz"

utils/chemiscope_input.py
```diff
@@ -210,3 +210,54 @@ def write_chemiscope_input(filename, frames, meta=None, extra=None, cutoff=None)
...
+                        help='Dataset name.')
+    parser.add_argument('--description', default="", type=str,
+                        help='Dataset description.')
+    parser.add_argument('--authors', nargs='*', type=str, default=[],
+                        help='List of dataset authors.')
+    parser.add_argument('--references', nargs='*', type=str, default=[],
+                        help='List of references for the dataset.')                 
+    args = parser.parse_args()
+
+    # reads filen with ASE
```
rosecers, line 251: Minor type "filen"

utils/chemiscope_input.py
Luthaf: If someone prefer another extension, they can set it explicitly with `--output`. I would not try to deal with file extension here, it can be hard to do correctly (for example, what should we do with `file.xyz.gz`, or `HISTORY` DL_POLY files?)

Luthaf: And I realize my initial comment was not very clear, I meant appending `_chemiscope.json.gz` to create the output file name, instead of `_chemiscope.json`

utils/chemiscope_input.py
```diff
@@ -210,3 +210,54 @@ def write_chemiscope_input(filename, frames, meta=None, extra=None, cutoff=None)
...
+    
+    # determines output file name automatically if missing
+    output = args.output
+    if args.output is None:
+        output = args.input + "_chemiscope.json"
+    
+    write_chemiscope_input(output, frames, 
+        meta=dict(name=args.name, description=args.description,
+                authors=args.authors, references=args.references),
+        extra=None, cutoff=args.cutoff)
```
rosecers, line 262: Ideally, one should be able to flag automatic property imports. If sticking to an ase-like format, my thoughts would be parser arguments to flag atom- or structure-centered properties.

Luthaf: A chemiscope file can contain both atom and structure properties, and if a file contains both, the viewer will default to atom properties. #1 is about allowing the user to display the other kind of properties.

Until #1 is implemented, I think having a flag to select atom or structure properites is a good solution. It can be implemented as 

```py
if args.only_structure:
    for frame in frames:
        frame.arrays = {}
elif args.only_atoms:
    for frame in frames:
        frame.info = {}
```

utils/chemiscope_input.py
ceriottm: Here I've mixed feelings. I avoided stripping the extension just because I didn't know how to do that. Perhaps ASE provides a way to remove the extension based on the same logic it uses to detect file format? otherwise we can just drop whatever comes after the last dot.  "DOS style"

utils/chemiscope_input.py
ceriottm: I agree this would be nice to have as makeshift while we wait for #1 
My impression was actually that it only uses environment properties if one selects a cutoff value, am I off? 
Another thing that is annoying - but is more a problem with `write_chemiscope_input` than with the command-line wrapper - is the fact that it picks up the atomic numbers array as a property that should be stored in chemiscope.

ceriottm:
@Luthaf feel free to do the changes you suggested. They are all very sensible and probably it'll take less time than to wait that I find the time to do those myself ^_^

utils/chemiscope_input.py
Luthaf: > My impression was actually that it only uses environment properties if one selects a cutoff value, am I off?

No, that's right, I remembered the logic wrongly (cf https://github.com/cosmo-epfl/chemiscope/blob/815eabec23c6e30910077d07d2cc5da54e6857b0/src/index.ts#L126), passing a `cutoff` value created the `environments` array.

> Another thing that is annoying - but is more a problem with write_chemiscope_input than with the command-line wrapper - is the fact that it picks up the atomic numbers array as a property that should be stored in chemiscope.

This is actually an ASE thing, which always create `frame.arrays["number"]`. We could ignore it in `write_chemiscope_input`, but then it is sometime useful, cf the CSD-1000 dataset. We could add a note on this subject in the documentation though.

utils/chemiscope_input.py
Luthaf: Since this is just a default and the user can always specify an output file name, I would keep it like this and don't try to deal with file extension.

utils/chemiscope_input.py
ceriottm: I would ignore arrays["numbers"]: if one wants that information, *e can add a string property with the element labels, or just duplicate that property calling it "atomic_number". more often than not I suspect this is not useful, and unfortunately seems you cannot remove "numbers" from arrays before you pass frames to chemiscope_input because it breaks the internal ASE handling of the frames.

Luthaf: This convinced me =). I'll check if the code below still works then:

```py
if args.only_structure:
    for frame in frames:
        frame.arrays = {}
```

utils/chemiscope_input.py
rosecers: Fine with me

Issue #7, opened 2020-01-31
Add "download as standalone" button in default interface


Luthaf:
We already have the ability to generate a "standalone" visualization as an HTML file, the idea is to make it easier for users to download such file including their dataset.

- building the file automatically on Travis & deploying it to github pages
- when the user clicks on the button, download `standalone.html`, stringify the dataset back to JSON (including visualization state #6) and add it at the end of the file. 
- open a download dialog from Javascript for the user to save the file.

ceriottm:
The PR #29 literally solves this issue, but not in spirit - we need first to be able to save the status of che viewer in a json (#24 etc) and then we can decide whether to download the json or a standalone viewer that incorporates it.

Luthaf:
Done as part of #36

PR #29, opened 2020-06-05
Make the standalone viewer downloadable 

ceriottm:
Travis deploys chemiscope_standalone.html when it builds the website. A button is added to make it downloadable.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -46,10 +46,10 @@
                         <li class="nav-item">
                             <a class="nav-link" data-toggle="modal" href="#about"><i class="fa fa-user"></i> About</a>
                         </li>
-                        <li class="nav-item">
+                        <li class="nav-item hide-if-standalone">
                             <a class="nav-link" data-toggle="modal" href="#download"><i class="fa fa-download"></i> Download</a>                        
                         </li>
-                        <li class="nav-item dropdown">
+                        <li class="nav-item dropdown hide-if-standalone">
                             <a class="nav-link dropdown-toggle" href="#" id="examples" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                                 <i class="fa fa-images"></i> Examples
                             </a>
--- a/app/index.html
+++ b/app/index.html
@@ -47,7 +47,7 @@
                             <a class="nav-link" data-toggle="modal" href="#about"><i class="fa fa-user"></i> About</a>
                         </li>
                         <li class="nav-item hide-if-standalone">
-                            <a class="nav-link" data-toggle="modal" href="#download"><i class="fa fa-download"></i> Download</a>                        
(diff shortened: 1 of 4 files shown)


ceriottm:
This addresses partly #7, but not completely as it doesn't download the JSON

app/index.html
```diff
@@ -180,6 +180,45 @@ <h5>References</h5>
...
+                    which is available on GitHub. 
+                </p>
+
+                <p>
+                    If you are a user, you may want to download a 
+                    <a href="chemiscope_standalone.html" download> 
+                    standalone version
+                    </a> of the standard viewer, which you can use to 
+                    display locally datasets prepared in the chemiscope 
+                    JSON format.
```
Luthaf, line 215: Should we add some basic instructions on how to use the standalone viewer (i.e. append the JSON file)? Or should we leave it as is and wait for the full implementation of "download current dataset as standalone"?

app/index.html
```diff
@@ -180,6 +180,45 @@ <h5>References</h5>
...
+
+                <p>
+                    If you are a developer you may be interested in downloading
+                    the <a href="https://github.com/cosmo-epfl/chemiscope"> source code</a>,
+                    which is available on GitHub. 
+                </p>
+
+                <p>
+                    If you are a user, you may want to download a 
+                    <a href="chemiscope_standalone.html" download>
```
Luthaf, line 211: Regarding the "recursive" issue, we could have some CSS to hide this and a other part of the page (such as the list of examples) which are not relevant in the standalone mode, which would be activate depending on the `isStandalone` variable. Something like

```html
<div class="hide-if-standalone">

</div>

<script>
if (isStandalone) {
    for (const element of document.getElementsByClassName("hide-if-standalone") {
        element.style.display = "none";
    }
}
</script>
```

ceriottm: this would be great. you know how to implement it easily?

app/index.html
ceriottm: I was thinking that this "static standalone viewer" can be used as a local GUI, and you have to load your dataset in it.

app/index.html
Luthaf: The code above should work. Do you want to add this to this PR or leave it for later?

ceriottm:
Ready to merge IMO @Luthaf

Luthaf:
The commit history is very strange ... Did you `git rebase master` and then `git pull` `git push`? The same commits exist multiple times, and I see already on master. Usually, after a rebase, you want to do a force push with `gith push --force-with-lease` (--force-with-lease only force push if no one else pushed to the branch in the meantime)

ceriottm:
uhm. might have done that, indeed. can be fixed or we need to nuke the
branch?

PR #32, opened 2020-06-05
Option to automatically center the selected environment in the viewer

### LINKED ISSUE #23: Center selected environment in viewer
Add a button (perhaps to the right of the environment indicator 
image
That would center the JSMol widget on the selected environment. 
In JSMol language, this seems to be as simple as 
select (*)[index]
centerAt average

ceriottm:
Also rearranged the environment options to make space for the centering switch.
Fixes #23 - all in all it is both simpler and better to have this applied automatically on environment switch than to have a button that one has to click manually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ node_js:
   - 12
 
 install:
+    # we need the full history to be able to get the chemiscope version with
+    # git describe
+    - git fetch --unshallow
     - npm ci
     - python3 -m pip install -r docs/requirements.txt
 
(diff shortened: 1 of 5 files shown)


ceriottm:
Probably we need to center the whole thing by default, as I noticed that otherwise the JSMol viewer remain centred on the last environment, if you enable and then disable the autocenter feature.

Luthaf:
My only issue with this is that the environment settings are starting to be a bit messy: 

Capture d’écran 2020-06-08 à 10 26 34

I think we can organize them a bit better to improve the overall UX of these settings

Luthaf:
> Probably we need to center the whole thing by default, as I noticed that otherwise the JSMol viewer remain centred on the last environment, if you enable and then disable the autocenter feature.

That would work, otherwise users can translate the view with ctrl + drag or shift + ctrl + drag.

ceriottm:
OK fixed the collapsible box. Looks good enough to me, The enable/disable button in my opinion could go once we have implemented #1

PR #36, opened 2020-06-06
Loadable settings for map

### LINKED ISSUE #6: Default visualization settings in JSON input
For the default interface, we could have a way to specify a visualization state in the input JSON file. 

On the map side, this means describing which property should be used as color/size/symbols, x/y/z ranges, etc.

On the structure viewer side, this means supercell settings, visualizations settings, etc.

---

This can be implemented by adding serialization/de-serialization of the settings to JSON, and having an additional section for these in the JSON input file.

ceriottm:
Initial attempt at an infrastructure that fixes #6 
Very much needing discussion and pair programming.


ceriottm:
This is an example JSON with settings for the axes and coloring of the map. 
[test.json.gz](https://github.com/cosmo-epfl/chemiscope/files/4742039/test.json.gz)

src/map/map.ts
```diff
@@ -579,9 +581,9 @@ export class PropertiesMap {
 
     /**
      * Fill possible values for the settings, depending on the properties in
-     * the dataset
+     * the dataset. Can be initial
      */
-    private _setupSettings() {
+    private _setupSettings(presets :  { [name: string] : any } | undefined) {
```
Luthaf, line 586: FYI, `value: Type | undefined` can be written `value?: Type`

src/map/map.ts
```diff
@@ -594,13 +596,33 @@ export class PropertiesMap {
...
+        };
+
+        if ( presets !== undefined ) {
+            console.log("setupSettings presets");
+            console.log(presets)
+            console.log("color", presets["color"])
+            for ( const preset in default_presets) {
+                console.log(preset, " >> ", presets[preset])
+                if ( preset in presets) {
+                    default_presets[preset] = presets[preset];
```
Luthaf, line 614: This is were we should validate that the preset/user-provided settings correspond to an existing property in the dataset

src/dataset.ts
```diff
@@ -37,6 +37,16 @@ export interface Dataset {
      * environment. This may change in the future.
      */
     environments?: Environment[];
+    
+    /**
+     * Visualization settings to be used when loading the dataset.
+     * 
+     * Basically just a dictionary: visualization options will have to be
```
Luthaf, line 44: This is not a dictionnary in JS/TypeScript, but an object with arbitrary keys. The `Map` type would be closer to a dictionary.

src/dataset.ts
```diff
@@ -37,6 +37,16 @@ export interface Dataset {
      * environment. This may change in the future.
      */
     environments?: Environment[];
+    
+    /**
+     * Visualization settings to be used when loading the dataset.
+     * 
+     * Basically just a dictionary: visualization options will have to be 
+     * matched to whatever exists in a given implementation of the visualizer.
```
Luthaf, line 45: This interface is used for the DefaultVisualizer only, so we can control what is accepted there. Instead of using an arbitrary object (`{ [name: string] : any };`), I would create an explicit interface for it and check that the dataset we get follow said interface. Something like 

```ts
interface MapSettings {
   color: string,
   x: string,
   y: string,
   z: string,
    // ...
}

presets?: {
    map?: Partial<MapSettings>,
    // ...
}
```

Luthaf: Having an explicit interface allow to (a) better document what is accepted and (b) get the compiler to yell at us if we make a mistake.

Checking is allow to catch "wrong" input files early, and give the user a good error message.

app/static/js/default.js
```diff
@@ -146,3 +149,16 @@ function setupDefaultChemiscope(j2sPath) {
         reader.readAsArrayBuffer(upload.files[0]);
     }
 }
+
+function download(filename, text) {
+  var element = document.createElement('a');
+  element.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(text));
```
Luthaf, line 155: Since this is using a `data:` URL, we should check for URL size limit. I don't know if large files will be downloadable this way.

src/map/map.ts
```diff
@@ -128,38 +128,89 @@ function arrayMaxMin(values: number[]): {max: number, min: number} {
...
-
-        this.property = new HTMLSetting('string', initial);
+        this.property = new HTMLSetting('string', '');
+        this.scale = new HTMLSetting('string', '');
+                
+        this.applyPresets(presets);
+        
         this.property.validate = (value) => {
-            if (properties.includes(value) || (initial === '' && value === '')) {
+            if (properties.includes(value) || value === '') {
```
Luthaf, line 170: Here we loose the validation that prevent setting x/y to an empty string. We could keep both the inital (perhaps renamed to default) & preset arguments to be able to do this.

src/map/map.ts
```diff
@@ -128,38 +128,89 @@ function arrayMaxMin(values: number[]): {max: number, min: number} {
...
+    constructor(properties: string[], presets?: AxisPresets) {
+        this._properties = properties;
+        
+        // creates settings
         this.max = new HTMLSetting('number', 0);
         this.min = new HTMLSetting('number', 0);
-
-        this.property = new HTMLSetting('string', initial);
+        this.property = new HTMLSetting('string', '');
+        this.scale = new HTMLSetting('string', '');
```
Luthaf, line 165: Should be initialized to `'linear'`, since an empty string is not a valid value

src/map/map.ts
```diff
@@ -128,38 +128,89 @@ function arrayMaxMin(values: number[]): {max: number, min: number} {
...
+                
+        this.applyPresets(presets);
```
Luthaf, line 167: I would apply the presets after setting `validate`, or else bad preset will not be caught

src/map/map.ts
```diff
@@ -128,38 +128,89 @@ function arrayMaxMin(values: number[]): {max: number, min: number} {
...
+    scale?: string;
     /// The minimal value for this axis
-    public min: HTMLSetting<'number'>;
+    min?: number;
     /// The maximal value for this axis
-    public max: HTMLSetting<'number'>;
+    max?: number;
+}
+
+const AxisDefaults = {
```
Luthaf, line 142: Being a global variable, I would use `SCREAMING_SNAKE_CASE` instead of `CapitalCamelCase` which is usually reserved for types

src/dataset.ts
ceriottm: I created an interface for the input to the default visualizer, although it doesn't seem I ever need to use it...

app/static/js/default.js
ceriottm: could download a ~10MB json here. this whole concept is far from ideal though, e.g. hardcoded filename is crap. I let you change this based on best practices and your own instinct.

src/map/map.ts
ceriottm: I think that the fact that x and y axes are special should be dealt with in the MapSettings - in general we want to be able to have axes with no settings.

src/map/map.ts
ceriottm: Had put it there so that there was an initial value already set for the validate function to pick up. I now streamlined the whole validation thing

app/static/js/default.js
Luthaf: Another options seems to be URL.createObjectURL(), which points to a `Blob` object, which is as unlimited as JS code goes (https://stackoverflow.com/a/43816041/4692076)

Code would look like this:
```js
const a = document.createElement('a');
a.download = filename;
a.href = URL.createObjectURL(new Blob([content], {type: mime}));
a.style.display = 'none';

document.body.appendChild(a);
a.click();
setTimeout(() => {
    document.body.removeChild(a);
    URL.revokeObjectURL(a.href);
}, 2000);
```

ceriottm: Really, do as you see fit: I have no strong feeling at all.

src/map/map.ts
```diff
@@ -925,7 +739,10 @@ export class PropertiesMap {
         }
         
         // make a copy of the default layout
-        const layout = {...DEFAULT_LAYOUT};
+        //const layout = {...DEFAULT_LAYOUT};
+        // makes a deep copy - otherwise DEFAULT_LAYOUT entries get overwritten 
+        const layout = JSON.parse(JSON.stringify(DEFAULT_LAYOUT));
```
Luthaf, line 826: Good catch! The previous code only did a copy of top level items =/. You can also remove the old code here!

package.json
```diff
@@ -45,6 +45,7 @@
     "webpack-dev-server": "^3.11.0"
   },
   "dependencies": {
+    "clone": "^2.1.2",
```
Luthaf, line 48: Does not seems to be used, since you went for `JSON.parse(JSON.stringify))` instead

ceriottm: True! Wanted to ask you if that is good enough, or you want to pull in another dependency. Actually this change we should probably cherry-pick into master, because it gives rise to a nasty bug where when you zoom in and then load another dataset the "default zoom level" is set to the zoom of the previous dataset.

Luthaf: Well, I am a bit ambivalent on very small JS dependencies where you end up depending on 10k packages all doing one small thing. Here, I think the JSON solution is pretty good, especially if it is commented.

ceriottm: ok then perfect. just do it. if you think you can work on this PR and merge it before the end of the week, then I'd leave this in here, otherwise I think a bugfix applied on master would be a good idea as this change in plot range persisting across loading new datasets is weird.

rosecers:
Currently the version at `chemiscope.org` doesn't reset the axes when switching datasets (i.e. I zoomed in on one dataset, switched the dataset, and it stayed zoomed in). This PR fixes that issue, so I wanted to boost it in case you weren't aware of the bug

Luthaf:
I'll take the corresponding fix out of this PR and create a separate one then =)

Luthaf:
Done in #49. I'll rebase this PR to fix conflicts and incorporate the fix.

src/map/settings.ts
```diff
@@ -0,0 +1,201 @@
...
+        this.applyPresets(presets);        
+    }
+        
+    public applyPresets(presets?: AxisPresets) : void {
+        /**
+         * Applies presets, possibly filling in with default values
+         */
+                 
+        /// Combine defaults with the given presets    
+        const initial : Required<AxisPresets> = { ...AXIS_DEFAULTS, ...presets };
```
Luthaf, line 55: This would remove all customization made by the user when applying presets. Is it expected?

Luthaf:
Squashed, rebased and partially refactored. Please use `git fetch; git reset --hard origin/feat/json_settings` when updating this branch locally 😃 

Still TBD:

- [ ] Move code dealing with HTML part of the settings to `structures/settings.ts`
- [ ] Setting validation on the structure side
- [ ] Use Blob to download dataset + preset
    - [ ] decide on UI, should downloading/applying settings alone allowed?
- [ ] Add presets to the examples online

ceriottm:
I have already an example with settings for the structure, that was loading on my branch (haven't rechecked)
[CSD-500.json.gz](https://github.com/cosmo-epfl/chemiscope/files/4814000/CSD-500.json.gz)
re (3.1) I would allow downloading and applying settings alone, I think it can be useful not to move around heaps of data. 
one important decision to make is if we want to save "multi-viewer" state, or just save the state of the active structure and always load a single panel.

Luthaf:
Ok, this should be ready to review! @ceriottm and @rosecers do you have a bit of time to have a look? The overall PR is quite large, but commits should be self contained and reviewable individually.

Still TBD before final merge:

- [ ] Write documentation

Luthaf:
You can download new version of the example files with presets by running `npm run download-example-input` again

Luthaf:
Ok, I'm done with this, except for any bug you may find!

rosecers:
We should provide better error when someone tries to load a dataset/visualization state with the wrong load.

Loaded `dataset.json` into visualization state and got:

```Error: invalid "presets.meta" key in chemiscope config

Backtrace
Error: invalid "presets.meta" key in chemiscope config
at validatePresets (http://localhost:8080/chemiscope.min.js:147868:19)
at DefaultVisualizer.Chemiscope../src/index.ts.DefaultVisualizer.applyPresets (http://localhost:8080/chemiscope.min.js:147988:9)
at FileReader.reader.onload (http://localhost:8080/static/js/default.js:197:24)```

Loaded `settings.json` into dataset and got:

```Error: missing "meta" key in the dataset

Backtrace
Error: missing "meta" key in the dataset
at Object.validateDataset (http://localhost:8080/chemiscope.min.js:147489:15)
at new DefaultVisualizer (http://localhost:8080/chemiscope.min.js:147882:19)
at http://localhost:8080/chemiscope.min.js:147956:30
at new Promise (<anonymous>)
at Function.Chemiscope../src/index.ts.DefaultVisualizer.load (http://localhost:8080/chemiscope.min.js:147955:16)
at setupChemiscope (http://localhost:8080/static/js/default.js:70:34)
at FileReader.reader.onload (http://localhost:8080/static/js/default.js:174:13)
```

rosecers:
What is supposed to happen if you load irrelevant settings? Nothing happens on my build. 

I loaded `settings.json` generated for Arg-Dipep on azaphenacenes, and the window did nothing.

rosecers:
I cancelled loading settings and got:

```
TypeError: Failed to execute 'readAsArrayBuffer' on 'FileReader': parameter 1 is not of type 'Blob'.

Backtrace
TypeError: Failed to execute 'readAsArrayBuffer' on 'FileReader': parameter 1 is not of type 'Blob'.
at HTMLInputElement.loadPresets.onchange (http://localhost:8080/static/js/default.js:199:16)

```

rosecers:
Should we provide an option to not override the selected widgets/markers? It seems likely that someone saves a visualization setting (coloring, axis, etc) that they might want to apply to multiple sets of molecules/materials

rosecers:
Should the settings also preserve the zoom window?

Luthaf:
> We should provide better error when someone tries to load a dataset/visualization state with the wrong load.

That's a good idea!

> What is supposed to happen if you load irrelevant settings?

You get an error? The current one is not perfect, but better than silently failing.

> I cancelled loading settings and got

Somehow expected, should we try to make the error message better?

> Should we provide an option to not override the selected widgets/markers? It seems likely that someone saves a visualization setting (coloring, axis, etc) that they might want to apply to multiple sets of molecules/materials

The core of the issue here is that `saveSettings` saves everything, including the pinned environments. We could provide checkboxes to select only one of the three settings parts (map/structure/pinned), what do you think? It is also possible to manually create a setting file (or delete irrelevant parts from the saved one) and load that.

----

BTW, I've updated the examples files to use the new name (settings vs presets), make sure to get them!

Luthaf:
> Should the settings also preserve the zoom window?

Yes, this is to be done later

rosecers:
> > I cancelled loading settings and got
> 
> Somehow expected, should we try to make the error message better?
> 
I don't think this should have an error -- if someone clicks `load/save` accidentally or changes their mind, that's not error-worthy

> > Should we provide an option to not override the selected widgets/markers? It seems likely that someone saves a visualization setting (coloring, axis, etc) that they might want to apply to multiple sets of molecules/materials
> 
> The core of the issue here is that `saveSettings` saves everything, including the pinned environments. We could provide checkboxes to select only one of the three settings parts (map/structure/pinned), what do you think? It is also possible to manually create a setting file (or delete irrelevant parts from the saved one) and load that.
> 
I think we should have checkboxes for map/structure/pinned

> BTW, I've updated the examples files to use the new name (settings vs presets), make sure to get them!
/thumbsup

ceriottm:
To me choosing which part to save is overkill, and making the UI too
complicated. If one wants something special, the format is easy enough to
edit.
What is the use case you have in mind? The only things I can think of are
setting up the best "first impression" for a dataset (incorporating
settings into the database file) or "hey look at this" where we are both
working on the same dataset and I just want to highlight something. For
that a whole settings dump is the most effective.

Luthaf:
> To me choosing which part to save is overkill, and making the UI too complicated. If one wants something special, the format is easy enough to edit.

I've added the checkboxes in the mean time, so everyone can have a quick look. It does clutter UI a bit, but that's bearable. If we decide against it I can remove the corresponding commit 😃

ceriottm:
It is a bit confusing that when you load or save something the modal remains open. 
I think it'd be good to close the modal as soon as the button is clicked or the filename selected, and show a progress indicator of some sort.

Luthaf:
Agreed that this would be nice to have, however I would keep it for a PR later so that I can merge this and deploy in on materials cloud before tomorrow.

PR #39, opened 2020-06-07
Full support for sparse environment list

ceriottm:
First commit. Basically just disabled a bunch of checks so that an environment list that is not as long as #atoms can be loaded


ceriottm:
Test JSON with sparse environments 
[sparse-environs.json.gz](https://github.com/cosmo-epfl/chemiscope/files/4744948/sparse-environs.json.gz)

Luthaf:
Closing since there is not much code worth salvaging in this PR. We need someone to spend a bit of time to implement this feature =)

PR #41, opened 2020-06-08
Also remove settings modales on remove()

Luthaf:


Luthaf:
Merging quickly since master (meaning the deployed version too) is currently broken

PR #43, opened 2020-06-09
Not very elegant, but at least we avoid an ugly error box that has ap…

ceriottm:
…parently no reason and no impact. @Luthaf  you'll love it ^_^

### DIFF
--- a/app/static/js/default.js
+++ b/app/static/js/default.js
@@ -73,6 +73,11 @@ function setupChemiscope(dataset) {
 }
 
 function displayError(error) {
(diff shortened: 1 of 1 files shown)


Luthaf:
This works for me =) @rosecers, you had another error message for this, right?

ceriottm:
@rosecers if you have a different error why don't you add it in this same PR? Or perhaps we could have a list of ignored errors from JSMol?

rosecers:
> @rosecers if you have a different error why don't you add it in this same PR? Or perhaps we could have a list of ignored errors from JSMol?

This is the primary error that I get. I get another, very similar one, but coincidentally cannot recreate it. I'll make an issue next time I come across it.

Luthaf:
Let's update the condition next time we find one of these errors

PR #44, opened 2020-06-09
Remove usage of any

Luthaf:
While preparing the code to use eslint instead of the [deprecated tslint](https://medium.com/palantir/tslint-in-2019-1a144c2317a9), I found a few potential ameliorations.

The main one consist in using `Record<string, unknown>` and `unknown` in general instead of `any`. This means the compiler will ensure we check the type/values before using them.

### DIFF
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -143,28 +143,31 @@ export interface Environment {
     cutoff: number;
 }
 
+/** Arbitrary javascript object, to be validated */
+export type JsObject = Record<string, unknown>;
+
 /** @hidden
  * Check that the given object, potentially comming from javascript, has all
  * required properties to be a dataset.
  */
-export function checkDataset(o: any) {
-    if (!('meta' in o && typeof o.meta === 'object')) {
+export function validateDataset(o: JsObject) {
+    if (!('meta' in o && typeof o.meta === 'object' && o.meta !== null)) {
         throw Error('missing "meta" key in dataset');
     }
-    checkMetadata(o.meta);
+    checkMetadata(o.meta as JsObject);
 
-    if (!('structures' in o && o.structures.length !== undefined)) {
+    if (!('structures' in o && Array.isArray(o.structures))) {
         throw Error('missing "structures" key in dataset');
     }
     const [structureCount, envCount] = checkStructures(o.structures);
 
-    if (!('properties' in o && typeof o.properties === 'object')) {
+    if (!('properties' in o && typeof o.properties === 'object' && o.properties !== null)) {
         throw Error('missing "properties" key in dataset');
     }
-    checkProperties(o.properties, structureCount, envCount);
+    checkProperties(o.properties as JsObject, structureCount, envCount);
 
     if ('environments' in o) {
-        if (typeof o.environments.length === undefined) {
+        if (!Array.isArray(o.environments)) {
             throw Error('"environments" must be an array in dataset');
         }
 
@@ -175,7 +178,7 @@ export function checkDataset(o: any) {
     }
 }
 
-function checkMetadata(o: any) {
+function checkMetadata(o: JsObject) {
     if (!('name' in o && typeof o.name === 'string')) {
         throw Error('missing "meta.name" key in dataset');
     }
@@ -185,7 +188,7 @@ function checkMetadata(o: any) {
     }
 
     if ('authors' in o) {
-        if (typeof o.authors.length === undefined) {
+        if (!Array.isArray(o.authors)) {
             throw Error('"meta.authors" must be an array in dataset');
         }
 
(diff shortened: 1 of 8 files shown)


src/dataset.ts
```diff
@@ -143,28 +143,31 @@ export interface Environment {
...
+
 /** @hidden
  * Check that the given object, potentially comming from javascript, has all
  * required properties to be a dataset.
  */
-export function checkDataset(o: any) {
-    if (!('meta' in o && typeof o.meta === 'object')) {
+export function validateDataset(o: JsObject) {
+    if (!('meta' in o && typeof o.meta === 'object' && o.meta !== null)) {
         throw Error('missing "meta" key in dataset');
```
rosecers, line 155:
```suggestion
        throw Error('Missing "meta" key in dataset containing the metadata.');
```

rosecers: In general I think the errors could be more descriptive of what they're flagging.

src/dataset.ts
```diff
@@ -185,7 +188,7 @@ function checkMetadata(o: any) {
     }
 
     if ('authors' in o) {
-        if (typeof o.authors.length === undefined) {
+        if (!Array.isArray(o.authors)) {
             throw Error('"meta.authors" must be an array in dataset');
```
rosecers, line 192: Can we add string comprehension in here?

src/dataset.ts
```diff
@@ -220,33 +223,34 @@ export function checkStructures(o: any): [number, number] {
...
             throw Error(`structure ${i} must contains either 'names'/'x'/'y'/'z' or 'data'`);
-
         }
     }
 
     return [o.length, envCount];
 }
 
-function checkProperties(properties: any, structureCount: number, envCount: number) {
+function checkProperties(properties: JsObject, structureCount: number, envCount: number) {
```
rosecers, line 251: Could we instead have properties be a list of JsObjects, then the casting below becomes unnecessary?

src/structure/structure.ts
```diff
@@ -33,22 +33,27 @@ function groupByStructure(n_structures: number, environments?: Environment[]): E
  * issue with `s` if any, or the empty string if `s` looks like a valid
  * structure.
  */
-function checkStructure(s: any): string {
+function checkStructure(s: JsObject): string {
```
rosecers, line 36: Would it be possible to move this function to `dataset.ts` and employ it in dataset.checkStructures? The two functions contain a lot of the same code.

src/structure/structure.ts
```diff
@@ -243,7 +248,7 @@ export class StructureViewer {
     private _structureForJSmol(index: number): string {
         if (this._cachedStructures[index] === undefined) {
             const s = this.loadStructure(index, this._structures[index]);
-            const check = checkStructure(s);
+            const check = checkStructure(s as unknown as JsObject);
```
rosecers, line 251: I don't understand this syntax.

src/dataset.ts
Luthaf: You mean something like this?

```
            throw Error(`"meta.authors" must be an array in dataset, got ${o.authors}`);
```

src/dataset.ts
Luthaf: properties is an object, so we can not have a list here

```js
properties = {
    "KPCA[1]": {...},
    "KPCA[2]": {...},
     energy: {...} 
}
```

src/structure/structure.ts
Luthaf: Good idea!

src/structure/structure.ts
Luthaf: Here the compiler think that `s` is a `Structure`, but we want to check that. We can not cast `s as JsObject` since `Structure` and `JsObject` are too different, so we have to go through `unknown` first.

src/dataset.ts
rosecers: Error is better, but by string comprehension, I mean not enforcing that Authors need to be a list, but rather parse a string into a list ala bibtex

src/dataset.ts
rosecers: Ah, so could it be a map of JSObjects?

Luthaf: Could be a `Record<string, JsObject>` indeed. I'll change this

src/dataset.ts
Luthaf: I don't understand what you mean here. Transforming `"Doe, John and Jean, Michel"` into `["John Doe", "Michel Jean"]`? I would rather not go this way, if people want to use something based on bibtex they can write some code to create the corresponding JSON file.

rosecers: Sure, just a suggestion.

src/dataset.ts
Luthaf: I tried to improve the messages, although in this specific case more specific messages will be given in the `checkMetadata` function

PR #46, opened 2020-06-12
Fixes a bug with the standalone generator.

ceriottm:
Would be good to figure out why the standalone is not autodeployed, while we are at it. @Luthaf ?


Luthaf:
> Would be good to figure out why the standalone is not autodeployed, while we are at it.

What do you mean? `https://chemiscope.org/chemiscope_standalone.html` exists, and the download link works for me.

ceriottm:
yes, but it's not updated. still downloads the single-panel version.

utils/generate_standalone.py
```diff
@@ -87,6 +87,8 @@ def main():
                         help='chemiscope JSON input that is appended to provide default data')
     parser.add_argument('-o', '--output', type=str, default="",
                         help='output file name. Defaults to chemiscope.html or the name of the JSON')
+    parser.add_argument('-l', '--loader', default=False, action='store_true',
```
Luthaf, line 90: I don't think this flag warrant a short form, I usually prefer command line arguments to be more explicit.

ceriottm: fine with me

Issue #45, opened 2020-06-12
Refactor, cleanup and improvement linked to the pinning feature


Luthaf:
Taking the list from #25 review

- [x] Do not call `select({-1, -1}, guid)` to indicate removal, instead add specific functions for each possible action: `addPinned`, `removePinned`, `changeActivePinned`. Then `select` always act on the current `active` and only move the marker/change the structure
- [x] Remove the `starterGUID` parameter for `PropertyMap` constructor to allow using the map without a structure viewer. This should be easy to do after the above changes
- [ ] Extract code dealing with selected markers (things like `classList.toggle('chsp-active-structure-marker', false)`) to a separate `HTMLMarker` class doing all of this. 

On a related note, `src/map/map.ts` is getting quite large, so we should try to extract functionalities and potentially move it out of this file.

ceriottm:
Re (3) I think that some of that could be done as part of the settings PR. All of those settings-related code could go in a separate file, would be tidier and easier to navigate.

Luthaf:
- [ ] Use a `addPinned()` function instead of using setter. As far as I understand, a new pinned is currently automagically created when doing `map.active = guid` with a new guid.

ceriottm:
Isn't this to be closed? Would be good to clean up a bit the open issues, many seem obsolete

Luthaf:
There is still a remaining point from the initial comment.

PR #50, opened 2020-06-19
Feat/pin refactor

rosecers:
Refactoring the pinning infrastructure with respect to #45 

- [x] Do not call select({-1, -1}, guid) to indicate removal
- [ ] instead add specific functions for each possible action: 
    - [ ] addPinned 
    - [x] removePinned
    - [x] changeActivePinned. 
- [x] Then select always act on the current active
- [x] only move the marker/change the structure
- [x] Remove the starterGUID parameter for PropertyMap constructor to allow using the map without a structure viewer. This should be easy to do after the above changes
- [ ] Extract code dealing with selected markers (things like classList.toggle('chsp-active-structure-marker', false)) to a separate HTMLMarker class doing all of this. 
- [ ] On a related note, src/map/map.ts is getting quite large, so we should try to extract functionalities and potentially move it out of this file.

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -329,7 +329,9 @@ export class PropertiesMap {
         assert(markerData !== undefined);
 
         /// Sets the active marker on this map
-        if (markerData.current === undefined || indexes.environment !== markerData.current ) {
+        /// default of markerData.current is -1 so that currents of value 0 are
+        /// still updated
+        if (markerData.current < 0 || indexes.environment !== markerData.current ) {
             markerData.current = indexes.environment;
             this._updateSelectedMarker(selectedGUID, markerData);
         }
@@ -1439,7 +1441,7 @@ export class PropertiesMap {
      * @param  addedGUID unique string identifier of the marker to add
      * @param  index     numeric index of the structure (with respect to dataset) to show
      */
-    private _addMarker(addedGUID: string, index: number = 0): void {
+    private _addMarker(addedGUID: string, index: number= -1): void {
         if (!this._selected.has(addedGUID)) {
             const activeButton = getByID(`chsp-activate-${addedGUID}`);
             const marker = document.createElement('div');
@@ -1454,7 +1456,7 @@ export class PropertiesMap {
             marker.style.backgroundColor = color;
             const newMarkerData = {
                                       color : color,
-                                      current: Math.max(0, index),
+                                      current: index,
                                       marker: marker,
                                   };
             this._selected.set(addedGUID, newMarkerData);
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -199,8 +199,10 @@ interface MarkerData {
  */
 export class PropertiesMap {
     /** Callback fired when the plot is clicked and a new point is selected */
-    public onselect: (indexes: Indexes, selectedGUID?: string) => void;
+    public onselect: (indexes: Indexes) => void;
+    /** Callback fired when a guid is removed from the map */
     public onremove: (removedGUID: string) => void;
+    /** Callback fired when a new active marker is designated */
     public activate: (activeGUID: string) => void;
 
     /**
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -201,6 +201,7 @@ export class PropertiesMap {
     /** Callback fired when the plot is clicked and a new point is selected */
     public onselect: (indexes: Indexes, selectedGUID?: string) => void;
     public onremove: (removedGUID: string) => void;
+    public activate: (activeGUID: string) => void;
 
     /**
      * Callback to get the initial positioning of the settings modal.
@@ -263,6 +264,7 @@ export class PropertiesMap {
         this._indexer = indexer;
         this.onselect = () => {};
         this.onremove = () => {};
+        this.activate = () => {};
         this._selected = new Map();
 
         this._root = getByID(id);
@@ -382,7 +384,7 @@ export class PropertiesMap {
         assert(markerData !== undefined);
 
         const indexes = this._indexer.from_environment(markerData.current);
-        this.onselect(indexes, this._active);
+        this.activate(this.active);
     }
 
     /**
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -953,7 +953,7 @@ export class PropertiesMap {
                 this._indexer.from_environment(event.points[0].pointNumber);
 
             this.select(indexes, this._active);
-            this.onselect(indexes, this._active);
+            this.onselect(indexes);
         });
         this._plot.on('plotly_afterplot', () => this._afterplot());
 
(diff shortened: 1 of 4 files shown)


rosecers:
@Luthaf can you elaborate on a few of these things? 

- What are the purpose of `addPinned` and `changeActivePinned` beyond `addWidget` and `set active`?
- We already only act on the active widget. What is different that is being requested?

src/utils/index.ts
```diff
@@ -5,7 +5,7 @@
 
 import {makeDraggable} from './draggable';
 import {DisplayMode, EnvironmentIndexer, Indexes} from './indexer';
-import {foreachSetting, HTMLSetting, SettingGroup, SettingModificationOrigin} from './settings';
+import {foreachSetting, HTMLSetting, SettingGroup, SettingModificationOrigin, Validator} from './settings';
```
Luthaf, line 8: Why is this required? The corresponding code does not exists on master

rosecers: On master, npm was throwing a compiler error for this not existing

Luthaf: Could to check again? Master is building fine without this: https://github.com/cosmo-epfl/chemiscope/runs/788274793

rosecers: Seems that the error wasn't recreated, so there might have been an artifact from creating the branch. I'll remove this line from the PR.

rosecers:
> > What are the purpose of addPinned and changeActivePinned beyond addWidget and set active?
> 
> Mostly renaming to be consistent: `widget` does not make sense from the map point of view, hence the `pinned`; and having both `this.active = '...'` and `this.removeActive('...')` is strange, that's why I suggested to replace `set active()` with `changeActivePinned`.
> 
> > We already only act on the active widget. What is different that is being requested?
> 
> Removing the `activeGUID` parameter to 'onselect' and friends, which could be used to act on another widget.

So in other words -- onselect would always point to the active, but there would be a separate communicating function to change the active?

Luthaf:
Exactly =)

rosecers:
I think I'd like to leave the remaining issues for an additional PR, this is already getting somewhat big.

src/map/map.ts
```diff
@@ -200,6 +200,8 @@ interface MarkerData {
 export class PropertiesMap {
     /** Callback fired when the plot is clicked and a new point is selected */
     public onselect: (indexes: Indexes, selectedGUID?: string) => void;
+    public onremove: (removedGUID: string) => void;
```
Luthaf, line 203: This needs documentation

src/map/map.ts
```diff
@@ -200,6 +200,8 @@ interface MarkerData {
 export class PropertiesMap {
     /** Callback fired when the plot is clicked and a new point is selected */
     public onselect: (indexes: Indexes, selectedGUID?: string) => void;
```
Luthaf, line 202: Should the `selectedGUID` be removed here?

src/map/map.ts
```diff
@@ -304,38 +298,40 @@ export class PropertiesMap {
     }
 
     /** Change the selected environment to the one with the given `indexes` */
-    public select(indexes: Indexes, selectedGUID: string = this._active) {
+    public select(indexes: Indexes, selectedGUID?: string) {
```
Luthaf, line 301: This needs documentation for `selectedGUID`

src/map/map.ts
```diff
@@ -346,6 +342,14 @@ export class PropertiesMap {
         this._settingsModal.remove();
     }
 
+    public get active() {
```
Luthaf, line 345:
```suggestion
    public get active(): string {
```

This needs documentation

src/map/map.ts
```diff
@@ -346,6 +342,14 @@ export class PropertiesMap {
...
+      if (this._selected.size > 0) {
+        if (this._active === undefined) {
+        this.active = this._selected.keys().next().value;
+        return this._selected.keys().next().value;
+        } else { return this._active; }
```
Luthaf, line 350: minor nitpick: I prefer having the statements for the else case on a separate line, I feel it is easier to parse quickly when looking at the code

src/map/map.ts
```diff
@@ -386,9 +392,40 @@ export class PropertiesMap {
...
+        this.activate(this._active);
     }
 
+    /**
+     * Removes a marker from the map.
+     *
+     * The parameter deleteFromGUIDs pertains to removing the guid from the guid
+     * list. This is *not* the same as the `force` parameter in
+     * `structure._removeWidget`. One may want to remove a marker when switching
+     * from 2D --> 3D but retain the information.
```
Luthaf, line 404: I am not entirely happy with this: the function seems to do both hiding of HTML markers when switching from 2D to 3D; and marker removal. Could these two actions be separated easily?

src/map/map.ts
```diff
@@ -1111,9 +1149,9 @@ export class PropertiesMap {
             // default to 0 (i.e. circles)
             return this._selectTrace<string | string[]>('circle', 'circle', trace);
         }
-
+        assert(this._selected.size > 0);
```
Luthaf, line 1152: why is this assertion needed here?

src/structure/structure.ts
```diff
@@ -83,7 +83,9 @@ interface WidgetGridData {
  */
 export class StructureViewer {
     /** Callback used when the user select an environment */
-    public onselect: (indexes: Indexes, selectedGUID?: string) => void;
+    public onselect: (indexes: Indexes) => void;
+    public onremove: (removedGUID: string) => void;
```
Luthaf, line 87: needs documentation

src/structure/structure.ts
```diff
@@ -169,7 +173,8 @@ export class StructureViewer {
         root.appendChild(this._root);
 
         this._setupGrid(1);
-        this._active = this._viewers.keys().next().value;
+        this.active = this._viewers.keys().next().value;
```
Luthaf, line 176: Should we have an `assert(this._viewer.size > 0)` here?

Luthaf: Or even extract this into a function `_firstPinned(): string`?

src/structure/structure.ts
```diff
@@ -292,6 +296,34 @@ export class StructureViewer {
         }
     }
 
+    /*
+     * Removes a widget from the structure viewer grid.
+     * The parameter force pertains to removing the *only* widget in the grid,
+     * which should only be done when changing datasets.
```
Luthaf, line 302: Since we are no longer changing datasets, should we remove this `force` parameter?

src/structure/structure.ts
```diff
@@ -326,44 +358,40 @@ export class StructureViewer {
...
-        assert(data !== undefined);
-        let indexes;
-        if (this._indexer.mode === 'structure') {
-            indexes = this._indexer.from_structure_atom(data.current.structure);
-        } else {
-            const structure = data.current.structure;
-            const atom = data.current.atom;
-            indexes = this._indexer.from_structure_atom(structure, atom);
+        // choose *if* over assertion -- there's a weird error that I cannot pin-
+        // point where active is being passed an undefined.
```
Luthaf, line 363: I'll try to give this a look, it might come from the constructor calling `this.active = '...'` before filling in `this.widgets`

PR #57, opened 2020-07-03
Do not include source maps in production mode

Luthaf:
They take ~9mb of additional space, making chemiscope.min.js 10.6Mb. After this PR, the minified build is back to 1.6Mb.


Luthaf:
Ping @ceriottm, thanks for noticing that something was wrong here =)

ceriottm:
Can I just merge it? Seems to be doing the right thing, and it does load much faster

PR #58, opened 2020-07-06
More refactoring of the pin functionality

### LINKED ISSUE #56: Symbols get messed up with multiple selections
Using symbols to display categorical data interferes with selection highlighting, particularly when using multiple structure panels.

Expected behavior:
Highlighted structures are shown using the same symbol used for the category, which is just made a bit bigger and colored according to the structure panel color key. 

Observed behavior:
in 2D, a circle is used regardless of the underlying class
in 3D (even more problematic) the symbol of the active panel is used for all the selected structures (so that a structure with a square symbol might turn into a cross if I select a cross structure as the active one)

Might be linked to #50  and #11

Luthaf:
Fix #56 

Continue working on #45: all of the communication between map & viewer grid now happen through public method & callback (no `getById`); and there are separated callback/methods for 

- adding a new pinned environment; 
- removing a pinned environment; 
- changing which pinned environment is the active one
- changing which environment is showed by the active viewer

All these methods exists on the viewer side, while the map only exposes the "change which env is showed" & "change the active viewer" since that's all the user can do with the map.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -24,7 +24,7 @@ import {MetadataPanel} from './metadata';
 import {StructureViewer} from './structure';
 
 import {Dataset, JsObject, Structure, validateDataset} from './dataset';
-import {addWarningHandler, EnvironmentIndexer} from './utils';
+import {addWarningHandler, EnvironmentIndexer, getNextColor} from './utils';
 
 // tslint:disable-next-line: no-var-requires
 require('./static/chemiscope.css');
@@ -121,6 +121,7 @@ class DefaultVisualizer {
 
         this.meta = new MetadataPanel(config.meta, dataset.meta);
 
+        // Structure viewer setup
         this.structure = new StructureViewer(
             config.structure,
             config.j2sPath,
@@ -129,25 +130,43 @@ class DefaultVisualizer {
             dataset.environments,
         );
 
-        this.map = new PropertiesMap(
-            config.map,
-            this._indexer,
-            dataset.properties,
-        );
-
-        this.map.active = this.structure.active;
-
         if (config.loadStructure !== undefined) {
             this.structure.loadStructure = config.loadStructure;
         }
 
-        this.structure.onselect = (indexes) => {
-            this.map.select(indexes);
-            if (indexes.structure > 0 && indexes.environment > 0) {
-              this.info.show(indexes);
-            }
+        this.structure.activeChanged = (guid, indexes) => {
+            this.map.setActive(guid);
+            this.info.show(indexes);
         };
 
+        this.structure.onselect = (indexes) => {
+            this.map.select(indexes);
+            this.info.show(indexes);
+        };
+
+        this.structure.onremove = (guid) => {
+            this.map.removeMarker(guid);
+        };
+
+        this.structure.oncreate = (guid, color, indexes) => {
+            this.map.addMarker(guid, color, indexes);
+            this.info.show(indexes);
+        };
+
+        // map setup
+        this.map = new PropertiesMap(config.map, this._indexer, dataset.properties);
+
+        this.map.onselect = (indexes) => {
+            this.info.show(indexes);
+            this.structure.show(indexes);
+        };
+
+        this.map.activeChanged = (guid, indexes) => {
+            this.info.show(indexes);
+            this.structure.setActive(guid);
+        };
+
+        // information table & slider setup
         this.info = new EnvironmentInfo(config.info, dataset.properties, this._indexer);
         this.info.onchange = (indexes) => {
             this.map.select(indexes);
@@ -156,31 +175,10 @@ class DefaultVisualizer {
         this.info.startStructurePlayback = (advance) => this.structure.structurePlayback(advance);
         this.info.startAtomPlayback = (advance) => this.structure.atomPlayback(advance);
 
-        this.map.onselect = (indexes) => {
-            this.info.show(indexes);
-            this.structure.show(indexes);
-        };
-
-        this.structure.onremove = (guid) => {
-            this.map.removeMarker(guid);
-        };
-
-        this.map.onremove = (guid) => {
-            this.structure.removeWidget(guid);
-        };
-
-        this.structure.activate = (guid) => {
-            this.map.active = guid;
-        };
-
-        this.map.activate = (guid) => {
-            this.structure.active = guid;
-        };
-
         const initial = {environment: 0, structure: 0, atom: 0};
+        this.map.addMarker(this.structure.active, getNextColor([]), initial);
         this.structure.show(initial);
         this.info.show(initial);
-        this.map.select(initial);
     }
 
     /**
--- a/src/index.ts
+++ b/src/index.ts
@@ -161,20 +161,20 @@ class DefaultVisualizer {
             this.structure.show(indexes);
         };
 
-        this.structure.onremove = (removedGUID) => {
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -129,58 +130,55 @@ class DefaultVisualizer {
...
-        );
-
-        this.map.active = this.structure.active;
-
         if (config.loadStructure !== undefined) {
             this.structure.loadStructure = config.loadStructure;
         }
 
+        this.structure.activeChanged = (guid, indexes) => {
+            this.map.setActive(guid);
```
rosecers, line 138: Is it kosher to set up these actions wrt the map before initializing said map?

src/map/map.ts
```diff
@@ -198,12 +201,15 @@ interface MarkerData {
  * plot markers.
  */
 export class PropertiesMap {
-    /** Callback fired when the plot is clicked and a new point is selected */
+    /** Callback fired when the plot is clicked and the position of the active marker changes */
     public onselect: (indexes: Indexes) => void;
-    /** Callback fired when a guid is removed from the map */
-    public onremove: (removedGUID: string) => void;
```
rosecers, line 204: Why have you removed the `onremove` call? Now under 3D the deleted widgets still remain as markers on the 3D map, so I'm assuming the removal of widgets is no longer effectively communica
Screen Shot 2020-07-06 at 11 23 36 AM
ted.

rosecers: Also, after the marker is not removed, the activating of markers on the right does not correspond with the markers 
Screen Shot 2020-07-06 at 11 25 27 AM
on the right.

src/map/map.ts
```diff
@@ -347,92 +313,114 @@ export class PropertiesMap {
...
-     * @param  activeGUID the GUID of the new active viewer
+     * @param guid the GUID of the new active viewer
      */
-    public set active(activeGUID: string) {
-        if (this._active === activeGUID) {
-            // guard against infinite recursion
-            // `Structure set active => StructureViewer.show => Map set active
-            // => PropertiesMap.select => Structure set active`
-            return;
+    public setActive(guid: GUID) {
```
rosecers, line 343: Why did you change from the `set active` style?

src/map/map.ts
```diff
@@ -347,92 +313,114 @@ export class PropertiesMap {
...
-            this._active = activeGUID;
-            newButton.classList.toggle('chsp-active-structure', true);
+    /**
+     * Add a new marker to the map. The new marker is set as the active one
+     *
+     * @param guid GUID of the new marker
+     * @param indexes indexes of the environment that the new marker should show
+     */
+    public addMarker(guid: GUID, color: string, indexes: Indexes): void {
+        assert(!this._selected.has(guid));
```
rosecers, line 367: Why do we need this assertion?

src/index.ts
Luthaf: Since the callback will not be called before everything is initialized, I don't see this as a problem. I find the code easier to follow if everything related to the structure is close together

src/map/map.ts
Luthaf: > Why have you removed the onremove call?

Because the map can never initiate marker removal, and all the callbacks are used to transmit actions from the users on a given panel to all others panels. 

I'll look into the bugs you've found!

src/map/map.ts
Luthaf: For two reason: we sometimes want to use `this._active = '...'` and sometimes `this.active = '...'`. Having a setter instead of a function make the code confusing, since you have to remember the difference. It also allow to follow the same style as the rest of the actions on the map/structure grid: `map.setActive(...)`, `map.select(...)`, etc.

src/map/map.ts
Luthaf: To ensure we don't add the same marker multiple times

src/index.ts
rosecers: I don't disagree, but it is a bit disconcerting to write operations on things that don't yet exist.

src/map/map.ts
rosecers: I'm often more in favor of quitting the function than throwing an error, but that's most likely a style choice.

src/map/map.ts
rosecers: I strongly think we should retain a shadow functionality to remove from the map, it may be something we want to have in the future (happy to discuss in the future)

src/map/map.ts
Luthaf: The issue with quitting the function is that it can hide programming errors since it fails silently. I try to use [defensive programming](https://en.wikipedia.org/wiki/Defensive_programming) and fail as early as possible to catch control flow errors & other issues

src/map/map.ts
Luthaf: > shadow functionality to remove from the map,

What do you mean by shadow functionality?

Luthaf: And if this is something we may want to have in the future, then we can implement it in the future =) But since there is no use case for it now, I would rather keep the code as simple as possible, since that makes future modification easier

Luthaf: Regarding the bugs, the `Map.removeMarker` was missing `if (this._is3D()) {this._updateAll3DMarkers();}`, everything should work now.

ceriottm:
Possibly related to this effort: in the master, if you change the selected structure on the structure panel, the info panel is not updated.

Luthaf:
> Possibly related to this effort: in the master, if you change the selected structure on the structure panel, the info panel is not updated. Might relate to Rose's question on whether the link should be one-sided or not.

That's fixed with this branch, as a side effect of separating "change the active viewer" from "change what the active viewer shows"

----

I am not sure I understand what you mean with one-sided relationship here: that only the structure can create/remove pinned viewer? That is by design since there is no UI to add/remove markers on the map side. Or are you referring to something else?

ceriottm:
> > Possibly related to this effort: in the master, if you change the selected structure on the structure panel, the info panel is not updated. Might relate to Rose's question on whether the link should be one-sided or not.
> 
> That's fixed with this branch, as a side effect of separating "change the active viewer" from "change what the active viewer shows"
> 
> I am not sure I understand what you mean with one-sided relationship here: that only the structure can create/remove pinned viewer? That is by design since there is no UI to add/remove markers on the map side. Or are you referring to something else?

OK great - no I didn't undertand the issue either apparently so if this fixes it, all good and many thumbs up

Luthaf:
> such as using setActive instead of set active, even though I get why

I kinda like seeing entity objects in the Smalltalk sense, i.e. objects are actors that communicate with one another by passing messages (calling methods from other objects, or reacting with callbacks). 

This is also related to the entity class vs value class distinction (I've mostly seen this distinction in C++ land): entity class are full actors, that maintain internal invariant and communicate with methods. Value classes are bags of data that mostly don't have any invariant and 'just' store multiple fields together. For classes with value semantics, getter/setters make more sense, since they make the code look like an actual bag of value, even if we need some internal checks.

The easiest way to see if a class have value or entity semantics is to check if two instances with the same members values have the same identity. Two `StructureGrid` would be different and behave differently here; but two `strings` with the same content would be completely interchangeable.

A bit of reading on the subject if you are interested =)

https://akrzemi1.wordpress.com/2012/02/03/value-semantics/
https://softwareengineering.stackexchange.com/questions/408540/value-reference-type-object-and-semantics

Issue #52, opened 2020-06-23
Reverse size property in the map


Luthaf:
When something like energy is used as a property for dot size in the map, the most interesting points are the one with lowest energy, but they end up being the smallest.

We could provide an option to use `-<property>` instead of `<property>` for the size to support this use case.

ceriottm:
I think we can use a "scale" option, that describes the relation between
the property and the radius. we could have "sqrt" (that should be default),
linear, log, -sqrt, etc.
Similarly, one should also be able to adjust the range.

Luthaf:
> Similarly, one should also be able to adjust the range.

Do you means, setting min & max radius value?

ceriottm:
I'm thinking to have size considered as an axis. So it's mapped between the
property range and sz_min -> sz_max, using a mapping function, very much
like an axis indeed.

Luthaf:
Would that mean we also provide different mapping (sqrt, -x, …) for the x/y/z/color axis too?

ceriottm:
If it's easy to achieve in plotly, I don't see why not. Wouldn't lose hours
otherwise

PR #61, opened 2020-07-09
Feat/custom scales

rosecers:
In response to #52, adds a functionality to the size options for different size scaling. Current functionalities include log scaling, inverse scaling (is really reverse scaling), and sqrt scaling.

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -7,7 +7,7 @@ import assert from 'assert';
 
 import {Property} from '../dataset';
 import {EnvironmentIndexer, HTMLSetting, Indexes, PositioningCallback, SettingGroup, SettingModificationOrigin} from '../utils';
-import {foreachSetting, generateGUID, getByID, makeDraggable, sendWarning} from '../utils';
+import {foreachSetting, getByID, makeDraggable, sendWarning} from '../utils';
 
 import Plotly from './plotly/plotly-scatter';
 import {Config, Data, Layout, PlotlyScatterElement} from './plotly/plotly-scatter';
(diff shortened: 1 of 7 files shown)


Luthaf:
The code looks good overall, but it is going to introduce conflicts with #36, which is almost done 😐 

From the initial issue, what's your opinion on 
> Would that mean we also provide different mapping (sqrt, -x, …) for the x/y/z/color axis too?

rosecers:
How would it need to change wrt #36?
Coloring could potentially be pretty easy too

Luthaf:
I refactored the settings code there, and added validation to settings values

rosecers:
@Luthaf I merged in the newest version of json_settings, let me know if this works for you

src/map/map.ts
```diff
@@ -1140,9 +1140,23 @@ export class PropertiesMap {
             const defaultSize = this._is3D() ? 2000 : 150;
             values = sizes.map((v: number) => {
                 // normalize between 0 and 1, then scale by the user provided value
-                const scaled = userFactor * (v + 0.05 - min) / (max - min);
+                let scaled;
+                switch (scaleMode) {
```
Luthaf, line 942: Instead of switching inside the `map` callback; if might be more efficient to use different callback for different modes:

```ts
switch (scaleMode) {
    case 'inverse':
        values = sizes.map((v: number) => (max - v + 0.05) / (max - min));
    case 'log':
        values = sizes.map((v: number) => Math.log((max - v + 0.05) / (max - min)));
...
}
```

I am not sure how much of a difference this will be in practice though

rosecers: I actually prefer how it's currently written so that you don't have to repeat `                return defaultSize * scaled * scaled * userFactor * userFactor;` for every single one.

Luthaf: fine with me =)

rosecers:
@Luthaf then should be good to go once #36 is merged

PR #62, opened 2020-07-09
Update depdencies and move ploty & markdown-it to dev dependencies

Luthaf:
Since we are bundling everything in chemiscope.min.js, users of chemiscope do not need to install these


Luthaf:
🤣 package-lock.json updates be like:

Capture d’écran 2020-07-09 à 18 32 37

Luthaf:
Let me merge this and create a 0.2.1 release with it + the size reduction from #57

PR #70, opened 2020-07-21
Feat/custom scales2

rosecers:
In response to #52, adds a functionality to the size options for different size scaling. Current functionalities include log scaling, inverse scaling (is really reverse scaling), and sqrt scaling.

I had to re-make the PR because of the branching from master.

### DIFF
--- a/docs/src/tutorial.rst
+++ b/docs/src/tutorial.rst
@@ -286,7 +286,7 @@ values:
                     // if false, markers scale from smallest to largest property value
                     // if true, marker scale from largest to smallest property value
                     // in the case of `inverse` scaling, this is reversed.
-                    "reversed": false | true,
+                    "reverse": false | true,
                 },
             },
             // Settings related to the structure viewers grid. This is an array
--- a/docs/src/tutorial.rst
+++ b/docs/src/tutorial.rst
@@ -274,10 +274,12 @@ values:
                     // scaling factor for the axis, between 1 and 100
                     "factor": 55,
                     // mode to scale the markers with respect to the properties
-                    "scale": "constant" | "inverse" | "linear" | "log" | "sqrt",
+                    "mode": "constant" | "inverse" | "linear" | "log" | "sqrt",
                     // name of the property to use for the markers size, this
                     // must be one of the key from the root `properties` table.
                     "property": "<name>",
+                    // whether to reverse the marker scaling
+                    "reversed": false | true,
                 },
             },
             // Settings related to the structure viewers grid. This is an array
--- a/docs/src/tutorial.rst
+++ b/docs/src/tutorial.rst
@@ -271,11 +271,13 @@ values:
                 "palette": "cividis",
                 // settings related to the markers sizes
                 "size": {
+                    // scaling factor for the axis, between 1 and 100
+                    "factor": 55,
+                    // mode to scale the markers with respect to the properties
+                    "scale": "constant" | "inverse" | "linear" | "log" | "sqrt",
                     // name of the property to use for the markers size, this
                     // must be one of the key from the root `properties` table.
                     "property": "<name>",
-                    // scaling factor for the axis, between 1 and 100
-                    "factor": 55,
                 },
             },
             // Settings related to the structure viewers grid. This is an array
--- a/docs/src/tutorial.rst
+++ b/docs/src/tutorial.rst
@@ -274,11 +274,18 @@ values:
                     // scaling factor for the axis, between 1 and 100
                     "factor": 55,
                     // mode to scale the markers with respect to the properties
-                    "mode": "constant" | "inverse" | "linear" | "log" | "sqrt",
+                      // `constant`: all markers are same size, scaled by `factor`
+                      // `linear`: markers are directly proportional to the property
+                      // `log`: markers are proportional to the logarithm of the property
+                      // `sqrt`: markers are proportional to the square root of the property
+                      // `inverse`: markers are inversely proportional to the property
+                    "mode": "constant" | "linear" | "log" | "sqrt | "inverse"",
                     // name of the property to use for the markers size, this
                     // must be one of the key from the root `properties` table.
                     "property": "<name>",
-                    // whether to reverse the marker scaling
+                    // if false, markers scale from smallest to largest property value
+                    // if true, marker scale from largest to smallest property value
+                    // in the case of `inverse` scaling, this is reversed.
                     "reversed": false | true,
                 },
             },
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -336,9 +336,7 @@ export class PropertiesMap {
             data.marker.style.display = 'none';
         } else {
             this._active = guid;
-            const factor = this._options.size.factor.value;
-            const scaleMode = this._options.size.mode.value;
-            this._restyle({'marker.size': this._sizes(1)} as Data, 1);
+            this._updateSelectedMarker(data);
         }
 
         this.setActive(guid);
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -612,8 +612,9 @@ export class PropertiesMap {
         this._options.size.mode.onchange = () => {
             if (this._options.size.mode.value !== 'constant') {
               this._options.size.property.enable();
-              // hacky default to first plotted parameter
-              this._options.size.property.value = this._options.x.property.value;
+              if (this._options.size.property.value === '') {
+                this._options.size.property.value = this._options.x.property.value;
+              }
             } else {
               this._options.size.property.value = '';
               this._options.size.property.disable();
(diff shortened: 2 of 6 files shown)


src/map/map.ts
```diff
@@ -335,7 +335,10 @@ export class PropertiesMap {
         if (this._is3D()) {
             data.marker.style.display = 'none';
         } else {
-            this._updateSelectedMarker(data);
+            this._active = guid;
+            const factor = this._options.size.factor.value;
+            const scaleMode = this._options.size.mode.value;
+            this._restyle({'marker.size': this._sizes(1)} as Data, 1);
```
Luthaf, line 341: I don't understand why this is required in `addMarker`, could you explain?

src/map/options.ts
```diff
@@ -106,6 +108,7 @@ export class MapOptions extends OptionsGroup {
                 throw Error(`size factor must be between 0 and 100, got ${value}`);
             }
         };
+        this.size.mode.validate = optionValidator(['', 'linear', 'log', 'sqrt', 'inverse'], 'size');
```
Luthaf, line 111: why default to `''` instead of `linear`?

rosecers: I just added a push which designates "constant" as the marker size default, which is different than "linear"

src/map/options.html
```diff
@@ -176,11 +176,28 @@ <h5 class="chsp-settings-section-title">Style</h5>
...
+                                <label class="input-group-text" for=chsp-size-mode>scale:</label>
+                            </div>
+                            <select id=chsp-size-mode class="custom-select" disabled>
+                                <option value="linear">linear</option>
+                                <option value="log">log</option>
+                                <option value="sqrt">sqrt</option>
+                                <option value="inverse">inverse</option>
+                            </select>
+                      </div>
+                      <div class="input-group input-group-sm">
```
Luthaf, line 196:
```suggestion
                      <div class="input-group input-group-sm" style="grid-column: auto / span 2">
```

to ensure the settings look like this:

```
|scale      ^|   | factor ---------------+--------------|
```

instead of the current
```
|scale      ^|   | factor --+-----|  <BLANK>
```

src/map/options.ts
Luthaf: We already use `this.size.property` set to an empty string for constant size, this feels a bit redundant. Do you have a use case for having both options?

src/map/map.ts
rosecers: There's a weird bug where if I don't _restyle, any new added markers get placed in the far bottom left of the graph

src/map/options.ts
rosecers: wouldn't it make more sense to disable this.size.property in case of constant scaling versus the other way around?

Luthaf: yep, that makes more sense!

src/map/map.ts
Luthaf: And calling `this._updateSelectedMarker(data);` do not work? I can try to give this a look once you are done with this PR

Luthaf:
All the changes in here also needs to be added to the settings documentation in https://chemiscope.org/docs/tutorial.html#input-file-structure

src/map/options.ts
rosecers: Let me know what you think of the changes

src/map/map.ts
```diff
@@ -603,12 +604,25 @@ export class PropertiesMap {
...
             this._restyle({ 'marker.size': this._sizes(0) } as Data, 0);
         };
 
-        this._options.size.property.onchange = sizeChange;
-        this._options.size.factor.onchange = sizeChange;
+        this._options.size.mode.onchange = () => {
+            if (this._options.size.mode.value !== 'constant') {
+              this._options.size.property.enable();
+              // hacky default to first plotted parameter
+              this._options.size.property.value = this._options.x.property.value;
```
Luthaf, line 616: Why do we need to change `this._options.size.property.value`? Could we just keep it as when enabling/disabling? 

And since `this._options.size.property.value = ''` is no longer a valid value, how about changing the validation & initial value directly? Using the same property as x axis works for me!

rosecers: As a user, I'd want to see some immediate feedback on the plot when switching size modes -- this won't happen if we don't designate some property. _However_, I should write it only to set as x axis if we're going from constant to not constant. Will patch this morning.

Luthaf: > As a user, I'd want to see some immediate feedback on the plot when switching size modes -- this won't happen if we don't designate some property.

Completely agree with you. My point was more about removing the possibility of property being and empty string, by changing the validator & initial value in MapOptions, removing the need for changing it here

rosecers: So I just pushed a few changes. One non-trivial one is I added an enabled() and disabled() function to the HTMLOptions to check if an element is enabled. That way, for this and as well as others, we don't have to rely on values being empty strings to check if something is enabled.

src/map/map.ts
```diff
@@ -915,7 +914,7 @@ export class PropertiesMap {
         const userFactor = logSlider(this._options.size.factor.value);
 
         let values;
-        if (this._options.size.property.value !== '') {
+        if (this._options.size.property.value !== 'constant') {
```
Luthaf, line 917: `this._options.size.mode.value`?

rosecers: Yes. Fixed and committed

src/map/map.ts
Luthaf: 👍

docs/src/tutorial.rst
```diff
@@ -271,11 +271,13 @@ values:
                 "palette": "cividis",
                 // settings related to the markers sizes
                 "size": {
+                    // scaling factor for the axis, between 1 and 100
+                    "factor": 55,
+                    // mode to scale the markers with respect to the properties
```
Luthaf, line 276: Could you add a few words describing what each option does? constant, inverse and linear might not be 100% evident.

side note, in French inverse means 1/x, not -x. Is it the same in English? If so, I would propose `reverse` for this mode

src/map/map.ts
```diff
@@ -335,6 +335,7 @@ export class PropertiesMap {
         if (this._is3D()) {
             data.marker.style.display = 'none';
         } else {
+            this._active = guid;
```
Luthaf, line 338: It is strange that this is required, `_updateSelectedMarker` should be able to update even non-active markers. I'll have a quick look.

src/map/map.ts
```diff
@@ -603,12 +604,25 @@ export class PropertiesMap {
...
             this._restyle({ 'marker.size': this._sizes(0) } as Data, 0);
         };
 
-        this._options.size.property.onchange = sizeChange;
-        this._options.size.factor.onchange = sizeChange;
+        this._options.size.mode.onchange = () => {
+            if (this._options.size.mode.value !== 'constant') {
+              if (this._options.size.property.disabled()) {
+                this._options.size.property.enable();
+                this._options.size.property.value = this._options.x.property.value;
```
Luthaf, line 616: this should no longer be required

src/map/options.ts
```diff
@@ -100,9 +100,9 @@ export class MapOptions extends OptionsGroup {
         this.size = {
             factor : new HTMLOption('number', 50),
             mode : new HTMLOption('string', 'constant'),
-            property : new HTMLOption('string', ''),
+            property : new HTMLOption('string', propertiesName[0]),
         };
-        this.size.property.validate = optionValidator(propertiesName.concat(['']), 'size');
+        this.size.property.validate = optionValidator(propertiesName, 'size');
```
Luthaf, line 105: This is breaking the Arginine-Dipeptide example, as well as any files saved with `size: {property: ''}` in the settings (in particular the KPCovR ones), which is a bit unfortunate 😐 

The best fix I see is to catch the errors when applying settings & convert them to warnings, but it would be a bit noisy. Alternatively, we could manually deal with this case for backward compatibility.

docs/src/tutorial.rst
rosecers: Another non-trivial fix: I added a "reverse size" checkbox so that any size-scaling mode can be reversed

docs/src/tutorial.rst
```diff
@@ -274,10 +274,12 @@ values:
                     // scaling factor for the axis, between 1 and 100
                     "factor": 55,
                     // mode to scale the markers with respect to the properties
-                    "scale": "constant" | "inverse" | "linear" | "log" | "sqrt",
+                    "mode": "constant" | "inverse" | "linear" | "log" | "sqrt",
                     // name of the property to use for the markers size, this
                     // must be one of the key from the root `properties` table.
                     "property": "<name>",
+                    // whether to reverse the marker scaling
+                    "reversed": false | true,
```
Luthaf, line 282: the property being called `reverse`, the setting should also be called `reverse`

src/map/options.ts
Luthaf: Fixed in 99ca9c9

Issue #68, opened 2020-07-21
Loader menu in standalone viewer


ceriottm:
Previously the standalone script would by default hide the loader when a data file was included in the html. 
Now (1) the hiding code doesn't work anymore and (2) I question whether it makes sense to hide it, given that being able to save visualization state is useful. This issue is to track (1) and discuss on (2)

PR #71, opened 2020-07-22
Remove leftover flag from generate_standalone

Luthaf:
The Python code also went through a round of black formatting


Luthaf:
Fixes #68

Issue #60, opened 2020-07-08
Extract selected markers from map


rosecers:
Extract code dealing with selected markers (things like classList.toggle('chsp-active-structure-marker', false)) to a separate HTMLMarker class doing all of this.

ceriottm:
I think we can close this. map.ts and viewer.ts would benefit a lot from being broken down into smaller files but probably by now there are different dimensions along which to clean up.

PR #75, opened 2020-07-22
Cleanup/reduce map

rosecers:
This is a PR focused on re-allocating code in src/map/map.ts to other modules where possible to reduce the size of src/map/map.ts (currently ~1200 lines). This is in response to #60 but also ongoing conversation.

First moves:
Making a class `src/marker/marker.ts` where all marker functions can live. Difficulty is that many marker functions rely on the map._plot aspect, so this only reduced `map.ts` by about 70 lines.
Moving _option-only dependent functions to src/map/options.ts. This was more fruitful (~150 lines), but may cause problems if there are functions which are *this* map specific and options is being used for multiple maps.


src/map/map.ts
```diff
@@ -284,20 +249,26 @@ export class PropertiesMap {
...
+        data.activate();
 
         if (this._is3D()) {
             this._restyle({'marker.size': this._sizes(1)} as Data, 1);
         }
+
+        const activeData = this._selected.get(guid);
+        assert(activeData !== undefined);
+        const activeIndexes = this._indexer.from_environment(activeData.current);
+        this.activeChanged(guid, activeIndexes);
```
Luthaf, line 271: why are you calling `activeChanged` here? 

The goal for this callback is to be called when the user change the active marker by interacting with the map; while the goal of `setActive` is to change the active marker on the map when the user interacted with another panel.

src/map/map.ts
```diff
@@ -1138,84 +1008,53 @@ export class PropertiesMap {
...
-              marker.style.right = `${plotWidth - x}px`;
-          }
+        if (!this._is3D()) {this._updateMarkers(); }
     }
 
     /**
      * Update the position, color & size of all markers in the map
      */
-    private _updateAllMarkers(): void {
+    private _updateMarkers(data: MarkerData[] = Array.from(this._selected.values())): void {
```
Luthaf, line 1031: The doc comment no longer correspond to the functions

src/map/map.ts
```diff
@@ -1138,84 +1008,53 @@ export class PropertiesMap {
...
         if (this._is3D()) {
+            for (const datum of data) {
```
Luthaf, line 1033: could also be `data.forEach((d) => d.toggleVisible(false))` if you want

src/map/marker.ts
```diff
@@ -0,0 +1,75 @@
...
+    public color: string;
+    /// index of the currently displayed structure
+    public current: number;
+    /// Marker indicating the position of the point in 2D mode
+    /// Using an HTML element is much faster than using Plolty to restyle the full plot,
+    /// especially with more than 100k points. In 3D mode, a separate trace is
+    /// used instead.
+    public marker: HTMLElement;
+
+    constructor(
```
Luthaf, line 17: All functions need documentation here

src/map/marker.ts
```diff
@@ -0,0 +1,75 @@
+import assert from 'assert';
+
+import {EnvironmentIndexer, Indexes} from '../indexer';
+
+/// interface to contain synchronized parameters for marker instances
```
Luthaf, line 5:
```suggestion
/// Data associated with markers on the map indicating selected environments/structures
```

src/map/map.ts
rosecers: Yes, yes I do.

Luthaf:
Tests are failling due to formatter checks, you can reformat the code with `npm run pretty`. Please also install a https://prettier.io/ plugin in your editor to automatically format files as you edit them, this will reduce diff noise!

Luthaf:
It's going live! 🚀

PR #76, opened 2020-07-27
Make sure we can scroll on the property table if it does not fit in the left panel

### LINKED ISSUE #64: Property table is missing a scrollbar when there are too many properties
For example when loading the QM9 KPCovR map, and then clicking on the `Strucure XXX` button, there are too many properties to fit in the space occupied by the structure viewer, and no scroll bar to get back to the properties on the top.

This should be a simple fix to set `overflow` CSS property to `scroll` instead of `hidden`.

Luthaf:
Fix #64 

This was much harder than I anticipated, in part because the properties table do not live inside its parent div, but rather on top of the structure one; and the use of embed-responsive CSS as introduced in ebfe170b09c7080e3e4275b86963967374ddd795 to fix the aspect ratio of panels.

### DIFF
--- a/app/static/js/default.js
+++ b/app/static/js/default.js
@@ -111,6 +111,7 @@ function setupChemiscope(json) {
                 };
             };
 
+            updateChspInfoHeight();
             stopLoading();
         })
         .catch((e) =>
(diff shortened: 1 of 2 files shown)


ceriottm:
Looks like it *always* shows scrollbars, even when there's no need

Luthaf:
I'm not sure what you mean here, I don't see scroll bars on FF/Chrome/Safari. There was a bit of additional space below the table that chrome would allow to scroll to (with a hidden scroll bar by default), I remove that.

Issue #21, opened 2020-05-25
Introduce end to end testing


Luthaf:
One of the reviewer concern on the [JOSS paper](https://github.com/openjournals/joss-reviews/issues/2117#issuecomment-624957786) was that we lack automated tests. We can do manual testing, but that mean it can be hard to refactor without introducing bugs for corner cases. At the same time, chemiscope being mostly a GUI is not well suited for standard unit testing (checking public classes/functions one by one, single behavior by single behavior)

I think we should add some form of automated testing, end to end testing might be the less bad one. The idea is to specify high-level behavior such as `selecting a new property for axis x changes the plot & axis title on the plot`. Such tests might be painful to write since we have to make them generic enough to not break due to layout changes but specific enough to catch the issues; but I still think they are worth it.

Cypress looks like a nice framework for such end to end testing: https://www.cypress.io/.

PR #80, opened 2020-08-27
Minimal testing framework

Luthaf:
WIP toward #21 

Minimal basis of testing framework using [chai](https://www.chaijs.com/) (assertions), [mocha](https://mochajs.org/) (unit test framework), and [jsdom](https://github.com/jsdom/jsdom) (mock implementation of the DOM api running on node). 

It looks like we could be able to do quite a bit just with this, since jsdom is giving us "full" DOM access. The only thing that would still benefit from cypress or something similar is changes happening across panels (click to set the new active, etc.)


Luthaf:
This should be ready for review. Once merged, other can chime in to implement tests for existing code, and we can start requiring tests for new code!

src/map/options.ts
```diff
@@ -72,6 +72,8 @@ export class MapOptions extends OptionsGroup {
         reverse: HTMLOption<'boolean'>;
     };
 
+    /// The HTML buttom to open the settings modal
+    private _openSettings: HTMLElement;
```
rosecers, line 76: What is the purpose of this file being included in this PR? These changes look like cleanups.

tests/jsdom.ts
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable */
+const { JSDOM } = require('jsdom');
+
+function copyProps(src: any, target: any) {
```
rosecers, line 4: Can you add some documentation here to point to what this test does?

tests/map/options.test.ts
```diff
@@ -0,0 +1,41 @@
...
+        setupJSDOM();
+    });
+
+    it('validate new values for properties');
+
+    it('creates HTML elements with unique id');
+
+    it('can save settings');
+
+    it('can apply saved settings');
```
rosecers, line 29: I'm not so familiar with this status -- can you explain what the successive `it(...)` statements do?

rosecers (commented):
I don't see anything immediate to change -- most of my comments are questions or general comments.

src/map/options.ts
Luthaf: This is a fix for one of the tests I added, which was failling, namely checking that calling `remove` doe indeed remove everything.

tests/map/options.test.ts
Luthaf: So the way mocha works is that test cases are introduced as calls to the `it` function, with a first argument describing the test case, and the second argument executing the test itself. If there is no second argument, as here, the test is skipped and considered "to be written later". 

I added these mostly as a TODO list for myself, but I can also keep them out of the code for now.

rosecers: It might be best to either omit in this PR or write the coinciding tests

src/map/options.ts
rosecers: Does it belong in this PR or should it be split off?

rosecers:
Is this active?

Luthaf:
Working on it right now =)

tests/map/options.test.ts
Luthaf: Removed, I'll send another PR implementing these tests

tests/jsdom.ts
Luthaf: done!

src/map/options.ts
Luthaf: I this it belongs to here, tests and corresponding fix are in the second commit

Luthaf:
This should be good to go!

Luthaf:
I'll merge this on Monday if nobody disagree!

rosecers (approved):

Issue #79, opened 2020-08-26
Enable Selective Opacity


rosecers:
- [ ] convert all 2D colormaps to RGBA format
- [ ] disable RGBA format for 3D colormaps until plotly can support

rosecers:
@Luthaf I've made it through the first pass on the branch `feat/rgba`. Currently, there are no opacity-affected colormaps, so I included a commented line in `src/maps/colorscales.ts` to uncomment to demonstrate the new functionality. Thoughts on the structure? If it looks good I'll move ahead with opacity modes.

ceriottm:
Why do you say that this wouldn't work in 3D? seems it's enough to set
trace opacity to 0.99999 and then it honors the alpha channel
https://github.com/plotly/plotly.js/issues/3492

Luthaf:
I remember that there where some other issues with transparency in 3D with plotly (related to depth buffer & points appearing on top of others), but we can check if the fix would work for us!

PR #81, opened 2020-08-27
Feat/rgba

rosecers:
Working on the first two points made in #79, making RGB maps to RGBA and converting between RGB and RGBA for 3D/2D, respectively

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -32,6 +32,23 @@ const DEFAULT_LAYOUT = {
             },
             y: 0,
             yanchor: 'bottom',
+            x: 1,
+            xanchor: 'right',
+        },
+        colorscale: [] as Plotly.ColorScale,
+        showscale: false,
+    },
+    coloraxis2: {
+        cmax: 0,
+        cmin: 0,
+        colorbar: {
+            len: 1,
+            thickness: 20,
+            title: {
+                text: '' as undefined | string,
+            },
+            y: 0,
+            yanchor: 'bottom',
         },
         colorscale: [] as Plotly.ColorScale,
         showscale: true,
@@ -467,10 +484,10 @@ export class PropertiesMap {
                 this._options.color.max.value = max;
 
                 this._relayout(({
-                    'coloraxis.colorbar.title.text': this._title(
+                    'coloraxis2.colorbar.title.text': this._title(
                         this._options.color.property.value
                     ),
-                    'coloraxis.showscale': true,
+                    'coloraxis2.showscale': true,
                 } as unknown) as Layout);
             } else {
                 this._options.color.disable();
@@ -480,8 +497,8 @@ export class PropertiesMap {
                 this._options.color.max.value = 0;
 
                 this._relayout(({
-                    'coloraxis.colorbar.title.text': undefined,
-                    'coloraxis.showscale': false,
+                    'coloraxis2.colorbar.title.text': undefined,
+                    'coloraxis2.showscale': false,
                 } as unknown) as Layout);
             }
 
@@ -499,13 +516,23 @@ export class PropertiesMap {
             const max = this._options.color.max.value;
             this._relayout(({
                 'coloraxis.cmax': max,
-                'coloraxis.cmin': min,
+                'coloraxis.cmin': min - (max - min),
                 // looks like changing only 'coloraxis.cmax'/'coloraxis.cmin' do
                 // not update the color of the points (although it does change
                 // the colorbar). Asking for an update of 'coloraxis.colorscale'
                 // seems to do the trick. This is possiblely a Ploty bug, we
                 // would need to investiguate a bit more.
                 'coloraxis.colorscale': this._options.colorScale(),
+            } as unknown) as Layout);
+            this._relayout(({
+                'coloraxis2.cmax': max,
+                'coloraxis2.cmin': min,
+                // looks like changing only 'coloraxis.cmax'/'coloraxis.cmin' do
+                // not update the color of the points (although it does change
+                // the colorbar). Asking for an update of 'coloraxis.colorscale'
+                // seems to do the trick. This is possiblely a Ploty bug, we
+                // would need to investiguate a bit more.
+                'coloraxis2.colorscale': this._options.colorBarScale(),
             } as unknown) as Layout);
         };
         this._options.color.min.onchange = colorRangeChange;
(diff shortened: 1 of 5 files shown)


Luthaf:
The code seems to work fine, we just have to decide what we want here!

I would invert main & background traces to have them ordered as `[background, main, selected]` instead of the current `[main, background, selected]` to ensure the points in main appear on top of the ones in background.

I would also introduce named constants for traces, i.e. `const MAIN_TRACE = 1; const BACKGROUND_TRACE = 0; const ALL_TRACES = [0, 1, 2]` and so on.

----

Regarding styling, I don't think we want for the user to be able to set min/max opacity, instead I would go with fixed values of 0.3 for min and 1 for max. 

On the top of this, we might want to try setting the background points color to grey instead of the main color, it might help to make selected points pop a bit more (potentially as an option? like this is done in the viewer's environment options).

src/map/map.ts
```diff
@@ -747,70 +896,53 @@ export class PropertiesMap {
...
+            return this._filter<string>(['rgba(1, 1, 1, 0.3)'], trace, ['black']);
         }
     }
 
     /**
      * Get the values to use as marker size with the given plotly `trace`, or
      * all of them if `trace === undefined`.
      */
-    private _sizes(trace?: number): Array<number | number[]> {
+    private _sizes(trace?: number): Array<number> | Array<number[]> {
```
Luthaf, line 945: nitpick: that should be equivalent to the previous code

src/map/map.ts
```diff
@@ -1119,4 +1239,72 @@ export class PropertiesMap {
         }
         return check;
     }
+
+    // this applies the opacity filter to the points, saving the points
+    // in the "main" trace (i.e. those meeting the filter criterion) in
+    // this._mainPoints, and putting all other points in this._backgroundPoints
+    private _updateFilter() {
```
Luthaf, line 1246: naming: how about `_updateGrouping`/`_updateBackground`?

src/map/map.ts
```diff
@@ -1119,4 +1239,72 @@ export class PropertiesMap {
...
+            }
+        }
+
+        this._mainPoints = mP;
+        this._backgroundPoints = bP;
+    }
+
+    // this is a *new* version of _selectTrace which not only returns the objects
+    // for the specific trace, but avoids unnecessary computation
+    private _filter<T>(objectsToFilter: T[], trace?: number, selectedObjects?: T[]): T[] | T[][] {
```
Luthaf, line 1284: naming: `_selectTrace` should still work, `_group` might also work but is it both a noun and a verb, `_filterTrace` sounds good as well

it could be called as `this._filterTrace(xxx, BACKGROUND_TRACE)`

src/map/map.ts
```diff
@@ -1119,4 +1239,72 @@ export class PropertiesMap {
...
+        let mainObjects: T[] = objectsToFilter;
```
Luthaf, line 1285: naming: `mainObjects` => `main`, `objectsToFilter` => `all`, `backgroundObjects` => `background`

src/map/map.ts
```diff
@@ -154,6 +154,12 @@ export class PropertiesMap {
     private _options: MapOptions;
     /// Button used to reset the range of color axis
     private _colorReset: HTMLButtonElement;
+    /// List of points rendered at lowest opacity.
+    /// Stored so that they only need be calculated whenever the filter changes
+    private _backgroundPoints: Array<number>;
```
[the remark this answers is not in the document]
rosecers: To be honest I feel exactly opposite

src/map/map.ts
```diff
@@ -747,70 +896,53 @@ export class PropertiesMap {
...
-        const selected = [];
-        for (const marker of this._selected.values()) {
-            if (this._is3D()) {
-                selected.push(values[marker.current]);
-            } else {
-                selected.push(NaN);
-            }
+            return [NaN];
+        } else {
+            return this._property(axis.property.value).values;
```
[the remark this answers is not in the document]
rosecers: There needs to be a way to get all coordinates when plotting the HTML markers in lines 1185-1186. We could add a function for `_filteredCoordinates` or `_unfilteredCoordinates` if you think that would be easier

src/map/map.ts
rosecers: all feels too vague for me. I wanted to be specific to say objects --> mainObjects, etc

rosecers:
> The code seems to work fine, we just have to decide what we want here!
> 
> I would invert main & background traces to have them ordered as `[background, main, selected]` instead of the current `[main, background, selected]` to ensure the points in main appear on top of the ones in background.
> 
> I would also introduce named constants for traces, i.e. `const MAIN_TRACE = 1; const BACKGROUND_TRACE = 0; const ALL_TRACES = [0, 1, 2]` and so on.
> 
> Regarding styling, I don't think we want for the user to be able to set min/max opacity, instead I would go with fixed values of 0.3 for min and 1 for max.
> 
> On the top of this, we might want to try setting the background points color to grey instead of the main color, it might help to make selected points pop a bit more (potentially as an option? like this is done in the viewer's environment options).

I handled the first part of this in the most recent commit.

I disagree on the second part, I think users should be able to decide on opacity, especially considering it is so easy to do so.

src/map/map.ts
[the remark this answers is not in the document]
Luthaf: Or use `this._property(axis.property.value).values` directly since it is only needed in one place (potentially directly in `_computeRSCoord`?)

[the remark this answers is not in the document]
rosecers: Agreed. Updated in latest commit!

ceriottm:
I saw there's lots of setting coordinates to NaN around. TBH couldn't carefully inspect the whole thing but I feel very strongly that we need to steer away from this NaN trick, not add more of it.

rosecers:
> I saw there's lots of setting coordinates to NaN around. TBH couldn't carefully inspect the whole thing but I feel very strongly that we need to steer away from this NaN trick, not add more of it.

Can you explain? The only _new_ introductions of NaN's come in initializing the background trace, where it is empty upon initialization. I modeled it after the other empty traces added for symbols and such. No other NaN's listed in the code are introduced by this PR.

Luthaf:
We ought to be able to initialise both additional traces with empty traces (`x: []` instead of `x: [NaN]`), especially now that the map supports having 0 selected/highlighted points. I remember having an issue with empty trace at some point, but we might be able to workaround it.

src/map/map.ts
```diff
@@ -1044,10 +1048,13 @@ export class PropertiesMap {
...
-        this._restyle(({
-            'marker.symbol': symbols,
-            type: 'scatter3d',
-        } as unknown) as Data);
+        this._restyle(
+            ({
+                'marker.symbol': symbols,
+                type: 'scatter3d',
+            } as unknown) as Data,
+            [0, 1, 2]
```
Luthaf: We don't want to only change the trace type and symbol for traces 0, 1, 2, we want to change it for all traces, including the ones used to display the symbol legend. I'm having the same issue in a separate branch, so I'll look for a fix.

rosecers:
@Luthaf Should I resurrect this?

Luthaf:
At some point, yes! If you rebase it on top of the current master, I'll can have a new look to it

PR #82, opened 2020-08-28
Fix id replace in structure viewer options

Luthaf:
HTML formatting by prettier added unexpected double quotes around the id.

Extracted from #80 since it is useful by itself.

### DIFF
--- a/src/structure/options.html
+++ b/src/structure/options.html
@@ -88,8 +88,9 @@
                                 for="env-center"
                                 title="center automatically structure or environment in the viewport"
                                 style="cursor: help;"
-                                >center</label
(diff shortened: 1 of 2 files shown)


Luthaf:
Without this I get errors like "Error: unable to get element with id chsp-8f70a20e-a181-4ce5-a26f-8b2e93d8ecc5-atom-labels", since the regex do not match the id.

And yes, I know I should not be parsing HTML with a regex: https://stackoverflow.com/a/1732454/4692076

Luthaf:
Merging this since https://chemiscope.org/ is currently broken ...

PR #86, opened 2020-09-10
Do not create a point at NaN-NaN in the selected trace

Luthaf:
It is no longer needed now that we support having an empty selected trace. We still require the NaN in the symbol traces, I've added a comment to explain why.

Ping @ceriottm 

---

This makes me a bit sad, NaN are very tasty!
Naan

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -572,40 +572,30 @@ export class PropertiesMap {
     private _createPlot() {
         this._plot.innerHTML = '';
 
-        const colors = this._colors();
-        const lineColors = this._lineColors();
-        const sizes = this._sizes();
-        const symbols = this._symbols();
-
-        const x = this._coordinates(this._options.x);
-        const y = this._coordinates(this._options.y);
-        const z = this._coordinates(this._options.z);
-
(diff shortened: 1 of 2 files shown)


src/map/map.ts
```diff
@@ -572,40 +572,30 @@ export class PropertiesMap {
...
                     width: 1,
                 },
                 // prevent plolty from messing with opacity when doing bubble
                 // style charts (different sizes for each point)
                 opacity: 1,
-                size: sizes[0],
+                size: this._sizes(0)[0],
                 sizemode: 'area',
-                symbol: symbols[0],
+                symbol: this._symbols(0)[0],
```
Luthaf, line 598: @rosecers The core of the issue with `symbolFunc is not a function` was that the interface of Plotly.createPlot & Plotly.restyle are different. The first one takes a list of traces, containing `x: [1, 2, 3, 4]`; while the second one takes a list of elements to change, containing one array for each trace `x: [[1, 2, 3, 4], [...], [...]]`. Trying to use `this._symbols(0)` in `_createPlot` results in `symbols: ['circle']`, but we want `symbols: 'circle'`.

Let me know if I'm unclear

rosecers (approved):

PR #88, opened 2020-10-07
Add properties metadata

### LINKED ISSUE #20: Add properties metadata
Adding `description` and `units` for all properties in the dataset would be good

Places where we can display them:
- use `name (unit)` on the axis of the map (just `name` if unit is undefined or empty)
- show a tooltip with the `description` when selecting properties in the map setting? Not sure if this is possible
- show description in a tooltip in the "info" panel
- show unit in the "info" panel

Do you see other metadata we would want to attach to the properties?

cbenmahm:
fix #20, allow adding units and description (optional) to properties and show them in the map and the info table

### DIFF
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -125,6 +125,10 @@ export interface Property {
      * or "C"); and numeric values should be use for everything else.
      */
     values: string[] | number[];
+    /** string containing the description of the property */
+    description?: string;
+    /** string containing the units of the property */
+    units?: string;
 }
 
 /**
@@ -322,6 +326,15 @@ function checkProperties(
             if (typeof value !== initial) {
                 throw Error(`'properties['${key}'].values' should be of a single type`);
             }
+        }
+
+        // check the meta is valid
+        if ('description' in property && typeof property.description !== 'string') {
+            throw Error(`'properties['${key}'].description' should contain string`);
+        }
+
+        if ('units' in property && typeof property.units !== 'string') {
+            throw Error(`'properties['${key}'].units' should contain string`);
         }
     }
 }
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -125,9 +125,9 @@ export interface Property {
      * or "C"); and numeric values should be use for everything else.
      */
     values: string[] | number[];
-    /** string containing the description of the property */
+    /** user-facing description of the property */
     description?: string;
-    /** string containing the units of the property */
+    /** unit of the property values */
     units?: string;
 }
 
@@ -328,13 +328,13 @@ function checkProperties(
             }
         }
 
-        // check the meta is valid
+        // check that units & description are valid
         if ('description' in property && typeof property.description !== 'string') {
-            throw Error(`'properties['${key}'].description' should contain string`);
+            throw Error(`'properties['${key}'].description' should contain a string`);
         }
 
         if ('units' in property && typeof property.units !== 'string') {
-            throw Error(`'properties['${key}'].units' should contain string`);
+            throw Error(`'properties['${key}'].units' should contain a string`);
         }
     }
 }
--- a/src/info/table.ts
+++ b/src/info/table.ts
@@ -55,7 +55,23 @@ export class Table {
         for (const name in properties) {
             const tr = document.createElement('tr');
             const td = document.createElement('td');
-            td.innerText = name;
+
+            //  add the units to the property if it exists, this is identical to the _title in ../map/map.ts
+            const units = properties[name].units;
+            let title = name;
+            if (units !== undefined) {
+                title += ` [${units}]`;
+            }
+            td.innerText = title;
+
+            // add a tooltip containing the description of the property and underline if it exists
+            const description = properties[name].description;
+            if (description !== undefined) {
+                td.style.borderBottom = '1px dotted #00f';
+                td.style.display = 'inline';
+                td.title = description;
+            }
+
             tr.appendChild(td);
             const cell = document.createElement('td');
             tr.appendChild(cell);
--- a/src/info/table.ts
+++ b/src/info/table.ts
@@ -60,7 +60,7 @@ export class Table {
             const units = properties[name].units;
             let title = name;
             if (units !== undefined) {
-                title += ` [${units}]`;
+                title += `/${units}`;
             }
             td.innerText = title;
 
--- a/src/info/table.ts
+++ b/src/info/table.ts
@@ -68,6 +68,7 @@ export class Table {
             const description = properties[name].description;
             if (description !== undefined) {
                 td.style.borderBottom = '1px dotted #00f';
+                td.style.cursor = "help"
                 td.style.display = 'inline';
                 td.title = description;
             }
--- a/src/info/table.ts
+++ b/src/info/table.ts
@@ -68,7 +68,7 @@ export class Table {
             const description = properties[name].description;
             if (description !== undefined) {
                 td.style.borderBottom = '1px dotted #00f';
-                td.style.cursor = "help"
+                td.style.cursor = 'help';
                 td.style.display = 'inline';
                 td.title = description;
             }
--- a/src/map/data.ts
+++ b/src/map/data.ts
@@ -1,3 +1,5 @@
+import { getUnpackedSettings } from 'http2';
+import { PropertiesMap } from '..';
 /**
  * @packageDocumentation
  * @module map
@@ -14,6 +16,8 @@ export interface NumericProperty {
     values: number[];
     /** string interner if the property was a string-valued property */
     string?: StringInterner;
+    /** add units as optinal*/
+    units?: string;
 }
 
 /** @hidden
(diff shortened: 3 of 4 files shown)


Luthaf:
FYI, if you write `fix #20` in the first comment, this will automatically close the issue when the PR is merged. https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

src/map/map.ts
```diff
@@ -678,13 +678,13 @@ export class PropertiesMap {
...
+        layout.xaxis.title = this._title(this._options.x.property.value);
+        layout.yaxis.title = this._title(this._options.y.property.value);
         layout.xaxis.type = this._options.x.scale.value;
         layout.yaxis.type = this._options.y.scale.value;
-        layout.scene.xaxis.title = this._options.x.property.value;
-        layout.scene.yaxis.title = this._options.y.property.value;
-        layout.scene.zaxis.title = this._options.z.property.value;
+        layout.scene.xaxis.title = this._title(this._options.x.property.value);
+        layout.scene.yaxis.title = this._title(this._options.y.property.value);
+        layout.scene.zaxis.title = this._options.z.property.value; // using this._title(this._options.z.property.value) here seems to cause troubles
```
Luthaf, line 687: I guess the issue is that `this._options.z.property.value` can be set to the empty string `''` when in 2D mode, and that `this._title` does not account for it.

src/dataset.ts
```diff
@@ -125,6 +125,10 @@ export interface Property {
      * or "C"); and numeric values should be use for everything else.
      */
     values: string[] | number[];
+    /** string containing the description of the property */
+    description?: string;
+    /** string containing the units of the property */
+    units?: string;
```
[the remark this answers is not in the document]
rosecers: I would go with `units` over `unit`

src/map/map.ts
```diff
@@ -774,6 +774,16 @@ export class PropertiesMap {
...
+    /** Prepare a title of an axis with potential units */
+    private _title(name: string): string{
+        let title = name;
+        let units = this._property(title).units;
+            if (units !== undefined){
+                title += ` [${units}]`;
+            }
+        return title;
+    }
+
```
rosecers, line 786:
```suggestion
    /** Prepare a title of an axis with potential units */
    private _title(name: string): string{
        if(this._property(name).units !== undefined) {
            return name + ` [${units}]`;
        } else { return name; }
    }
    
```

Just to make it more compact

Luthaf: Single line else are not part of our prettier config ^_^

But why not something like

```ts
private _title(name: string): string{
    const units = this._property(name).units;
    if (units !== undefined) {
        return name + ` [${units}]`;
    } else {
        return name;
    }
}
```

rosecers: Yah, that would probably work best... 

Side note: `prettier` is wrong, single line `else`s are awesome. But alas.

Luthaf: > Side note: prettier is wrong, single line elses are awesome. But alas.

Only if you are trying to win [IOCCC](https://www.ioccc.org/) and write code like [this](https://www.ioccc.org/2019/adamovsky/prog.c) 😁

rosecers:
```suggestion
    private _title(name: string): string{
        if (name !== '') {
            const units = this._property(name).units;
            if (units !== undefined) {
                return name + ` [${units}]`;
            } 
        }
        return name
    }
    
```

What about adding the catchall so that we can use `this._title` for the `zaxis` for consistency?

src/map/map.ts
```diff
@@ -774,6 +774,16 @@ export class PropertiesMap {
...
     }
 
+    private _title(name: string): string{
+        if (name !== '') {
+            const units = this._property(name).units;
+            if (units !== undefined) {
+                return name + ` [${units}]`;
+            } 
+        }
+        return name
```
rosecers, line 784:
```suggestion
        return name;
```

src/map/map.ts
rosecers:
```suggestion
        layout.scene.zaxis.title = this._title(this._options.z.property.value);
```

given that this._title should handle this, can we implement as with x/y axis?

rosecers (changes requested):
I'm good once the z-axis title is resolved.

Luthaf:
Also TBD before merging this:

- [ ] Add a visual indication that the description is available by hovering over the name in the properties panel
- [ ] Update the example files with unit & description
- [ ] rebase/squash unnecessary commits. I can help with this if you want!

src/info/table.ts
```diff
@@ -55,7 +55,23 @@ export class Table {
...
+            let title = name;
+            if (units !== undefined) {
+                title += ` [${units}]`;
+            }
+            td.innerText = title;
+
+            // add a tooltip containing the description of the property and underline if it exists
+            const description = properties[name].description;
+            if (description !== undefined) {
+                td.style.borderBottom = '1px dotted #00f';
```
Luthaf, line 70: you can set the cursor type here, with `td.style.cursor = "help";`

src/map/data.ts
```diff
@@ -54,16 +58,18 @@ export class StringInterner {
 }
 
 /** Transform a property to a numeric property */
-function propertyToNumeric(name: string, property: number[] | string[]): NumericProperty {
-    const prop_type = typeof property[0];
+function propertyToNumeric(name: string, property: Property): NumericProperty {
+    //console.log(name, property.units);
```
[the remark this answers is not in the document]
Luthaf: This is still TBD =)

src/map/map.ts
Luthaf: This is still TBD

Luthaf (commented):
Looks good to me, only three points still need a bit of work!

Luthaf:
@cbenmahm updated the example files to include description & unit, and they are now online!

src/map/map.ts
```diff
@@ -678,13 +678,13 @@ export class PropertiesMap {
...
         layout.yaxis.type = this._options.y.scale.value;
-        layout.scene.xaxis.title = this._options.x.property.value;
-        layout.scene.yaxis.title = this._options.y.property.value;
-        layout.scene.zaxis.title = this._options.z.property.value;
+        layout.scene.xaxis.title = this._title(this._options.x.property.value);
+        layout.scene.yaxis.title = this._title(this._options.y.property.value);
+        layout.scene.zaxis.title = this._options.z.property.value; // using this._title(this._options.z.property.value) here seems to cause troubles
         layout.coloraxis.colorscale = this._options.colorScale();
         layout.coloraxis.cmin = this._options.color.min.value;
         layout.coloraxis.cmax = this._options.color.max.value;
```
Luthaf, line 690: We may also want to display the unit in the color bar title, so below in `layout.coloraxis.colorbar.title.text`; and the in callbacks. What do you think?

ceriottm:
FYI, I'm liking a lot the ISO standard that prescribes that units in axes
should be indicated as quantity/unit (e.g. volume/angstrom3).
Incidentally, do we have support for subscripts/superscripts?

Luthaf:
> Incidentally, do we have support for subscripts/superscripts?

We don't interpret the "unit" string at all, so `A^3` will not be rendered as `A³`. However, we pass unicode characters through without touching them, so users can use [unicode superscript](https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts)

ceriottm:
can plotly format labels using latex? I think that it might be easier for
people (and similar to what you do in matplotlib) than having them type
unicode characters.

Luthaf:
TIL that plotly supports latex: https://plotly.com/javascript/LaTeX/. So it should work fine for axis titles, and maybe colorbar title. The remaining issue concerns the other places where we would like to output units, i.e. the property table and maybe the `<select>` in options. There we would have to render LaTeX manually.

I don't know if we can piggy-back on Plotly's LaTeX support or if we need to include another dependency such as KaTeX to enable this, I'll have a quick look.

Another thing is that if we start supporting LaTeX in units, why not support it in `meta.references`, `meta.description` and friends? Overall, I think that LaTeX support might be best left to a future PR.

Luthaf:
So looks like plotly is using with MathJax for latex rendering but do not included it by default. This means we could render latex manually/automagically by setting up MathJax and maybe KaTeX manually ourself.

However, Latex support inside the plots seems a bit hit or miss: https://github.com/plotly/plotly.js/issues/4605 and https://github.com/plotly/plotly.js/issues/608. In particular, LaTex rendering does not work at all for 3D plots, and would be very hard to fix on the plotly side.

Given that this seems non-trivial to implement and ensure that it keeps working, I think we should leave latex support for another PR.

ceriottm:
sounds good, but would be a very nice feature to have - perhaps could you
open a feature issue for that?

src/map/map.ts
cbenmahm: I think it may look a bit crowded but it is added for now

src/map/map.ts
```diff
@@ -774,6 +776,16 @@ export class PropertiesMap {
         return this._selectTrace<number[]>(values, selected, trace);
     }
 
+    private _title(name: string): string {
+        if (name !== '') {
+            const units = this._property(name).units;
+            if (units !== undefined) {
+                return name + ` [${units}]`;
```
Luthaf, line 783: should this be `${name)/${units}` instead, per https://github.com/cosmo-epfl/chemiscope/pull/88#issuecomment-716748562?

Luthaf (approved):
Looks good to me =) I would squash commits when merging if that's fine with everyone!

src/map/map.ts
cbenmahm: I missed that comment. Let's do that!

Luthaf:
Thanks a lot for implementing this @cbenmahm! And sorry for the delay in review

PR #89, opened 2020-10-08
Changed the filename for CSD

rosecers:
The app currently points to the wrong CSD filename. Small, one line fix.


Luthaf:
The file on the website is called `CSD-100R`, cf https://github.com/cosmo-epfl/chemiscope/tree/gh-pages. However the name changed recently-ish (as part of ef6058984224c94468867670a7e4f2690974487a), so you may have an old version of it locally. You should be able to update by running `npm run download-example-input` again.

PR #90, opened 2020-10-26
Update dependencies

Luthaf:
The main change is the switch to a new major version of webpack. Everything should be working fine, although I would appreciate a second set of eyes on this.

@cbenmahm, do you think you could give it a look? Checkout the PR locally, removing your `node_modules` and checking that nothing broke?


cbenmahm:
This looks good to me, on macos with an updated version of the command-line tools

PR #93, opened 2020-11-13
Update documentation with properties metadata

Luthaf:
Follow up for #88

### DIFF
--- a/docs/src/tutorial.rst
+++ b/docs/src/tutorial.rst
@@ -135,9 +135,9 @@ structures.
 
 If all the properties you want to include into chemiscope are already stored
 in an  `ase`_-readable file, you can also use the `chemiscope_input`_ script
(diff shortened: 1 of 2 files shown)


docs/src/tutorial.rst
```diff
@@ -176,15 +176,20 @@ values:
 
         // list of properties in this dataset
         "properties": {
-            // each property have a name, a target and some values
+            // each property have at least a name, a target and some values
```
cbenmahm, line 179: I would be more thorough here. For example, each property must have at least a name, a target, and some values. Optional entries for the units and descriptions can be added to the property.

cbenmahm (approved):

docs/src/tutorial.rst
Luthaf: Sounds good! I've added this

Issue #38, opened 2020-06-07
Sparse environment list


ceriottm:
Using NaNs to indicate environments that are not linked to points on the map is cumbersome and inefficient. The indexing is already allowing for a sparse list, it only needs to be handled on the JSMol side. Also related to #4

PR #96, opened 2020-11-17
Enable installing the python helpers with pip

### LINKED ISSUE #94: install, and easy import of write_chemiscope_input utils
I find often myself using ugly code such as 

```
from os.path import expanduser
sys.path.append(expanduser('~')+'/lavoro/code/chemiscope/utils')
from chemiscope_input import write_chemiscope_input
```

to be able to write chemiscopes from my analysis notebooks.
I think it would be nice to have a pip package that installs the "utils" section of the chemiscope package, to facilitate its usage.
I was kind of torn as to whether these utils should rather live in a separate repo, but I actually think it makes sense for them to be associated with chemiscope, as that will simplify ensuring that the utils and the actual viewer stay in sync.

Luthaf:
Fix #94 

One unresolved question is what should be the name of the package. I went with `chemiscope`, but this might create a bit of confusion with the npm/javascript chemiscope package. An alternative would be `chemiscope-input`, but then if/when we create a jupyter plugin for chemiscope it would have to live in a different package.


Luthaf:
@rosecers do you think you could give this a look?

ceriottm:
A few thoughts. 
1. This is important, and an under-developed part of chemiscope. We should make it as easy as possible to generate chemiscopes as part of a ML analysis workflow. As far as I understand, currently the workflow goes through creating an ASE frames that store the various properties - it is unclear to me if this allows exploiting fully the features of chemiscope, see e.g. the definition of units or property descriptions
2. Something like chemiscope-python, for the package name ?

Luthaf:
> As far as I understand, currently the workflow goes through creating an ASE frames that store the various properties 

That's only true for the command line version. The function supports an `extra` argument (should probably be renamed to `properties`) to specify arbitrary properties, including the newly added description and units. I've also started to decouple these scripts from ASE to allow for other structure/properties backends.

> We should make it as easy as possible to generate chemiscopes as part of a ML analysis workflow

Completely agree, but I'm not sure we want to tackle this in this repository, and definitively not in this PR =)

This was the goal of our initial idea with @rosecers for the chemistack proposal, and seems to be related to what I understand M-stack would be. The issue here is that there are a lot of different ways one could want to create the map part of chemiscope. While at COSMO we use SOAP/librascal + and a few different mapping algorithms, there is no reason to marry chemiscope with these specific descriptors and mapping algorithms if we want it to be used by the widest variety researchers.

We could create a prototype chemistack easily using notebooks, @rosecers already has one prototype using matplotlib to display the map. This also links into the FlowingSOAP idea at materials cloud. Overall I think we should have a meeting/brainstorming session to decide what we want to do here.

There is also the ASAP tool to so something like this, so we could also help and work with them instead of re-inventing our own wheel.

> Something like chemiscope-python, for the package name ?

I think I prefer having the same name in pip and in code (so `pip install XXX` and then `from XXX import`), and `from chemiscope_python import write_input` feels a bit strange. Overall I'm not opposed to `chemiscope` as a name for the Python package as well, do you think there could be confusion with the npm package with the same name?

ceriottm:
> > We should make it as easy as possible to generate chemiscopes as part of a ML analysis workflow
> 
> Completely agree, but I'm not sure we want to tackle this in this repository, and definitively not in this PR =)

This PR I agree, this repo we should. I don't mean that the analysis should be part of this repo, but that we should make converting the output of the analysis into a chemiscope as seamless as possible.  So M-stack is a separate discussion, here I just wanted to stress that the process of getting an xyz and an analysis output into a chemiscope could be easier.
 
> > Something like chemiscope-python, for the package name ?
> 
> I think I prefer having the same name in pip and in code (so `pip install XXX` and then `from XXX import`), and `from chemiscope_python import write_input` feels a bit strange. Overall I'm not opposed to `chemiscope` as a name for the Python package as well, do you think there could be confusion with the npm package with the same name?

My thinking is probably less informed than yours. To me it's not the end of the world to have a pip chemiscope that differs from npm chemiscope but perhaps someone better informed has discussed the case and come up with good practices?

Luthaf:
>  I don't mean that the analysis should be part of this repo, but that we should make converting the output of the analysis into a chemiscope as seamless as possible. So M-stack is a separate discussion, here I just wanted to stress that the process of getting an xyz and an analysis output into a chemiscope could be easier.

Oh, right 😄. Completely agree with this, do you have specific pain points that could be improved? Or an ideal workflow that we could try to enable? 

By analysis output, do you mean a in-memory numpy array, or some data save on the disk?

ceriottm:
> Oh, right 😄. Completely agree with this, do you have specific pain
> points that could be improved? Or an ideal workflow that we could try to
> enable?
>
> By analysis output, do you mean a in-memory numpy array, or some data save
> on the disk?
>
I think ideally we should keep the dual workflow with a function that is
mirrored by a command-line utility. People from "my generation" still have
an instinct to go full bash onto postprocessing.
So I think we want to be able to easily combine structures (here having
something that can be read by ASE or an Atoms list seem to cover quite some
grounds) and arrays of values (that maps easily into column files) or
dicts.
One thing that often bugs me is that I want to drop info from the ASE file
so there could also be a switch that allows you to drop those fields.

I don't know if you get the idea or want to discuss a more formal API/set
of use cases.

python/chemiscope/adapters.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def frames_to_json(frames):
+    frames = list(frames)
+
+    if HAVE_ASE and isinstance(frames[0], ase.Atoms):
+        return [_ase_to_json(frame) for frame in frames]
+    else:
+        raise Exception(f"unknown frame type: '{frames[0].__class__.__name__}'")
+
```
rosecers, line 19: To improve readability, this function should be at the end of the file.

python/chemiscope/adapters.py
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    for frame in frames[1:]:
+        for name, value in frame.arrays.items():
+            if name in IGNORED_ASE_ARRAYS:
+                continue
+            properties[name]["values"] = np.concatenate(
+                [properties[name]["values"], value]
+            )
+    return properties
```
rosecers, line 70: I don't love this instantiation. If there are frames which don't have a full set of properties, then we'll end up with later frames' properties being associated with the wrong frame. Also causing an error: frames with a property that doesn't exist in previous frames. How about:

```suggestion
    properties = {
        name: {"target": "atom", "values": value}
        for name, value in frames[0].arrays.items()
        if name not in IGNORED_ASE_ARRAYS
    }

    for frame in frames[1:]:
        for name, value in frame.arrays.items():
            if name in IGNORED_ASE_ARRAYS:
                continue
            if name not in properties:
                names = list(properties.keys())
                properties[name] = {"target": "atom",  "values": np.nan * np.zeros(properties[names[-1]]['values'].shape)}
        for name in properties.keys():
            value = frame.arrays.get(name, np.nan * np.zeros(len(frame.positions)))
            properties[name]["values"] = np.concatenate(
                [properties[name]["values"], value]
            )
            
    return properties
```

python/chemiscope/adapters.py
```diff
@@ -0,0 +1,82 @@
...
+def _ase_structure_properties(frames):
+    properties = {
+        name: {"target": "structure", "values": []} for name in frames[0].info.keys()
+    }
+
+    for frame in frames:
+        for name, value in frame.info.items():
+            properties[name]["values"].append(value)
+
+    return properties
```
rosecers, line 82: Same issue as above. If a frame is missing a property or is carrying an extra one, there will be an error.
```suggestion
def _ase_structure_properties(frames):
    properties = {
        name: {"target": "structure", "values": []} for name in frames[0].info.keys()
    }

    for frame in frames:
        for name, value in frame.info.items():
            if name in IGNORED_ASE_ARRAYS:
                continue
            if name not in properties:
                names = list(properties.keys())
                properties[name] = {"target": "structure", "values": np.nan * np.zeros(properties[names[-1]]['values'].shape)}

        for name in properties.keys():
                properties[name]["values"].append(frame.info.get(name, np.nan * properties[name]["values"][-1]))

    return properties
```

python/chemiscope/input.py
```diff
@@ -0,0 +1,187 @@
...
+                "target": property["target"],
+                "values": _typetransform(list(property["values"])),
+            }
+        elif len(property["values"].shape) == 2:
+            if property["values"].shape[1] == 1:
+                data[name] = {
+                    "target": property["target"],
+                    "values": _typetransform(list(property["values"])),
+                }
+            else:
```
rosecers, line 50: Will this work for an array of lists?

python/chemiscope/main.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+    group.add_argument(
+        "--only-structures",
+        action="store_true",
+        help="only use per-structure properties from the input file",
+    )
+    parser.add_argument("--name", default="", type=str, help="name of the dataset")
+    parser.add_argument(
+        "--description", default="", type=str, help="description of the dataset"
+    )
```
rosecers, line 61: Could we also make it possible to load the description from a text file? Seems likely that this will be a very long argument.

python/chemiscope/main.py
```diff
@@ -0,0 +1,108 @@
...
+        type=str,
+        default=[],
+        help="list of references for the dataset",
+    )
+    args = parser.parse_args()
+
+    if args.only_atoms and args.cutoff is None:
+        raise Exception("--only-atoms requires to give --cutoff")
+    if args.only_structures and args.cutoff is not None:
+        raise Exception("--only-structure can not be given with --cutoff")
```
rosecers, line 77: Should this be a warning, rather than an exception?

rosecers (approved):
This in general looks good -- there were a few places that I have questions and suggestions, but there's nothing prohibitive.

python/chemiscope/adapters.py
Luthaf: I prefer having the main function on the top of the file, and the specific implementation further down. This way the main part if visible first when opening the file.

python/chemiscope/main.py
Luthaf: The command line script is more for quick & dirty JSON generation. I would rather make people use the python function directly instead if the command line script is not enough for them.

python/chemiscope/main.py
Luthaf: Why? It is bogus, there is no obvious way to recover (should cutoff or only-structure take precedence?), and the user can correct this easily by changing the command line invocation.

python/chemiscope/adapters.py
Luthaf: Make sense, I'll change it!

python/chemiscope/input.py
Luthaf: no, but you have to try pretty hard to create an array of list with numpy, manually setting `dtype=object`. I'll have to check, but I think this will fails with an assert later.

python/chemiscope/adapters.py
Luthaf: Done in 73d60deb65a9a778e1be9b5dea35af7fac3f8f11

python/chemiscope/adapters.py
Luthaf: Done in 73d60deb65a9a778e1be9b5dea35af7fac3f8f11

python/chemiscope/input.py
Luthaf: Yep, this is caught by a check later.

Luthaf:
So I went a bit ahead and added unit tests for most of the python code. This should help to improve it later, but makes this PR a bit large. I'll try not to add anything else to it =)

Luthaf:
Looks like GA is no longer running tests. I'll check this

python/chemiscope/adapters.py
```diff
@@ -14,6 +15,12 @@ def frames_to_json(frames):
 
     if HAVE_ASE and isinstance(frames[0], ase.Atoms):
         return [_ase_to_json(frame) for frame in frames]
+    elif HAVE_ASE and isinstance(frames[0], ase.Atom):
+        raise Exception(
+            "expected ase.Atoms, got ase.Atom. "
+            + "Try passing a list of frames instead of a single frame "
+            + "(`[frame]` instead of `frame`)"
+        )
```
rosecers, line 23: Why would we want this, opposed to reading the frames as `frames = [frame]`?

python/chemiscope/adapters.py
```diff
@@ -54,15 +73,47 @@ def _ase_to_json(frame):
...
+
+        for name in remove:
+            all_names.remove(name)
+            extra.add(name)
+
+    if len(extra) != 0:
+        warnings.warn(
+            "the following atomic properties properties are only defined "
+            + f"for a subset of frames: {list(sorted(extra))}; they will be ignored"
+        )
```
rosecers, line 105: Design choice: omission of partial properties opposed to partial property lists?

I strongly prefer partial property lists, where we might throw a warning saying "xx property found for xx/xx frames". This will avoid a property being omitted across the board if it is missing from one frame.

python/chemiscope/input.py
```diff
@@ -156,32 +92,182 @@ def write_input(path, frames, meta=None, extra=None, cutoff=None):
 
         for key in meta.keys():
             if key not in ["name", "description", "authors", "references"]:
-                warnings.warn("ignoring unexpected metadata: {}".format(key))
+                warnings.warn(f"ignoring unexpected metadata: {key}")
```
rosecers, line 95: Why ignore instead of incorporate?

python/chemiscope/input.py
```diff
@@ -156,32 +92,182 @@ def write_input(path, frames, meta=None, extra=None, cutoff=None):
...
+
     if path.endswith(".gz"):
         with gzip.open(path, "w", 9) as file:
             file.write(json.dumps(data).encode("utf8"))
     else:
         with open(path, "w") as file:
             json.dump(data, file)
+
+
+def _typetransform(data, name):
```
rosecers, line 159: needs doc

python/chemiscope/input.py
```diff
@@ -156,32 +92,182 @@ def write_input(path, frames, meta=None, extra=None, cutoff=None):
...
+                f"unsupported type in property '{name}' values: "
+                + "should be string or number"
+            )
+
+
+def _linearize(name, property, n_structures, n_atoms):
+    """
+    Transform 2D arrays in multiple 1D arrays, converting types to fit json as
+    needed.
+    """
```
rosecers, line 179: needs parameter documentation

python/chemiscope/input.py
```diff
@@ -156,32 +92,182 @@ def write_input(path, frames, meta=None, extra=None, cutoff=None):
...
+            f"unknown type ({type(property['values'])}) for property '{name}'"
+        )
+
+    if "units" in property:
+        for item in data.values():
+            item["units"] = str(property["units"])
+
+    if "description" in property:
+        for i, item in enumerate(data.values()):
+            # add [component XX] to the description if values was a ndarray
```
rosecers, line 217: why?

python/chemiscope/adapters.py
```diff
@@ -0,0 +1,159 @@
...
+
+try:
+    import ase
+
+    HAVE_ASE = True
+except ImportError:
+    HAVE_ASE = False
+
+
+def frames_to_json(frames):
```
rosecers, line 13: needs doc

python/chemiscope/adapters.py
```diff
@@ -0,0 +1,159 @@
...
+        raise Exception(
+            "expected ase.Atoms, got ase.Atom. "
+            + "Try passing a list of frames instead of a single frame "
+            + "(`[frame]` instead of `frame`)"
+        )
+    else:
+        raise Exception(f"unknown frame type: '{frames[0].__class__.__name__}'")
+
+
+def atom_properties(frames):
```
rosecers, line 28: needs doc

python/chemiscope/adapters.py
```diff
@@ -0,0 +1,159 @@
...
+        raise Exception(
+            "expected ase.Atoms, got ase.Atom. "
+            + "Try passing a list of frames instead of a single frame "
+            + "(`[frame]` instead of `frame`)"
+        )
+    else:
+        raise Exception(f"unknown frame type: '{frames[0].__class__.__name__}'")
+
+
+def structure_properties(frames):
```
rosecers, line 43: needs doc

python/chemiscope/adapters.py
```diff
@@ -0,0 +1,159 @@
...
+        raise Exception(
+            "expected ase.Atoms, got ase.Atom. "
+            + "Try passing a list of frames instead of a single frame "
+            + "(`[frame]` instead of `frame`)"
+        )
+    else:
+        raise Exception(f"unknown frame type: '{frames[0].__class__.__name__}'")
+
+
+def _ase_to_json(frame):
```
rosecers, line 58: needs doc

python/chemiscope/adapters.py
```diff
@@ -0,0 +1,159 @@
...
+    data["y"] = [float(value) for value in frame.positions[:, 1]]
+    data["z"] = [float(value) for value in frame.positions[:, 2]]
+
+    if (frame.cell.lengths() != [0.0, 0.0, 0.0]).all():
+        data["cell"] = list(np.concatenate(frame.cell))
+
+    return data
+
+
+IGNORED_ASE_ARRAYS = ["positions", "numbers"]
```
rosecers, line 72: are there any others we should consistently ignore?

python/chemiscope/adapters.py
```diff
@@ -0,0 +1,159 @@
...
+        for name, value in frame.arrays.items():
+            if name not in all_names:
+                continue
+            properties[name]["values"] = np.concatenate(
+                [properties[name]["values"], value]
+            )
+    return properties
+
+
+def _ase_structure_properties(frames):
```
rosecers, line 124: Should we ignore non-numeric properties?

python/chemiscope/adapters.py
Luthaf: You mean manually creating the list ourself if the user passes a single frame? I guess we can do this.

python/chemiscope/input.py
Luthaf: You mean add the keys to the JSON anyway? Because then we can end up with JSON files containing random keys that do not conform to whatever we expect on the javascript side, making our live harder if/when we want to make changes to the format.

python/chemiscope/input.py
Luthaf: To transform something like `PCA of SOAP vector` to `PCA of SOAP vector [component 1]` when associated with the `PCA[1]` property. I don't have strong feeling on this, we can remove it.

python/chemiscope/adapters.py
Luthaf: no, these are the one always created by ASE, anything else represents a real property

python/chemiscope/adapters.py
Luthaf: No, we also support string properties in chemiscope, and boolean properties are mapped to 1/0. We could ignore anything that is not either a string or convertible to float though.

python/chemiscope/adapters.py
Luthaf: Well, chemiscope does not support partial property list for now =). There is the NaN hack, but it is a hack, and I would much rather leave this kind of decision to the user. This is already dealing with a very corner case, and the user gets a warning if something strange is going on, so they can act on it.

---- 

Overall, I would be in favor of **NOT** loading properties from ASE automatically, and instead have the user always provide what they want us to load. We could have an API like this:

```py
from chemiscope import create_input

data = create_input(
    frames = frames,
    properties = {
        "PCA": np.array(),  # automatically guess target
        "something else": {"target": "atom", "values": XXX},  # or allow to specify it
    },
    atom_properties_from_frames = ["name", "other name"],
    structure_properties_from_frames = ["something else"],
    meta={},  # as now
    cutoff=4.5,  # as now
)
```

This should be left for a future PR though.

python/chemiscope/adapters.py
rosecers: might be a good idea to ignore and warn

python/chemiscope/adapters.py
rosecers: Okay, let's do it this way for now and revisit later

Luthaf:
Everything should be addressed!

rosecers (approved):
Looks good to me. On the record, I'd like us to revisit in the future the decision to omit unexpected/incomplete entries.

Luthaf:
> On the record, I'd like us to revisit in the future the decision to omit unexpected/incomplete entries.

As soon as we have support for them (#38), I would agree! See also https://github.com/cosmo-epfl/chemiscope/issues/97#issuecomment-731207436 on this subject.

PR #99, opened 2020-12-02
Adding a small example for creating chemiscope data

rosecers:
Thoughts on a lightweight example for the docs?

### DIFF
--- a/python/chemiscope/input.py
+++ b/python/chemiscope/input.py
@@ -72,6 +72,35 @@ def create_input(frames, meta=None, properties=None, cutoff=None):
     will generate four properties named ``cheese[1]``, ``cheese[2]``,
     ``cheese[3]``,  and ``cheese[4]``, each containing 300 values.
 
+    Here is a quick example of generating a chemiscope input from an `ase.Atoms`_
+    file, a `json`_-type metadatafile, and a text file which contains additional
+    atomic properties as columns, with headers.
+
+    .. code-block:: python
+
+       from ase.io import read
+       import numpy as np
+       import json
+
+       input_file = 'trajectory.xyz'
+       property_file = 'properties.txt'
+       meta_file = 'metadata.json'
+
+       frames = read(input_file, ':')
+       property_values = np.genfromtxt(property_file, names=True)
+       property_dictionary = {
+                              name: {
+                                        'target': 'atom',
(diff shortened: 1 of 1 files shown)


python/chemiscope/input.py
```diff
@@ -72,6 +72,35 @@ def create_input(frames, meta=None, properties=None, cutoff=None):
...
+       from ase.io import read
+       import numpy as np
+       import json
+
+       input_file = 'trajectory.xyz'
+       property_file = 'properties.txt'
+       meta_file = 'metadata.json'
+
+       frames = read(input_file, ':')
+       property_values = np.genfromtxt(property_file, names=True)
```
Luthaf, line 90: I'm not sold on using the very specific `np.genfromtxt` function in an example, and loading data from files in general. 

For this example, we could have an ellispis around how data is loaded, something like 
```py
# example 1: list containing the energy of each structure, from a set of calculations
energy = [ ... ]

# example 2: PCA projection computed using sklearn [LINK].
# X contains a multi-dimensional descriptor of the structure
X = np.array( ... )
pca = sklearn.decomposition.PCA(n_components=3).fit_transform(X)

properties = {
    "PCA": {
        "target": "atom",
        "values": pca,
    },
    "energies": {
        "target": "structure"
        "values": energies,
        "units": "kcal/mol", # maybe leave this one out of the example?
    },
}
```

This also removes the use of a dictionary comprehension, which can be confusing to beginners.

python/chemiscope/input.py
```diff
@@ -72,6 +72,35 @@ def create_input(frames, meta=None, properties=None, cutoff=None):
...
+       property_dictionary = {
+                              name: {
+                                        'target': 'atom',
+                                        'values': property_values[name],
+                                    }
+                              for name in property_values.dtype.names
+                             }
+
+       meta = json.load(open(meta_file, 'r'))
+       c_input = create_input(frames=frames, meta=meta, properties=property_dictionary)
```
Luthaf, line 100: I think the example would be better suited to the `write_input` function, since that's the one people are expected to use.

Luthaf:
Why did you merge this? It had un-addressed comments left

Luthaf:
I'll send a follow-up PR to continue the discussion =)

PR #100, opened 2020-12-03
BUG FIX: only pull color from the 3rd property when _greater_ than 2 …

rosecers:
…properties exist


Luthaf (approved):

Luthaf:
This kind of bug is exactly why we need to add unit tests to this repo! I'll try to resurrect #80 soon

PR #106, opened 2021-01-27
Hot swap molecular renderer from JSmol to 3Dmol.js

### LINKED ISSUE #72: Issue when loading multiple structures rapidely in different viewers
When loading different structures in multiple viewers in the same page, it looks like some global state is not updated on the JSmol side, leading to the wrong structure being loaded in some viewers.

We mostly see this when applying saved settings containing multiple viewers, and slowing down loading seems to fix it:
https://github.com/cosmo-epfl/chemiscope/blob/acdae4d83cac62da0308c5b0d87046306a89c8bb/src/index.ts#L339-L347

We should still go for an actual fix as much as possible since the fix above makes everything very slow, and even 1s delay is not always enough.

Luthaf:
wheel

---

This should be functionally equivalent to the previous molecular viewer, with the added advantages of being faster, resolving #72 and opening possibilities to use chemiscope in more contexts.

The only previously supported setting that is not ported in this PR is the `packed cell` one from JSmol, which we could reproduce with a bit more efforts.

TODO before merging:
- [ ] Manually check that everything works as expected in multiple cases
- [ ] Add tests in the new test framework
- [x] Wait for a release of 3Dmol.js on npm containing my latest PR there (currently this PR is using the git version directly)
- [x] Add back the atom-labels option
- [x] Center camera when switching environment https://github.com/cosmo-epfl/chemiscope/pull/106#issuecomment-772366983

fixes #72


Luthaf:
CI fails because of the git dependency on 3Dmol

ceriottm:
*MC-BREAKING-THINGS*
the "center environment" feature does not work - i.e. when you switch environment it doesn't re-center it as it does with jsmol

ceriottm:
The appearance is kind of uglier than with JSMol. Perhaps just a matter of turning on antialiasing?

ceriottm:
> _MC-BREAKING-THINGS_
> the "center environment" feature does not work - i.e. when you switch environment it doesn't re-center it as it does with jsmol

Actually this works when you switch *structure*, so the problem is only when switching environments

src/structure/widget.ts
```diff
@@ -465,7 +578,11 @@ export class MoleculeViewer {
...
     }
 
-    private _atomSelected(atom: Partial<$3Dmol.AtomSpec>): void {
+    /**
+     * Func
+     * @param atom
+     */
+    private _selectAtom(atom: Partial<$3Dmol.AtomSpec>): void {
         // use atom.serial instead of atom.index to ensure we are getting the
         // id of the atom inside the central cell when using a supercell
```
rosecers, line 587: Is there a reason for using "atom.serial" instead of "guid"? Thinking of consistency of naming.

src/structure/widget.ts
```diff
@@ -648,14 +800,87 @@ export class MoleculeViewer {
         }
     }
 
+    /**
+     * Reset the view by re-centering it and zooming to fit the model as much as
+     * possible inside the vies.
+     *
```
rosecers, line 806: vies --> view

src/structure/3dmol/index.d.ts
```diff
@@ -475,10 +511,16 @@ export interface PresetElementColors {
...
 export declare const elementColors: PresetElementColors;
 
-declare namespace Parsers {
-    function assignBonds(atoms: Array<Partial<AtomSpec>>): void;
-}
-
 export declare class Vector3 {
     constructor(x: number, y: number, z: number);
+    public normalize(): Vector3;
+}
```
rosecers: Why declare a Vector3 class? Often you are using a three vector across the package, why does it necessitate a class here?

src/structure/widget.ts
```diff
@@ -648,14 +800,87 @@ export class MoleculeViewer {
...
     }
+
+    /**
+     * Rotate the viewed group so that the given direction (in group
+     * coordinates) is aligned with the z axis (in camera space)
+     *
+     * @param direction axis to align with the camera view
+     */
+    private _viewAlong(direction: [number, number, number]): void {
+        const norm = Math.sqrt(
```
rosecers, line 832: Would it be beneficial to use something like https://mathjs.org/docs/reference/functions/norm.html?

src/structure/widget.ts
```diff
@@ -648,14 +800,87 @@ export class MoleculeViewer {
...
+        const quaternion = [
+            // rotation axis is direction ^ Oz
+            // rotation[0] * sin(angle / 2)
+            (direction[1] / norm) * Math.sin(angle / 2),
+            // rotation[1] * sin(angle / 2)
+            (-direction[0] / norm) * Math.sin(angle / 2),
+            // rotation[2] * sin(angle / 2)
+            0,
+            Math.cos(angle / 2),
+        ];
```
rosecers, line 848: I'm not used to this definition of a quaternion. Shouldn't it be [cos(th/2), n * sin(th/2)]? ala https://mathworld.wolfram.com/Quaternion.html.

rosecers (commented):
Did not realize when we spoke that I was three lines from the end of the last commit... 

Looks fine overall. I like the appearance of 3dmol compared to jsmol in general, but am not aware of their difference in capabilities. Main visual bug is the missing atom labels, and an annoyance is that the mols are not centered in their supercell.

src/structure/widget.ts
Luthaf: That's the naming & data format used by 3Dmol

src/structure/3dmol/index.d.ts
Luthaf: This is declaring a class which exists in 3Dmol, to be able to use it with typescript

src/structure/widget.ts
Luthaf: We don't currently depend on math.js, so we don't have access to this function =)

src/structure/widget.ts
Luthaf: I used http://euclideanspace.com/maths/geometry/rotations/conversions/angleToQuaternion/index.htm and it work, but I can not say I understand quaternions ... This is the rotation from the Z axis to where we want to put the camera

ceriottm:
JSMol
image

vs 3Dmol.js
image

with grid view we have a lot of small molecules, and it would be good for them to look clear. antialias=true seems to do nothing in `widget.ts:189`.

Luthaf:
This aliasing issue is quite weird: the structure looks good on macOS, but not inside a VM running linux (so same hardware & screen). 

Linux:
linux

macOS:
mac

I'll look into this to see if there is a setting to add on the canvas element.

src/structure/widget.ts
Luthaf: looks like 3Dmol is using a different convention, i.e. `[n * sin(th/2), cos(th/2)]`

Luthaf:
Except for unit tests, this should be good to go. Please have a look and try to break everything @ceriottm and @rosecers! I'll try to add unit tests next week.

Luthaf:
I've started working on unit tests in a separate branch (cf #132). Adding tests also involve changes to the code and potential new bugs, so I think we should merge this branch as-is except if there is a blocking bug, and work on adding tests & fixing edge cases in later PR.

Luthaf:
Merged! Hope that I will not have to revert this before the weekend =)

Issue #108, opened 2021-02-03
Typo in tutorial "input file format for chemiscope"


agoscinski:
In the first pseudo script under the section "Creating an input file"
the ase.io import needs to be changed to 
```py
import ase.io
```
same for sklearn import
```py
import sklearn.decomposition
```
(or you change the usage later in the pseudo script)

PR #112, opened 2021-02-05
Small cleanup after this week feedback

Luthaf:
This is best reviewed one commit at the time =)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,6 @@ webpage: https://www.epfl.ch/labs/cosmo/people/)
 git clone https://github.com/cosmo-epfl/chemiscope
 cd chemiscope
 npm install
-npm run download-example-input
 npm start
 
 # navigate to localhost:8080
--- a/python/chemiscope/input.py
+++ b/python/chemiscope/input.py
@@ -145,9 +145,11 @@ def write_input(path, frames, meta=None, properties=None, cutoff=None):
 
     .. code-block:: python
 
+        import ase
         from ase import io
         import numpy as np
         import sklearn
+        from sklearn import decomposition
         from chemiscope import write_input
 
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -45,7 +45,6 @@ webpage: https://www.epfl.ch/labs/cosmo/people/)
 git clone https://github.com/cosmo-epfl/chemiscope
 cd chemiscope
 npm install
-npm run download-example-input
```
rosecers, line 48: Why have we removed this? If you are doing dev it's really important you have the example input. Is it because we're moving the package to be installed more beyond dev purposes?

python/chemiscope/input.py
```diff
@@ -145,9 +145,11 @@ def write_input(path, frames, meta=None, properties=None, cutoff=None):
 
     .. code-block:: python
 
+        import ase
         from ase import io
         import numpy as np
         import sklearn
+        from sklearn import decomposition
```
rosecers, line 152: Why is this necessary? With the import of sklearn, we shouldn't need the explicit decomposition import.

rosecers (approved):
image

README.md
Luthaf: No, this is because example input are now automatically downloaded when running `npm start` =)

python/chemiscope/input.py
Luthaf: At least locally, I get this error:

```
>>> import sklearn
>>> sklearn.decomposition.PCA
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'sklearn' has no attribute 'decomposition'
>>>
```

This was also reported by Alex in #108

PR #113, opened 2021-02-05
Working version of screened opacity for properties

rosecers:
This is a resurrection of #81, which addresses #79, ascribing selective opacity to points in the main trace. I've opened a new pull request because I've taken a different approach than in #81.

Here is the reasoning behind the design choices:
1. We would like to keep all of the main points in a single trace to make them clickable/discoverable.
2. Plotly does not allow varying opacity in a single trace. Therefore, in order to vary opacity in a single trace, we must convert the colormap to an RGBA (vs RGB).
3. The way I've done (2) is by stacking a bunch of RGBA colormaps of discrete opacities on top of each other (see screenshot). Because of this, I've separated the coloraxis for the colorbar, so it only appears with the full-opacity colormap.

User Functionality Questions:
What types of selective opacities do we envision? Here are some I anticipate:
1. Only highlight a structure / environment / all structures in the widget (trivial)
2. Scale the opacity with some property (trivial)
3. Render invisible any point which doesn't have the property solicited by x, y, z, or color (trivial)

Thoughts? Any other use cases we should anticipate

### DIFF
--- a/src/map/colorscales.ts
+++ b/src/map/colorscales.ts
@@ -3,6 +3,7 @@
  * @module map
  */
 
+import assert from 'assert';
 import { ColorScale } from './plotly/plotly-scatter';
 
 type RGBColorMap = [number, [number, number, number]][];
@@ -684,20 +685,27 @@ const BRG: RGBColorMap = [
 ];
 
 interface ColorMaps {
-    [key: string]: ColorScale;
+    [key: string]: RGBColorMap;
 }
 
 /* eslint-disable sort-keys */
 /** @hidden */
 export const COLOR_MAPS: ColorMaps = {
-    inferno: rgb_to_plotly(INFERNO),
-    magma: rgb_to_plotly(MAGMA),
-    plasma: rgb_to_plotly(PLASMA),
-    viridis: rgb_to_plotly(VIRIDIS),
-    cividis: rgb_to_plotly(CIVIDIS),
-    seismic: rgb_to_plotly(SEISMIC),
-    brg: rgb_to_plotly(BRG),
-    'twilight (periodic)': rgb_to_plotly(TWILIGHT),
-    'twilight dark (periodic)': rgb_to_plotly(TWILIGHT_SHIFTED),
-    'hsv (periodic)': rgb_to_plotly(HSV),
+    inferno: INFERNO,
+    magma: MAGMA,
+    plasma: PLASMA,
+    viridis: VIRIDIS,
+    cividis: CIVIDIS,
+    seismic: SEISMIC,
+    brg: BRG,
+    'twilight (periodic)': TWILIGHT,
+    'twilight dark (periodic)': TWILIGHT_SHIFTED,
+    'hsv (periodic)': HSV,
 };
+
+export function getColorMap(colormap: string): ColorScale {
+    const cmap = COLOR_MAPS[colormap];
+    assert(cmap !== undefined);
+
+    return rgb_to_plotly(cmap);
+}
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -39,6 +39,21 @@ const DEFAULT_LAYOUT = {
         colorscale: [] as Plotly.ColorScale,
         showscale: true,
     },
+    coloraxis2: {
+        cmax: 0,
+        cmin: 0,
+        colorbar: {
+            len: 1,
+            thickness: 20,
+            title: {
+                text: '' as undefined | string,
+            },
+            y: 0,
+            yanchor: 'bottom',
+        },
+        colorscale: [] as Plotly.ColorScale,
+        showscale: false,
+    },
     hovermode: 'closest',
     legend: {
         itemclick: false,
@@ -156,6 +171,10 @@ const DEFAULT_CONFIG = {
     ],
 };
 
+const MAIN_TRACE: number = 0;
+const SELECTED_TRACE: number = 1;
+const BACKGROUND_TRACE: number = 2;
+const ALL_TRACES: number[] = [MAIN_TRACE, SELECTED_TRACE, BACKGROUND_TRACE];
 /**
  * The [[PropertiesMap]] class displays a 2D or 3D map (scatter plot) of
  * properties in the dataset, using [plotly.js](https://plot.ly/javascript/)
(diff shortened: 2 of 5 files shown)


ceriottm:
Thanks Rose. Looks very nice in 2D. The 3D version though is not so great 
image
is there a fundamental reason? Should we just disable opacity in 3D?

rosecers:
@Luthaf and I were discussing 3D, and didn't come to a final decision, so I didn't test specifically for 3D. On one hand, it would be very nice to "mask" some points in 3D, aka set them to near or fully transparent, but the overlaying of the points by plotly is non-trivial. Let me see if I can make a few edits and see what it'd look like.

rosecers:
@ceriottm I don't know if turning opacity on/off for 3D with my original solution is trivial, and upon further reflection came up with another approach. From the commit message:


> This is another option for the opacity approach. IMO, I think this approach is overall simpler and leads to greater ease in the future, but will prohibit us from one key functionality: opacity scaled with a property. In this approach, rather than using stacked colormaps, I've created a background trace. This will incur some lag, as based upon how the values for each scatter point are stored, with each update we need to recompute which points fall into each trace multiple times. _However_, this does enable "ghosting" of points in 3D, and quite nicely, which I anticipate will be a more desirable feature than the scaled opacity. Let me know which approach you all prefer and I'll move forward.

rosecers:
@Luthaf reviving this PR before it becomes stale

Luthaf:
Trying to run this code gives me this error whenever I load a dataset

```
AssertionErrorERR_ASSERTIONisFinite(min) && isFinite(max)
Backtrace

Wrapper@http://localhost:8080/chemiscope.min.js:1728:473
AssertionError@http://localhost:8080/chemiscope.min.js:2073:80
innerOk@http://localhost:8080/chemiscope.min.js:1208:15
ok@http://localhost:8080/chemiscope.min.js:1227:11
arrayMaxMin@http://localhost:8080/chemiscope.min.js:149249:14
./src/map/map.ts/PropertiesMap</PropertiesMap.prototype._connectSettings@http://localhost:8080/chemiscope.min.js:145182:33
PropertiesMap@http://localhost:8080/chemiscope.min.js:144953:14
DefaultVisualizer@http://localhost:8080/chemiscope.min.js:143079:20
./src/index.ts/DefaultVisualizer</DefaultVisualizer.load/<@http://localhost:8080/chemiscope.min.js:143123:30
./src/index.ts/DefaultVisualizer</DefaultVisualizer.load@http://localhost:8080/chemiscope.min.js:143122:16
setupChemiscope@http://localhost:8080/static/js/default.js:91:34
loadExample/<@http://localhost:8080/static/js/default.js:178:40
```

Could you look if you can reproduce this? I'll have a look at the code in the meantime

src/map/map.ts
```diff
@@ -522,18 +567,185 @@ export class PropertiesMap {
...
             } as unknown) as Layout);
+            this._relayout(({
+                'coloraxis2.colorscale': this._options.colorScale(),
+            } as unknown) as Layout);
+        };
+
+        // ======= opacity
+        this._options.opacity.mode.onchange = () => {
+            if (this._options.opacity.mode.value === 'filter') {
+                this._options.opacity.enableColors.enable();
```
Luthaf, line 591: I'm not sold on using a lot of select/checkboxes as the user interface for this feature.

I think it is worth exploring something like a textbox/small code editor in which the user would enter mathematical expressions, something like

```
properties["PCA[1]"] < -0.2 && sin(properties["PCA[2]"]) > 0.2
```

Then we could take this expression and evaluate it for each point to decide whether to have it as opaque or transparent. This would replace `_filterForOpacity`.

src/map/map.ts
```diff
@@ -630,7 +846,34 @@ export class PropertiesMap {
             showlegend: false,
         };
 
-        const traces = [main as Data, selected as Data];
+        // The main trace, containing default data
```
Luthaf, line 849: Shouldn't this comment read "the background trace, containing transparent data"?

Luthaf (commented):
I think I like this approach much better than the one with variable colormap. I have a few small comments on code & how to present this feature to the users

src/map/colorscales.ts
```diff
@@ -684,20 +685,27 @@ const BRG: RGBColorMap = [
...
+    viridis: VIRIDIS,
+    cividis: CIVIDIS,
+    seismic: SEISMIC,
+    brg: BRG,
+    'twilight (periodic)': TWILIGHT,
+    'twilight dark (periodic)': TWILIGHT_SHIFTED,
+    'hsv (periodic)': HSV,
 };
+
+export function getColorMap(colormap: string): ColorScale {
```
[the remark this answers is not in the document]
rosecers: So it serves as a wrapper to 1) keep colorscales contained to `colorscales.ts` and 2) check whether a colorscale is defined.

src/map/map.ts
```diff
@@ -340,7 +355,7 @@ export class PropertiesMap {
         // ======= x axis settings
         this._options.x.property.onchange = () => {
             const values = this._coordinates(this._options.x) as number[][];
-            this._restyle({ x: values }, [0, 1]);
+            this._restyle({ x: values }, [0, 1, 2]);
```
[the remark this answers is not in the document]
rosecers: Fixed!

src/map/map.ts
rosecers: Fixed!

rosecers:
@Luthaf can we push this through at some point?

Luthaf:
Can you rebase & fix CI? After that I remember the main part of work remaining with this feature was the user interface side of it. Do you have some time today/next week to discuss it in person?

rosecers:
> Can you rebase & fix CI? After that I remember the main part of work remaining with this feature was the user interface side of it. Do you have some time today/next week to discuss it in person?

I can chat this afternoon for sure -- as for the eslint / prettier, I'm running eslint on my end and prettier, but it still fails in test. Is there a new style file that has to be updated locally?

Luthaf:
> as for the eslint / prettier, I'm running eslint on my end and prettier, but it still fails in test. Is there a new style file that has to be updated locally?

No, but you may want to update your prettier version: `rm -rf node_modules && node ci`. `node ci` will install the exact same version used in CI.

Luthaf:
@rosecers what should we do with this code? Do you still want this feature and should we try to revive the branch, or should this mainly be kept around as an example in case someone else wants to revive it later?

rosecers:
I think we ultimately chose to leave it here for posterity.

Luthaf:
Let's close the PR and keep the branch around in case someone wants to revive this.

Issue #109, opened 2021-02-03
Warning should be raised when log scale is used for negative values


rosecers:

Luthaf:
This is a good issue for newcomers!

The code to change would be around https://github.com/cosmo-epfl/chemiscope/blob/15bdb4147c3bb022d17a8b67d82fa53fd3145d88/src/map/options.ts#L212, adding a check that all values are positive before applying the operation in the `'log'` case.

3dvkr:
@Luthaf I think a quick array method on the `value` would do the check. I'm a little stuck on what kind of warning to raise. Should the function return `null` or would the warning need look like something else?

Luthaf:
Yes, `sendWarning` is the right function for this! You can call it with a string argument containing the warning message, something like `sendWarning("this property contains negative values, taking the log will discard them")`.

3dvkr:
@Luthaf I think I've got the right idea for what to do, but I'm having a bit of trouble with the linter. I wrote the check as an arrow function, and as per output I added parentheses and semi-colons. It is passing all other tests. Would you know why it's still failing that one with the linter?

rosecers:
Hi @3dvkr -- looking at #122 right now. I think I intended this issue to apply to the log scaling of the axes, as the sizes are never negative, but rather scaled between the max and min of the property value. This can be seen in the built chemiscope that the log-scaled sizes never make negative values disappear, but rather just get very small. However, when you apply a log scale to the axes, points do disappear. Sorry for the confusion! I should have corrected @Luthaf's original comment.

Luthaf:
Right, I should have read the surrounding code as well =/. Really sorry about that. 

This makes this issue a bit harder to deal with, the code to modify is now the three functions `this._options.x.scale.onchange`/`this._options.y.scale.onchange`/`this._options.z.scale.onchange`; as well as `changeRange`.

https://github.com/cosmo-epfl/chemiscope/blob/eecc7349b2c31f992930ff7343086a8d98a28e62/src/map/map.ts#L350-L358

https://github.com/cosmo-epfl/chemiscope/blob/eecc7349b2c31f992930ff7343086a8d98a28e62/src/map/map.ts#L362

[...]

PR #121, opened 2021-02-28
Update options.ts, RE: Issue #109

3dvkr:
Added a check that all values are positive before applying the operation in the `'log'` case. It will issue a warning for when the log scale is used for negative values.

### DIFF
--- a/src/map/options.ts
+++ b/src/map/options.ts
@@ -210,6 +210,9 @@ export class MapOptions extends OptionsGroup {
      * all of them if `trace === undefined`.
      */
     public calculateSizes(rawSizes: number[]): number[] {
+        if (!rawSizes.every(x => x > 0)) {
+                sendWarning("This property contains negative values, taking the log will discard them.");
+        }
         const logSlider = (value: number) => {
             const min_slider = 1;
             const max_slider = 100;
--- a/src/map/options.ts
+++ b/src/map/options.ts
@@ -210,7 +210,7 @@ export class MapOptions extends OptionsGroup {
      * all of them if `trace === undefined`.
      */
     public calculateSizes(rawSizes: number[]): number[] {
-        if (!rawSizes.every((x) => x > 0)) {
+        if (!rawSizes.every((x: number) => x > 0)) {
                 sendWarning("This property contains negative values, taking the log will discard them.");
         }
         const logSlider = (value: number) => {
--- a/src/map/options.ts
+++ b/src/map/options.ts
@@ -210,9 +210,7 @@ export class MapOptions extends OptionsGroup {
      * all of them if `trace === undefined`.
      */
     public calculateSizes(rawSizes: number[]): number[] {
-        if (!rawSizes.every((x: number) => x > 0)) {
-                sendWarning("This property contains negative values, taking the log will discard them.");
-        }
+
         const logSlider = (value: number) => {
             const min_slider = 1;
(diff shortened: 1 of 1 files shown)


Luthaf:
The linter fail is related to the code formatter we use (prettier: https://prettier.io/). You can run it to automatically format the code with `npx prettier -w src`. You can also configure your editor to do this automatically for you when saving a file, cf the prettier website.

3dvkr:
I ran the code with prettier, but even when I take out my added if-statement, the linter gives me an error regarding an unexpected `;` in line 212.

Luthaf:
Ok, that's strange. I'll look into it tomorrow!

src/map/options.ts
```diff
@@ -210,6 +210,9 @@ export class MapOptions extends OptionsGroup {
      * all of them if `trace === undefined`.
      */
     public calculateSizes(rawSizes: number[]): number[] {
+        if (!rawSizes.every((x: number) => x > 0)) {
```
Luthaf, line 213: This check is at the wrong place. It should only check for negative values when `scaleMode` (below) is `'log'`.

Luthaf:
> I ran the code with prettier, but even when I take out my added if-statement, the linter gives me an error regarding an unexpected ; in line 212.

How are you running prettier? 

I checked out your PR locally; and using `npx prettier -w src` in the chemiscope source folder worked just fine, and made the linter happy. If this does not for you, can you copy/paste the output of `npx prettier --version`? It should be `2.1.2`.

----

Unrelated question, are you using this project yourself? If so, which research group are you from? We would like to know how widely this tool is used 😃

src/map/options.ts
3dvkr: That makes more sense! Thank you. Is it also okay to take the log of zero, or is there logic later on to handle if `scaled` is `Infinity`?

3dvkr:
I'm not using this project myself, but really impressed with what you've built. Any plans to compile a list of research groups using chemiscope in an  .md file for use case examples? 

I was looking for a good first issue, and estimated this would be a quick change, so I've been using the web interface, and the online prettier website. I'll try running the tests locally.

src/map/options.ts
Luthaf: There is no special logic later, so warning on 0 is also good!

src/map/options.ts
```diff
@@ -243,6 +244,9 @@ export class MapOptions extends OptionsGroup {
                         scaled = 1.0 / scaled;
                         break;
                     case 'log':
+                        if (scaled < 0) {
+                            sendWarning("This property contains negative values, taking the log will discard them.");
```
Luthaf, line 248: This will send multiple warnings if the property contains multiple negative values. I think it would be better to move this check out of the `rawSizes.map` call; and only warn once.

Luthaf:
> Any plans to compile a list of research groups using chemiscope in an .md file for use case examples?

Either a MD file, or a section of the docs/readme compiling a list of paper using chemiscope would be great!

> I was looking for a good first issue, and estimated this would be a quick change, so I've been using the web interface, and the online prettier website. I'll try running the tests locally.

For a first PR, don't worry too much about formatting, I can run autoformat locally manually before merging. If you want to contribute more, getting the code to work locally is always a good idea!

Luthaf:
Closed in favor of #122

PR #122, opened 2021-03-01
RE Issue #109

3dvkr:
- trying a PR with locally run npm/prettier tests

### DIFF
--- a/src/map/options.ts
+++ b/src/map/options.ts
@@ -210,6 +210,9 @@ export class MapOptions extends OptionsGroup {
      * all of them if `trace === undefined`.
      */
     public calculateSizes(rawSizes: number[]): number[] {
+        if (!rawSizes.every(x => x > 0)) {
+                sendWarning("This property contains negative values, taking the log will discard them.");
+        }
         const logSlider = (value: number) => {
             const min_slider = 1;
             const max_slider = 100;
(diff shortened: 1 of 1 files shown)


src/map/options.ts
```diff
@@ -243,6 +243,11 @@ export class MapOptions extends OptionsGroup {
                         scaled = 1.0 / scaled;
                         break;
                     case 'log':
+                        if (scaled <= 0) {
+                            sendWarning(
```
Luthaf, line 247: Copy/pasting from #121:

This will send multiple warnings if the property contains multiple negative values. I think it would be better to move this check out of the rawSizes.map call; and only warn once.

Luthaf (commented):
This looks better! One final thing before we can merge this:

src/map/options.ts
3dvkr: Sounds good. I'm trying to find a good spot for it. Perhaps right before the if-statement that contains the rawSizes.map call? I was thinking that it would warn before `scaleMode` is declared/assigned, but not sure if there is a way to change the scale mode the user has selected by that point anyway.

Luthaf:
Closing due to inactivity, feel free to re-open a new PR with the updates requested above.

PR #129, opened 2021-05-13
Update dependencies and small improvements to point size in 3D

Luthaf:
The point size in 3D is now much closer to the point size in 2D:
2D

Before:
before

With this PR:
after


src/map/options.ts
```diff
@@ -224,47 +224,39 @@ export class MapOptions extends OptionsGroup {
 
         const userFactor = logSlider(this.size.factor.value);
 
-        let values;
-        if (this.size.mode.value !== 'constant') {
```
Luthaf, line 228: The whole if/else was no longer needed since the switch below covers the same case.

Luthaf:
Since this contains mostly trivial changes, I'll go ahead and merge this PR so that I can work on making #106 up to date.

PR #136, opened 2021-06-02
Various cleanup after the removal of JSmol

Luthaf:
- remove all uses of J2Spath
- rename partial build from jsmol-widget.min.js to molecule-viewer.min.js
- remove all remaining references to Jsmol


.github/workflows/website.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           cp -r app/ gh-pages/
           cp dist/*.min.js gh-pages/
           cp chemiscope_standalone.html gh-pages/
-          cp -r docs/_build/html/ gh-pages/docs/
```
rosecers, line 39: What does this change have to do with jsmol?

rosecers (approved):
This all-in-all looks good, is working well, and I can't find any additional JSMol fragments to remove. Can you add a quick note about the _build --> build change just for the git history?

.github/workflows/website.yml
Luthaf: not much, I was checking the docs and made a small change on the side. I'll extract the change to its own commit

PR #138, opened 2021-06-04
Refactor the App code, using typescript and async functions

Luthaf:
~This PR builds on top of #136, and will be rebased once it is merged.~

The main improvements are:
- the app and main library code use the same language
- the app uses async function, getting rid of callback hell and nested `.then()` calls
- the app loads (most of) its dependencies directly, instead of relying on external loading (this is mainly about bootstrap)


Luthaf:
This is now ready to be reviewed!

ceriottm:
Shouldn't we name the app class something more descriptive than "App"?

Luthaf:
We can, do you have a suggestion? Naming is less critical here than for the main library, since this code is only intended to be used for chemiscope.org.

ceriottm:
boh, even just ChemiscopeApp, CSApp, or something like that

Luthaf:
Renamed to `ChemiscopeApp`!

Luthaf:
Let's merge this so that other PR that depend on it can move forward!

PR #139, opened 2021-06-08
Added SVG and updated PNG download buttons in Plotly

### LINKED ISSUE #120: Allow saving the map to SVG
The idea would be to add a button to save the map to SVG in addition to (or in replacement of) the button to save the map to PNG. 

The main use case is to later edit this SVG file to create scientific figures; in particular removing the x/y axis.

jakublala:
Fixes #120


src/map/map.ts
```diff
@@ -102,6 +102,42 @@ const DEFAULT_CONFIG = {
...
...0,30.871-13.827,30.871-30.87v-160.54C519.297,210.838,505.47,197.019,488.426,197.019z M97.2,21.605 h250.193v110.513c0,5.967,4.841,10.8,10.8,10.8h95.407v54.108H97.2V21.605z M234.344,335.86v45.831h-31.601V229.524h40.184 l31.611,55.759c9.025,16.031,18.064,34.983,24.825,52.154h0.675c-2.257-20.103-2.933-40.643-2.933-63.44v-44.473h31.614v152.167 h-36.117l-32.516-58.703c-9.049-16.253-18.971-35.892-26.438-53.727l-0.665,0.222C233.906,289.58,234.344,311.027,234.344,335.86z M71.556,381.691V231.56c10.613-1.804,25.516-3.159,46.506-3.159c21.215,0,36.353,4.061,46.509,12.192 c9.698,7.673,16.255,20.313,16.255,35.219c0,14.897-4.959,27.549-13.999,36.123c-11.738,11.063-29.123,16.031-49.441,16.031 c-4.522,0-8.593-0.231-11.736-0.675v54.411H71.556V381.691z M453.601,523.353H97.2V419.302h356.4V523.353z M485.652,374.688 c-10.61,3.607-30.713,8.585-50.805,8.585c-27.759,0-47.872-7.003-61.857-20.545c-13.995-13.1-21.684-32.97-21.452-55.318 c0.222-50.569,37.03-79.463,86.917-79.463c19.644,0,34.783,3.829,42.219,7.446l-7.214,27.543c-8.369-3.617-18.752-6.55-35.458-6.55 c-28.656,0-50.341,16.256-50.341,49.22c0,31.382,19.649,49.892,47.872,49.892c7.895,0,14.218-0.901,16.934-2.257v-31.835h-23.493 v-26.869h56.679V374.688z',
```
Luthaf, line 114: This should use the webpack raw-loader we have configured for SVG instead of putting the path inline; i.e. something like 

https://github.com/cosmo-epfl/chemiscope/blob/9ec79b6c604f02be30afb467a0d7da734d761ea0/src/structure/grid.ts#L17-L18

This allow one to preview & easily change the svg here. One possible issue would be that plotly expect a path and not a full svg, but it might be possible to extract the path from the svg file.

package.json
```diff
@@ -71,5 +71,8 @@
     "webpack": "^5.37.0",
     "webpack-cli": "^4.7.0",
     "webpack-dev-server": "4.0.0-beta.3"
+  },
+  "dependencies": {
+    "extract-svg-path": "^2.1.0"
```
Luthaf, line 76: You should remove this dependency and undo the changes in package-lock.json

src/map/map.ts
```diff
@@ -1161,3 +1172,10 @@ export class PropertiesMap {
         return check;
     }
 }
+
+/** Extract an SVG path */
```
Luthaf, line 1176:
```suggestion
/** Extract the data associated with the first `path` element in an SVG string */
```

src/static/png-icon.svg
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
Luthaf, line 1: Where did you find this file? Are we sure we are allowed to use it?

src/static/svg-icon.svg
```diff
@@ -0,0 +1,56 @@
...
+		c16.751,0,25.607-6.935,25.607-17.465c0-10.052-7.662-15.795-27.05-22.734c-26.8-9.328-44.263-24.168-44.263-47.611
+		c0-27.524,22.971-48.579,61.014-48.579c18.188,0,31.591,3.823,41.159,8.131l-8.126,29.437c-6.465-3.116-17.945-7.657-33.745-7.657
+		c-15.791,0-23.454,7.183-23.454,15.552c0,10.296,9.089,14.842,29.917,22.731c28.468,10.536,41.871,25.365,41.871,48.094
+		c0,27.042-20.812,50.013-65.09,50.013C95.731,389.349,77.538,384.571,68.458,379.54z M453.601,523.353H97.2V419.302h356.4V523.353z
+		 M488.911,379.54c-11.243,3.823-32.537,9.103-53.831,9.103c-29.437,0-50.73-7.426-65.57-21.779
+		c-14.839-13.875-22.971-34.942-22.738-58.625c0.253-53.604,39.255-84.235,92.137-84.235c20.81,0,36.852,4.073,44.74,7.896
+		l-7.657,29.202c-8.859-3.829-19.849-6.95-37.567-6.95c-30.396,0-53.357,17.233-53.357,52.173c0,33.265,20.81,52.882,50.73,52.882
+		c8.375,0,15.072-0.96,17.94-2.395v-33.745h-24.875v-28.471h60.049V379.54L488.911,379.54z"/>
+</g>
+<g>
```
Luthaf, line 26: can you remove the extraneous `<g>` tags here as well?

src/map/map.ts
```diff
@@ -102,6 +105,50 @@ const DEFAULT_CONFIG = {
...
+                width: 400,
+                height: 424,
+                path: extractSvgPath(PNG_SVG),
+            },
+            click: function (gd: PlotlyScatterElement) {
+                Plotly.downloadImage(gd, {
+                    filename: 'newplot',
+                    format: 'png',
+                    width: gd._fullLayout.width,
+                    height: gd._fullLayout.height,
```
Luthaf, line 124: This is a bit small, and the plot ends up pixelated. I think we should multiply the width/height by 2, and make sure this is at least 600px in each direction.

src/map/map.ts
```diff
@@ -102,6 +105,50 @@ const DEFAULT_CONFIG = {
...
+        {
+            name: 'Download as PNG',
+            icon: {
+                width: 400,
+                height: 424,
+                path: extractSvgPath(PNG_SVG),
+            },
+            click: function (gd: PlotlyScatterElement) {
+                Plotly.downloadImage(gd, {
+                    filename: 'newplot',
```
Luthaf, line 121: How about something like this instead, to make the filename clearer?

```suggestion
                    filename: 'chemiscope-map',
```

Luthaf (commented):
This feature is working very well, I have some suggestions to improve the user interface.


In addition to the comments below, the modebar now looks a bit strange:

Capture d’écran 2021-06-14 à 10 19 30

Could we 
- move the download as SVG/PNG buttons to the left of the modebar?
- use thicker lines in the icons? They look very thin compared to the other icons.

src/map/map.ts
Luthaf: This proved to be harder than expected =( Changing the size of the plot also changes the spacing between points; and using the `scale` parameter does scale the plot up; but the resulting plot is still pixelated. I think this is because we are using `scattergl` trace, and plotly does not re-render the plot on a large canvas, but instead use the already existing one.

src/map/map.ts
```diff
@@ -102,6 +105,54 @@ const DEFAULT_CONFIG = {
...
+                    height: 424,
+                    path: extractSvgPath(PNG_SVG),
+                },
+                click: function (gd: PlotlyScatterElement) {
+                    Plotly.downloadImage(gd, {
+                        filename: 'chemiscope-map',
+                        format: 'png',
+                        width: Math.max(gd._fullLayout.width, 600),
+                        height: Math.max(gd._fullLayout.width, 600),
+                    } as any).catch((e) =>
```
Luthaf, line 126:
```suggestion
                    }).catch((e) =>
```

Luthaf (approved):
Looks good, just a small `any` to remove! Here again, make sure to squash commits before merging, either manually or with the github UI

PR #143, opened 2021-06-09
Reset map zoom on property change

Luthaf:
Keeping custom range when changing properties makes little sense, since different properties will have different units & scale.

@jakublala, do you think you could try to review this?

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -317,7 +317,7 @@ export class PropertiesMap {
         return this._options.saveSettings();
     }
 
-    /** Forward to Ploty.restyle */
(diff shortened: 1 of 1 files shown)


jakublala:
Changing the property for the z-axis does not rescale the graph.

Pulling up the z-axis also produces this error - TypeError: Cannot read property '_inputDomain' of undefined.

Luthaf:
Ok, I forgot to test the 3D mode, and obviously it is using different parameters to `autoscale` ...

jakublala (approved):
All good.

PR #144, opened 2021-06-12
Reorganizing the map settings UI

ceriottm:
Rationalizing the layout of options in the map settings.

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -657,8 +657,8 @@ export class PropertiesMap {
         };
 
         // ======= markers size
-        // setup initial state of the z axis settings
-        if (this._options.size.mode.value === 'constant') {
+        // setup initial state of the marker size settings
+        if (this._options.size.mode.value === '') {
             this._options.size.property.disable();
             this._options.size.reverse.disable();
         } else {
@@ -666,12 +666,12 @@ export class PropertiesMap {
             this._options.size.reverse.enable();
         }
 
-        this._options.size.mode.onchange = () => {
-            if (this._options.size.mode.value !== 'constant') {
-                this._options.size.property.enable();
+        this._options.size.property.onchange = () => {
+            if (this._options.size.property.value !== '') {
+                this._options.size.mode.enable();
                 this._options.size.reverse.enable();
             } else {
-                this._options.size.property.disable();
+                this._options.size.mode.disable();
                 this._options.size.reverse.disable();
             }
             this._restyle({ 'marker.size': this._sizes(0) } as Data, 0);
@@ -681,7 +681,7 @@ export class PropertiesMap {
             this._restyle({ 'marker.size': this._sizes(0) } as Data, 0);
         };
 
-        this._options.size.property.onchange = () => {
+        this._options.size.mode.onchange = () => {
             this._restyle({ 'marker.size': this._sizes(0) } as Data, 0);
         };
 
@@ -945,7 +945,14 @@ export class PropertiesMap {
      * all of them if `trace === undefined`.
      */
     private _sizes(trace?: number): Array<number | number[]> {
-        const sizes = this._property(this._options.size.property.value).values;
+        let sizes;
+        if (this._options.size.property.value !== '') {
+            sizes = this._property(this._options.size.property.value).values;
+        } else {
+            sizes = new Array(this._property(this._options.x.property.value).values.length).fill(
+                1.0
+            );
+        }
         const values = this._options.calculateSizes(sizes);
         const selected = [];
         if (this._is3D()) {
--- a/src/map/options.ts
+++ b/src/map/options.ts
@@ -100,7 +100,7 @@ export class MapOptions extends OptionsGroup {
 
         this.x = new AxisOptions(propertiesName);
         this.y = new AxisOptions(propertiesName);
-        // For z and color, '' is a valid value
+        // For z and color '' is a valid value
         this.z = new AxisOptions(propertiesName.concat(['']));
         this.color = new AxisOptions(propertiesName.concat(['']));
 
@@ -118,20 +118,18 @@ export class MapOptions extends OptionsGroup {
 
         this.size = {
             factor: new HTMLOption('number', 50),
-            mode: new HTMLOption('string', 'constant'),
-            property: new HTMLOption('string', propertiesName[0]),
+            mode: new HTMLOption('string', 'linear'),
+            property: new HTMLOption('string', ''),
             reverse: new HTMLOption('boolean', false),
         };
-        this.size.property.validate = optionValidator(propertiesName, 'size');
+
+        this.size.property.validate = optionValidator(propertiesName.concat(['']), 'size');
         this.size.factor.validate = (value) => {
             if (value < 1 || value > 100) {
                 throw Error(`size factor must be between 0 and 100, got ${value}`);
             }
         };
-        this.size.mode.validate = optionValidator(
-            ['constant', 'linear', 'log', 'sqrt', 'inverse'],
-            'size'
-        );
+        this.size.mode.validate = optionValidator(['linear', 'log', 'sqrt', 'inverse'], 'mode');
 
         this.x.property.value = propertiesName[0];
         this.y.property.value = propertiesName[1];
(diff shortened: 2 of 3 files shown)


ceriottm:
This might also be a good opportunity to fix the "grayed out options" issues. Besides the size, there is also a problem with the z axis, visible in the zeolite example

src/map/options.ts
```diff
@@ -224,17 +223,22 @@ export class MapOptions extends OptionsGroup {
...
         const defaultSize = this.is3D() ? 800 : 150;
         const bottomLimit = 0.1; // lower limit to prevent size of 0
 
         const values = rawSizes.map((v: number) => {
             // normalize between 0 and 1, then scale by the user provided value
-            let scaled = (v + bottomLimit - min) / (max - min);
-            if (reversed) {
-                scaled = 1.0 + bottomLimit - scaled;
+            let scaled = 0.55; // default
+            if (max - min === 0) {
```
Luthaf, line 235:
```suggestion
            if (max === min) {
```

src/map/options.ts
```diff
@@ -224,17 +223,22 @@ export class MapOptions extends OptionsGroup {
...
+                scaleMode = 'fixed';
```
Luthaf, line 236: the comment below still mention 'constant', one of these should be updated

src/map/options.ts
```diff
@@ -406,7 +410,9 @@ export class MapOptions extends OptionsGroup {
 
         // ======= marker size
         const selectSizeProperty = getByID<HTMLSelectElement>('chsp-size');
+        // first option is 'none'
```
Luthaf, line 413:
```suggestion
        // first option is 'fixed'
```

src/map/map.ts
```diff
@@ -207,6 +207,10 @@ export class PropertiesMap {
         };
 
         this._createPlot();
+
+        // Trigger a few updates that are needed to ensure a consistent state
+        this._options.size.property.onchange(this._options.size.property.value, 'JS');
+        this._options.z.property.onchange(this._options.z.property.value, 'JS');
```
Luthaf, line 213: I think this code should live the applySettings function, but this function should already be called:

The value is updated in applySettings:
https://github.com/cosmo-epfl/chemiscope/blob/d779d1b1d899cb67c5b607a9735809d1d5c23275/src/options.ts#L329

which then call the setter
https://github.com/cosmo-epfl/chemiscope/blob/d779d1b1d899cb67c5b607a9735809d1d5c23275/src/options.ts#L136-L138

which calls `onchange`
https://github.com/cosmo-epfl/chemiscope/blob/d779d1b1d899cb67c5b607a9735809d1d5c23275/src/options.ts#L219

---

I think the actual issue here is that we are applying the saved settings a few lines above in `MapOptions` constructor, before having called `this._connectSettings()`. But we can not connect settings before we have them constructed, so the best fix IMO would be to remove the settings from the MapOption constructor; and apply them after connecting everything.

Luthaf (commented):
I did a pass on the code only, I'll do a second round with a local checkout to hunt for bugs later!

src/map/map.ts
```diff
@@ -436,14 +440,15 @@ export class PropertiesMap {
                 if (!was3D) {
                     this._switch3D();
                 }
-            }
+                this._options.z.enable();
```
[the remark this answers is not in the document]
ceriottm: I agree this is redundant here, but it is needed when you load a chemiscope that is already in 3D. the option gets initialized in the grayed-out state, and one needs to enable them. the least bad way seemed to me to force an update of the z options, but then this needs to be here.

src/map/map.ts
```diff
@@ -436,14 +440,15 @@ export class PropertiesMap {
...
-                'scene.zaxis.title': this._title(this._options.z.property.value),
-                'scene.zaxis.autorange': true,
-            } as unknown as Layout);
+                const values = this._coordinates(this._options.z);
+                this._restyle({ z: values } as Data, [0, 1]);
+                this._relayout({
+                    'scene.zaxis.title': this._title(this._options.z.property.value),
+                    'scene.zaxis.autorange': true,
+                } as unknown as Layout);
+            }
```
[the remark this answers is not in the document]
ceriottm: again, having it outside the if (which is where it was to begin with) was giving all sorts of problems when switching from 3D to 2D view.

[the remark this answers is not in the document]
Luthaf: Even with a return in the first `if` block?

src/map/map.ts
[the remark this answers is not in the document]
Luthaf: ok, then this should be fixed at the settings level if possible. I've added a suggestion below, we can have a look at this line again once the setting behave properly!

src/map/map.ts
ceriottm: So the problem we were having back then - which I think contributes to this messy design - is that if you connect first, then apply the settings, it triggers a gazillion of redraw calls, and can be *really* slow. 
IMO fixing this requires a separate PR, in which we redesign the whole update mechanism, having a separate "redraw" call, and having an option to update() to trigger the redraw or not (so that if you are changing multiple settings you can redraw just once).

rosecers:
Can we agree on some consistent nomenclature? Right now there are three words we're using for "don't change this", namely "none", "fixed", and "default", when "default" and "fixed" serve the same purpose, no?

ceriottm:
> Can we agree on some consistent nomenclature? Right now there are three words we're using for "don't change this", namely "none", "fixed", and "default", when "default" and "fixed" serve the same purpose, no?

Yes and no... I think that none of these really convey the message of what the "no property" option implies for all of the settings. for the z-axis, the z-axis does not exist. For the color (it appears) it makes it a bluish hue that cannot be tweaked. For the symbol, it uses a default circle. For the size, it keeps a fixed size for all points. Can you think of an umbrella term that covers all these cases?

src/map/map.ts
Luthaf: Ok, that make sense. In that case I would rather not add a temporary hack to the code and make this PR focus on changing the map settings UI. 

I think we can live for a short while with greyed out boxes, and having them remind us there is something that needs to be fixed, instead of hiding it.

rosecers:
> > Can we agree on some consistent nomenclature? Right now there are three words we're using for "don't change this", namely "none", "fixed", and "default", when "default" and "fixed" serve the same purpose, no?
> 
> Yes and no... I think that none of these really convey the message of what the "no property" option implies for all of the settings. for the z-axis, the z-axis does not exist. For the color (it appears) it makes it a bluish hue that cannot be tweaked. For the symbol, it uses a default circle. For the size, it keeps a fixed size for all points. Can you think of an umbrella term that covers all these cases?

"constant" would be most appropriate to represent something that is property-independent and takes on a restricted set of values under this umbrella.

Luthaf:
It feels a bit weird to have `z = 'constant'` mean 2D plot. 

I'm not sure we need to use a single word here, what prevents us from using whatever makes the most sense in a given context? I think it makes the UI nicer to use, for a small increase in code complexity.

rosecers:
While “constant” would describe a 2D plot, I was more meaning in place of “default / fixed”

ceriottm:
So re the color/size/symbol issue: what would be ideal (but perhaps another PR?) would be to have the setting "manual", and then make it possible to pick the size (already possible) color and symbol. That is what would make most sense in my opinion, in the sense that rather than having those express a "dimension" of the data, they are manually set to a chosen value.

Luthaf:
> So re the color/size/symbol issue: what would be ideal (but perhaps another PR?) would be to have the setting "manual", and then make it possible to pick the size (already possible) color and symbol. That is what would make most sense in my opinion, in the sense that rather than having those express a "dimension" of the data, they are manually set to a chosen value.

That sounds like a good idea, but definitively something for a future PR! Could you open an issue for it?

ceriottm:
So for the moment we leave the different labels - no strong feelings in fixed vs constant, perhaps @rosecers  is right

src/map/map.ts
rosecers: I agree with the overarching "update", can we include in the UI a button ala "Apply settings" so that redraw isn't triggered with every change, but rather all at once?

ceriottm:
The lag is OK when you change settings in the UI. The problem is what
happens when you set each HTMLOption field from code, and each setting
update triggers a redraw.
Anyway I think this is a discussion for a separate PR, was just motivating
why this ugly hack.

src/map/map.ts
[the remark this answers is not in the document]
Luthaf: I added the return in #138 / c29bd85f59dc3807c29aed25875821358bfd445c since it was required to fix the loading of saved settings.

jakublala (approved):
All seems good!

PR #146, opened 2021-06-14
negativeLogWarning function

jakublala:
Added the negativeLogWarning function to detect whether any values displayed will be discarded as a result from changing the axis scale from linear to log. This will then display a warning to the user.

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -12,7 +12,7 @@ import { Property } from '../dataset';
 
 import { EnvironmentIndexer, Indexes } from '../indexer';
 import { OptionModificationOrigin, SavedSettings } from '../options';
-import { GUID, PositioningCallback, arrayMaxMin } from '../utils';
+import { GUID, PositioningCallback, arrayMaxMin, sendWarning } from '../utils';
 import { enumerate, getByID, getFirstKey } from '../utils';
 
 import { MapData, NumericProperty } from './data';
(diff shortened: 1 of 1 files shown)


jakublala:
Fixes #109

src/map/map.ts
```diff
@@ -337,8 +337,22 @@ export class PropertiesMap {
 
     /** Add all the required callback to the settings */
     private _connectSettings() {
+        // Reads a warning if negative values of log are discared and not shown
+        const negativeLogWarning = (axis: AxisOptions) => {
+            if (
+                axis.scale.value === 'log' &&
+                Math.min(...(this._coordinates(axis, 0)[0] as number[])) < 0 &&
```
Luthaf, line 344: This will break with large datasets. You should use `arrayMaxMin` instead

src/map/map.ts
```diff
@@ -378,6 +393,15 @@ export class PropertiesMap {
...
                 const min = axis.min.value;
                 const max = axis.max.value;
+
+                if (name === 'xaxis') {
+                    negativeLogWarning(this._options.x);
+                } else if (name === 'yaxis') {
+                    negativeLogWarning(this._options.y);
+                } else {
+                    negativeLogWarning(this._options.z);
+                }
```
Luthaf, line 403:
```suggestion
                negativeLogWarning(axis);
```

Luthaf (approved):
That looks good! Please squash commits together before merging! 

Here is some documentation on how/why do that, and we can do it together later today: https://www.internalpointers.com/post/squash-commits-into-one-git

PR #150, opened 2021-06-16
2D line and 3D scatter plots support in info bar

### LINKED ISSUE #78: Multi-dimensional Properties in Viewer
Multidimensional properties could be very nice to visualize. We can see two use-cases that could be implemented with the same infrastructure:
- DOS plots, where the property is interesting to plot by itself as well
- parametric properties (e.g. everything that depends on alpha in KPCovR), where we would want a slider to change the alpha  value and have the changes reflected in the projection/coloring/size/etc.

The same code could also be used to store (and display in the future) vector/tensorial properties as 3/9-values vectors.

---

1st milestone:
- [ ] decide & implement the data structure as stored in JSON. Something like this should work:
``` json5
"properties": {
    "<name>": {
        // array of arrays, of dimension N_structure/N_environments x p were p is the size of the parameter array below
        "values": [[...], [...]],
        // list of parameters, 1 to start but potentially multiple parameters later for 2+D properties
        "parameters": ['parameter 1'], 
    }
},
// parameters above refer to the values in this separate table so that multiple properties can use the same parameters
"parameters": {
    "parameter 1":  number[],  // p1 elements
    "parameter 2": number[],  // p2 elements
}
```
- [ ] handle non-scalar properties within code (i.e. make sure it doesn't break things)
- [ ] visualization - disable non-scalars for all plot styling / info bar

2nd milestone:
- [ ] provide necessary checks for the structure of data
- [ ] process 1D properties within code
- [ ] visualize 1D properties in plot mode

3rd milestone:
- [ ] provide slider for 1D property parameter --> plot styling or info bar

---

We may want to extend this to 2+D properties in the future, the current proposal should be forward compatible with this.

cbenmahm:
Fix #78 

For now, it is possible to visualize 1D properties as line plots (e.g. DOS plots or SOAP vectors) and 3D scatter plots in the table cells in the info bar for structures and environments. The data should be provided as [[x], [y], [z]] for every environment/structure. The z-axis is optional.

The current visualization setup is not optimal because the info bar covers entirely the structure viewer for smaller screens. I am hesitant to use bootstrap modals and have a floating window or create an extra column in the chemiscope viewer between the map and the structure viewer.

Checks for input data are not implemented (yet).

### DIFF
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -37,6 +37,8 @@ export interface Dataset {
     environments?: Environment[];
     /** Settings for visualization of this dataset */
     settings?: Partial<Settings>;
+    /** Parameters of multidimensional properties */
+    parameters?: { [name: string]: Parameter };
 }
 
 /**
@@ -137,12 +139,16 @@ export interface Property {
      *
      * string values should represent classification results (category "A", "B"
      * or "C"); and numeric values should be use for everything else.
+     *
+     * it supports 2D properties, i.e. arrays.
      */
-    values: string[] | number[];
+    values: string[] | number[] | number[][];
     /** user-facing description of the property */
     description?: string;
     /** unit of the property values */
     units?: string;
+    /** parameter of multidimensional property */
+    parameters?: string[];
 }
 
 /**
@@ -157,6 +163,16 @@ export interface Environment {
     center: number;
     /** Spherical cutoff radius, expressed in Angströms */
     cutoff: number;
+}
+
+/** Parameters */
+export interface Parameter {
+    /** values of the parameter */
+    values: number[];
+    /** description of the parameter */
+    description?: string;
+    /** units of the elements in the property array */
+    units?: string;
 }
 
 /** Arbitrary javascript object, to be validated */
@@ -204,7 +220,12 @@ export function validateDataset(o: JsObject): void {
         throw Error('"properties" must be an object in the dataset');
     }
 
-    checkProperties(o.properties as Record<string, JsObject>, structureCount, envCount);
+    checkProperties(
+        o.properties as Record<string, JsObject>,
+        structureCount,
+        envCount,
+        o.parameters as Record<string, JsObject>
+    );
 }
 
 function checkMetadata(o: JsObject) {
@@ -311,7 +332,8 @@ export function checkStructure(s: JsObject): string {
 function checkProperties(
     properties: Record<string, JsObject>,
     structureCount: number,
-    envCount: number
+    envCount: number,
+    parameters?: Record<string, JsObject>
 ) {
     for (const key in properties) {
         const property = properties[key];
@@ -343,13 +365,61 @@ function checkProperties(
         }
 
         const initial = typeof property.values[0];
-        if (initial !== 'string' && initial !== 'number') {
-            throw Error(`'properties['${key}'].values' should contain string or number`);
+        if (
+            initial !== 'string' &&
+            initial !== 'number' &&
+            !isMultidimensional(property.values as number[][])
+        ) {
+            throw Error(
+                `'properties['${key}'].values' should contain string or number or an array of numbers`
+            );
         }
 
         for (const value of property.values) {
             if (typeof value !== initial) {
                 throw Error(`'properties['${key}'].values' should be of a single type`);
+            }
+        }
+
+        // few checks on multidimensional properties
+        if (isMultidimensional(property.values as number[][])) {
+            // check if parameters exists
+            if (!parameters) {
+                throw Error(`'parameters' should be provided for 'properties[${key}]'`);
+            }
+            // check if parameter keyword exists and has the right format
+            const propertyParameters = property.parameters as string[];
+            if (!isArrayString(propertyParameters)) {
+                throw Error(`'properties['${key}'].parameters' should be an array of strings`);
+            }
+            // check if the length of parameters is 1 TODO: remove when support for multiple parameters is ready
+            if (propertyParameters.length !== 1) {
+                throw Error(`'properties['${key}'].parameters' should contain a single parameter`);
+            }
+            // check if parameters of the property exists in the parameters
+            for (const value of propertyParameters) {
+                if (!Object.keys(parameters).includes(value)) {
+                    throw Error(
+                        `parameter '${value}' of 'properties['${key}']' does not appear in the list provided parameters`
+                    );
+                }
+            }
+            // check if the length of the first array matches the length of the parameters
+            const initialValues = property.values[0] as number[];
+            for (const value of propertyParameters) {
+                const parameterValues = parameters[value].values as number[];
+                if (initialValues.length !== parameterValues.length) {
+                    throw Error(
+                        `'properties['${key}'].values' and 'parameters['${value}'].values' should have the same length`
+                    );
+                }
+            }
+
+            // check if all the multidimensial array elements have the same length
+            if (!isConsistent(property.values as number[][])) {
+                throw Error(
+                    `'properties['${key}].values' should contain arrays of the same length`
+                );
             }
         }
 
@@ -399,4 +469,31 @@ function checkEnvironments(o: JsObject[], structures: (Structure | UserStructure
 
 function isPositiveInteger(number: number): boolean {
     return Number.isInteger(number) && number >= 0;
+}
+
+function isMultidimensional(array: number[][]): boolean {
+    // check if an array is 2D
+    let result = true;
+    for (const value of array) {
+        result = Array.isArray(value) && result;
+    }
+    return result;
+}
+
+function isConsistent(array: number[][]): boolean {
+    // check if the elements of 2D array have the same length
+    const initial = array[0];
+    let result = true;
+    for (const value of array) {
+        result = value.length === initial.length && result;
+    }
+    return result;
+}
+
+function isArrayString(array: string[]): boolean {
+    let result = true;
+    for (const value of array) {
+        result = typeof value === 'string' && result;
+    }
+    return result;
 }
--- a/src/info/plotting.ts
+++ b/src/info/plotting.ts
@@ -1,0 +1,74 @@
+import Plotly from '../map/plotly/plotly-scatter';
+
+import { Data } from '../map/plotly/plotly-scatter';
+
+/**
+ * wrapper around scatter plots of Plotly.js for 2D properties
+ */
+export function plotMultidimProperties(
+    x: number[],
+    y: number[],
+    myDiv: HTMLElement,
+    plotWidth: number,
+    isStatic: boolean,
+    xlabel?: string,
+    ylabel?: string
+): void {
+    const trace = {
+        x: x,
+        y: y,
+        type: 'scatter',
+        mode: 'lines',
+    };
+
+    const layout = {
+        xref: 'paper',
+        yref: 'paper',
+        xaxis: {
+            title: xlabel,
+            titlefont: {
+                size: 12,
+            },
+            showgrid: false,
+            zeroline: true,
+            showline: true,
+            nticks: 5,
+        },
+        yaxis: {
+            title: ylabel,
+            titlefont: {
+                size: 12,
+            },
+            showgrid: false,
+            showline: true,
+            zeroline: false,
+            nticks: 4,
+        },
+        showlegend: false,
+        x: 0.2,
+        legend: {
+            y: 0.5,
+        },
+        autosize: true,
+        margin: {
+            l: plotWidth / 6.0,
+            r: 0,
+            b: plotWidth / 6.0,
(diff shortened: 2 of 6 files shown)


Luthaf:
Thanks a lot for creating the PR! I think the first thing we should discuss is how we want to store multi-dimensional properties in the datasets.

The current approach (using a separate `PropertiesND` field) works fine for DOS, and more generally whenever one wants properties where multiple values are associated with a single structure/atom. 

However there is another use case that could use the same code foundation: properties where each structure/atom is associated with a single scalar value; but parametrized by some parameter: for example a KPCovR map would be parametrized by the `alpha` parameter, or a PCA with values parametrized by `max_radial`. While we don't have to implement this "parametrized properties" use case right away, we will want to make sure the dataset interface remains stable when we want  (so that JSON files created with an older version of chemiscope can still be loaded).

There was a proposed Dataset interface in the initial message of #78 that could support both DOS style multi-dimensional properties and parametrized properties, did you consider using it?

---

Using the "parametrized" interface, a dataset containing both multi-dimensional properties and parametrized properties would look like this:

```js
dataset = {
    properties: {
        // already implemented
        "energy": {
            target: 'structure',
            values: [1.0, 2.0, 3.0],
            unit: "",
            description: "",
        },
        "KPCovR[1]": {
            // n_structures arrays each containing len(parameters["kpcovr-alpha"])
            // values
            values: [
                [0.3, 0.4, 0.5, 0.6],
                [0.2, 0.1, 0.2, 0.1],
                [0.7, 0.8, 0.9, 1.2],
            ],
            parameter: ["kpcovr-alpha"],
            /* unit & description */
        },
        "DOS": {
            // n_structures arrays each containing len(parameters["dos-energy"])
            // values
            values: [
                [0.3, 0.4, 0.5, 0.6],
                [0.2, 0.1, 0.2, 0.1],
                [0.7, 0.8, 0.9, 1.2],
            ],
            parameter: ["dos-energy"],
            /* unit & description */
        }
    },
    parameters: {
        "kpcovr-alpha": {
            values: [0.0, 0.5, 0.8],
            /* unit & description */
        },
        "dos-energy": {
            values: [-12, -11, -10, -9],
            /* unit & description */
        },
    }
}
```

---

Maybe the use cases for multi-dimensional and parametrized properties are different enough that we would want to use different code path & data storage for them; but I think it is worth to try to unify both features if possible.

ceriottm:
If we start fiddling with the format, it might be good to standardize a
bit, and to have a "version" field to make sure that the viewer is aware
of, and compatible with, the standard.

src/dataset.ts
```diff
@@ -344,6 +362,13 @@ function checkProperties(
         if ('units' in property && typeof property.units !== 'string') {
             throw Error(`'properties['${key}'].units' should contain a string`);
         }
+
+        // check if parameter keyword exist for multidimensional properties
+        if (Array.isArray(property.values[0])) {
+            if (typeof property.parameter !== 'string') {
+                throw Error(`'properties['${key}'].parameter' should contain a string`);
+            }
+        }
```
Luthaf, line 371: This should also check that each sub-array have the same size, and that this size matches the size of the parameter.

Luthaf (commented):
This looks pretty good, and mainly needs work on the UI side. One possibility would be to have a pretty small plot in the info table, and then have a way to open this plot in a modal (for example by clicking on the plot) when one needs to see it in more details.

cbenmahm:
I didn't know that force-pushing a branch would result in closing this issue

Luthaf (commented):
This seems to be missing code on the map side to handle multi-dimensional properties (i.e. what happens if someone set the color to "dos"?)

src/info/table.ts
```diff
@@ -5,12 +5,16 @@
 
 import assert from 'assert';
 
-import { Property, Target } from '../dataset';
+import { Parameter, Property, Target } from '../dataset';
 import { Indexes } from '../indexer';
+import { plotMultidimProperties } from './plotting';
 
 interface TableProperty {
```
Luthaf, line 12: Please also add comment here explaining how this is used

src/info/plotting.ts
```diff
@@ -0,0 +1,74 @@
...
+            showline: true,
+            zeroline: false,
+            nticks: 4,
+        },
+        showlegend: false,
+        x: 0.2,
+        legend: {
+            y: 0.5,
+        },
+        autosize: true,
```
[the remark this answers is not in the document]
cbenmahm: I don't think so, but I will remove it

cbenmahm:
thanks @Luthaf. I implemented you suggestions, now it is time to break it :)

src/dataset.ts
```diff
@@ -137,12 +139,18 @@ export interface Property {
      *
      * string values should represent classification results (category "A", "B"
      * or "C"); and numeric values should be use for everything else.
+     *
+     * it supports 2D properties, i.e. arrays, with plotting them as the main objective,
+     * where the first dimension corresponds to number of structures/atoms
+     * and the second dimension corresponds to the values
```
Luthaf, line 145:
```suggestion
     * 2D properties are also supported, with the goal of displaying one curve
     * for each atom/structure. In this case, the first dimension of the 2D
     * array corresponds to number of structures/atoms and the second dimension
     * corresponds to the values for each structure/atom.
```

src/dataset.ts
```diff
@@ -159,6 +167,21 @@ export interface Environment {
...
+ * These are the parameters used to charachterize the multidimensional properties
+ * if they are provided
+ * */
+export interface Parameter {
+    /** values of the parameter */
+    values: number[];
+    /** description of the parameter */
+    description?: string;
+    /** units of the elements in the property array */
+    units?: string;
```
Luthaf, line 182:
```suggestion
    /** name of the parameter */
    name?: string;
    /** units of the elements in the property array */
    units?: string;
    /** values of the parameter */
    values: number[];
```

Luthaf: We use "name" for properties, this makes it consistent

src/info/plotting.ts
```diff
@@ -0,0 +1,72 @@
...
+        },
+        margin: {
+            l: plotWidth / 6.0,
+            r: 0,
+            b: plotWidth / 6.0,
+            t: 0,
+            pad: 0,
+        },
+        width: plotWidth,
+        height: plotWidth / 1.35, // hardcoded ratio. TODO: to be changed
```
Luthaf, line 59: I think having a hardcoded ratio is fine

src/map/data.ts
```diff
@@ -86,8 +86,10 @@ function propertyToNumeric(name: string, property: Property): NumericProperty {
             values: values,
             units: property.units,
         };
+    } else if (Array.isArray(property.values[0])) {
+        // left blank to avoid raising an error with multidimensionla properties
```
Luthaf, line 90:
```suggestion
        // left blank to avoid raising an error with multidimensional properties
        return undefined;
```

src/map/data.ts
```diff
@@ -86,8 +86,10 @@ function propertyToNumeric(name: string, property: Property): NumericProperty {
...
     } else {
-        throw Error(`unexpected property type '${prop_type}'`);
+        throw Error(`unexpected property type '${prop_type}'`); // error thorwn even for 2D properties
```
Luthaf, line 92: I think you can remove this comment

Luthaf (commented):
Ok, so the JavaScript side of things looks good to me! Now, we should add a way to create such multi-dimensional properties from Python, and document them.

python/chemiscope/input.py
```diff
@@ -45,6 +46,8 @@ def create_input(
     :param dict settings: optional dictionary of settings to use when displaying
         the data. Possible entries for the ``settings`` dictionary are documented
         in the chemiscope input file reference.
+    :param dict parameters: optional dictionary of parameters of multidimensional
```
Luthaf, line 49:
```suggestion
    :param dict parameters: optional dictionary of parameters for multidimensional
```

python/chemiscope/input.py
```diff
@@ -103,6 +108,44 @@ def create_input(
     by comparing the numbers atoms and structures in the dataset to the
     length of provided list/np.ndarray.
 
+    It is possible to pass 2D properties to be plotted in the info bar by
```
Luthaf, line 111:
```suggestion
    It is possible to pass 2D properties to be displayed as curves in the info panel by
```

python/chemiscope/input.py
```diff
@@ -103,6 +108,44 @@ def create_input(
...
+    passing a 3D numpy array of values having the shape `(n_samples, 1, n_parameters)`
+    and ensuring the corresponding parameter exists. The previous example becomes:
+
+    .. code-block:: python
+
+        properties = {
+            'cheese': {
+                'target': 'atom',
+                'values': np.zeros((300, 1, 4)),
```
Luthaf, line 120: Instead of having this strange 3D hack, how about if there is a parameter field then we assume 2D properties and don't expand them? i.e. we add a flag to `_linearize` / call a different function for N-dimensional properties

python/chemiscope/input.py
```diff
@@ -103,6 +108,44 @@ def create_input(
...
+                # optional: property unit
+                'unit': 'random / fs',
+                # optional: property description
+                'description': 'a random property for example',
+                # optional: parameter keyword if the property is multidimensional
+                'paramters': ['color'],
```
Luthaf, line 126:
```suggestion
                'parameter': ['color'],
```

python/chemiscope/input.py
```diff
@@ -165,6 +208,46 @@ def create_input(
...
+        for key in parameters:
+            param = {}
+            if not isinstance(parameters[key], dict):
+                raise ValueError(
+                    f"expecting parameter {key} to be of type 'dict' not type {type(parameters[key])}"
+                )
+            if not isinstance(parameters[key]["values"], np.ndarray):
+                raise ValueError(
+                    f"expecting the 'values' of parameter {key} to be of type 'numpy.ndarray' not {type(parameters[key]['values'])}"
+                )
```
Luthaf, line 226: I would allow lists as well here, so how about converting this first with `np.asarray(parameters[key]["values"])`?

python/chemiscope/input.py
```diff
@@ -165,6 +208,46 @@ def create_input(
...
+            if len(parameters[key]["values"].shape) != 1:
+                raise ValueError(
+                    f"the 'values' of parameter {key} should be a 1D numpy array"
+                )
+            param["values"] = _typetransform(
+                list(parameters[key]["values"]), f"parameter {key}"
+            )
+            # I don't check if the lengths make sense wrt to a property
```
Luthaf, line 234:
```suggestion
```

python/chemiscope/input.py
```diff
@@ -272,15 +358,36 @@ def write_input(
...
+        ldos.reshape((ldos.shape[0], 1, ldos.shape[1]))
+        energy_grid = np.loadtxt(...) # energy grid for the DOS
+        multidim_properties = {
+            "DOS": {
+                target: "structure",
+                values: ldos,
+                parameters: ["xaxis"],
+            }
+        }
+        multidim_parameters = {
```
Luthaf, line 372:
```suggestion
        multidimensional_parameters = {
```

docs/src/tutorial/input-reference.rst
```diff
@@ -36,15 +36,38 @@ contains the following fields and values:
                 // the property target: is it defined per atom or for the full
                 // structures
                 "target": "atom" | "structure",
-                // values of the properties can either be numbers or strings.
+                // values of the properties can either be numbers, strings 
+                // or array of numbers.
                 // string properties are assumed to represent categories of
                 // data.
-                "values": [1, 2, 3, ...] | ["first", "second", "first", ...],
+                // the first diemnsion of the multidimensional property corresponds
```
Luthaf, line 43:
```suggestion
                // the first dimension of the multidimensional property corresponds
```

Luthaf (commented):
Random though while reading the code, what happens if the parameters values are not sorted?

python/chemiscope/input.py
cbenmahm: I agree

python/chemiscope/input.py
cbenmahm: in this case, I think the most straightforward way is to check is parameters exist or not, and/or for parameters only for multidimensional properties. What do you think?

Luthaf: You mean checking `properties[name]["parameters"]` right? If so, I agree this would be the most straightforward!

cbenmahm: yes that's what I meant

Luthaf (approved):
This is good to go for me! You'll need to move it out of draft mode, or we can not merge it!

PR #153, opened 2021-06-21
Added reset zoom level by double-clicking on the Molecule Viewer

### LINKED ISSUE #140: Reset zoom level in structure viewer
As discussed with @cbenmahm. 

If one scroll-zoom in such a way that the structure is no longer visible, it would be nice to have a way to reset the zoom level to its original state. The map already does this with a double click, and double click does nothing on the structure viewer, so using a double click to reset zoom seems like a good idea.

This should be relatively easy to do: one would need to register a "double click" event handler with 3Dmol, and then call `_resetView` from this event handler.

https://github.com/cosmo-epfl/chemiscope/blob/9ec79b6c604f02be30afb467a0d7da734d761ea0/src/structure/widget.ts#L882

jakublala:
Fixes #140

### DIFF
--- a/src/structure/widget.ts
+++ b/src/structure/widget.ts
@@ -639,6 +639,12 @@ export class MoleculeViewer {
             this._options.supercell[1].value = this._initialSupercell[1];
             this._options.supercell[2].value = this._initialSupercell[2];
         };
+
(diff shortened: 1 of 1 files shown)


Luthaf (approved):
Looks good!

PR #154, opened 2021-06-21
Download structure viewer as PNG

### LINKED ISSUE #115: Allow to save the structure viewer output as PNG
This should be easy, since all viewers are built on `canvas` elements, which can be rendered to an image with [`canvas.toDataURL`](https://developer.mozilla.org/fr/docs/Web/API/HTMLCanvasElement).

This should be nicer to use than using screenshots to extract structures views.

jakublala:
Fixes #115

A PNG download button is added to each structure viewer. Also, the SVG icons for PNG and SVG were edited to scale appropriately.

I agree that expanding this feature, where a full 'screenshot' of both map and structure are saved would be neat.


ceriottm:
Two comments:
(1) it would be great if the downloaded PNG had a transparent background. That makes it MUCH easier to  assemble a nice manual map in inkscape
(2) the icons in the structure viewer have a very different hover style than those for the map - this is very apparent for the PNG icon that is the same in both. would be great to take this opportunity to uniform the hover style

ceriottm:
I actually have a fix for the transparent background part, but you need to enable push from the PR

ceriottm:
further comment - would be nice if the download size was independent on the size of the widget, IDK if this one is feasible

jakublala:
> I actually have a fix for the transparent background part, but you need to enable push from the PR

I believe I am allowing edits by maintainers.

jakublala:
I have unified the design of the buttons for the map and structure - the CSS is a bit clumsy, because the mode bar buttons from Plotly are a little bit awkard to customise. The only remaining issue with the design is that the link of the Plottly buttons is larger than the actual icon (from a user experience I tried to solve this by changing the type of cursor accordingly). This then also leads to the tooltip being shown when not directly on the Plotly icons. Let me know what you think.

Regarding the download size - I will have a look but I have my doubts. It might be possible to create a separate 'virtual' viewer somehow that could re-render the structure (with the same camera config) and would export this, although this is just an idea.

ceriottm:
> I have unified the design of the buttons for the map and structure - the CSS is a bit clumsy, because the mode bar buttons from Plotly are a little bit awkard to customise. The only remaining issue with the design is that the link of the Plottly buttons is larger than the actual icon (from a user experience I tried to solve this by changing the type of cursor accordingly). This then also leads to the tooltip being shown when not directly on the Plotly icons. Let me know what you think.

I let @Luthaf  comment on the CSS. The result is at least homogeneous - perhaps I would have rather adapted the icons in the structure viewer rather than the plotly ones that are kind of standard
 
> Regarding the download size - I will have a look but I have my doubts. It might be possible to create a separate 'virtual' viewer somehow that could re-render the structure (with the same camera config) and would export this, although this is just an idea.

OK was just an idea, I'm not even sure it'd be a good idea given that the size of the widgets change depending on the grid and so it'd be hard to make it same shape while knowing which part to show

jakublala:
Alright. I agree we should wait for @Luthaf to comment on the style. I will leave this open in the meanwhile.

src/static/chemiscope.css
```diff
@@ -205,9 +208,61 @@
...
+    top: 140% !important;
+}
+
+.icon:hover + .modebar-btn::after {
+    visibility: visible;
+}
+
+.modebar-btn::before {
+    content: none !important;
+    font-family: Arial !important;
```
Luthaf, line 232: Why change the font family? Are we also changing it for the other tooltips?

src/static/chemiscope.css
```diff
@@ -205,9 +208,61 @@
     font-weight: bold;
 }
 
+/* Plotly Modebar */
+
+.modebar-group {
+    padding-left: 0 !important;
```
Luthaf, line 214: I don't enjoy having all the `!important` here, maybe we can improve this by using better scoping? Something like `.chsp-map .modebar-group {...`

src/static/chemiscope.css
```diff
@@ -205,9 +208,61 @@
     font-weight: bold;
 }
 
+/* Plotly Modebar */
```
Luthaf, line 211:
```suggestion
/* Change Plotly modebar style to match the style of the structure grid buttons */
```

Luthaf (commented):
Capture d’écran 2021-06-28 à 10 00 51

A few general comments: 

- I'd rather put the "download as PNG" icon on the structure between the circle & the "duplicate viewer" button instead of in between "duplicate" and "remove"
- the meaning of the icon is not immediately understandable for me, how about using/modifying a more standard [download icon](https://www.flaticon.com/free-icon/download_724933)?
- adding background and padding on the map modebar make it take more space, and looks strange. How about removing the background in both cases instead?
- The CSS changes also modified the open settings / hamburger menus. Was this intentional? If not I would revert this change to keep a difference between the action button and the setting menu.

ceriottm:
I'd also second making the structure icons more similar to the plotly
icons. Adding a "download arrow" on top of the PNG and SVG icons makes lots
of sense.

src/structure/grid.ts
```diff
@@ -347,7 +348,7 @@ export class ViewersGrid {
             assert(button.parentElement !== null);
             const tooltip = button.parentElement.querySelector('.chsp-tooltip');
             assert(tooltip !== null);
-            tooltip.innerHTML = toggle ? 'this is the active viewer' : 'choose as active';
+            tooltip.innerHTML = toggle ? 'This is the active viewer' : 'Choose as active';
```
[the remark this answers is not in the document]
jakublala: That's a good point. I have shortened 'Download plot as PNG' to 'Download PNG' for instance, as I feel it's unnecessary to specify plot/structure. 

And then chaning 'This is the active viewer' just to 'Active viewer' I think should be sufficient.

src/static/chemiscope.css
jakublala: I am changing it to better match with the fonts of the structure.

src/static/chemiscope.css
jakublala: Sure, I have removed `!important` everywhere I could and replaced it with `#chemiscope-map`, otherwise it didn't seem to work.

Luthaf: `#chemiscope-map` is set by the application (not the library) and not necessarily available in all deployments. You should use a class instead

src/static/chemiscope.css
```diff
@@ -205,9 +217,63 @@
...
+
+.chsp-map .icon:hover + .modebar-btn::after {
+    visibility: visible;
+}
+
+.chsp-map .modebar-btn::before {
+    content: none !important;
+}
+
+.modebar-btn .icon {
```
Luthaf, line 245:
```suggestion
.chsp-map .modebar-btn .icon {
```

Luthaf (approved):
I don't enjoy that we are modifying the plotly CSS since it can break under us at any time when they update their style. But this should be fine for now, and we can revisit it later!

PR #132, opened 2021-05-27
Add unit tests for modals & use unique ID for map modal

Luthaf:
This branch will live for 1-2 weeks, while I add unit tests as I go.

PR #156, opened 2021-06-23
Improving the Map Settings (max/min, step interval, width)

### LINKED ISSUE #124: Confusing min/max values when axis is in log mode
The min/max value for the axis are automatically set to the axis range as reported by plotly in 

https://github.com/cosmo-epfl/chemiscope/blob/eecc7349b2c31f992930ff7343086a8d98a28e62/src/map/map.ts#L1020-L1032

Using the Arginine-Dipeptide dataset as example, and CV1 as the property, in linear mode, the bounds are correctly reported as `-0.400` and `0.402`. But in log mode, the bounds are reported as `-4.011` and `-0.212`, which I find confusing. I would rather have them displayed as `10^(-4.011)` and `10^(-0.212)`, or the corresponding values `9.75e-05` and `0.6133`.

jakublala:
Fixes #124 

I would say that this is only a small enhancement from the current display of the settings. I had a few other suggestions that might improve the look/feel of the axis settings:

1. Increasing the width of the modal to better display longer strings of digits for the min/max range.
2. Adjusting the input step interval (the arrows that increase or decrease the value) based on some 'useful' range of the dataset. At the moment the step interval is always 1, which may be convenient for some datasets but not for others.
3. Allow to change the logarithm basis from 10 to 'e' (natural logarithm) as well. I am not sure if this a feature which would others find useful, but it just came to my mind. This one would take a little bit more work and integration with Plotly axes displaying.

Let me know what you think!

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -421,6 +421,7 @@ export class PropertiesMap {
                     'xaxis.autorange': true,
                 } as unknown as Layout);
             }
+            this._options.setScaleStep(this._options.x, 'x');
         };
 
         this._options.x.scale.onchange = () => {
@@ -481,6 +482,7 @@ export class PropertiesMap {
                     'yaxis.autorange': true,
                 } as unknown as Layout);
             }
+            this._options.setScaleStep(this._options.y, 'y');
         };
 
         this._options.y.scale.onchange = () => {
@@ -527,6 +529,7 @@ export class PropertiesMap {
                 'scene.zaxis.title': this._title(this._options.z.property.value),
                 'scene.zaxis.autorange': true,
             } as unknown as Layout);
+            this._options.setScaleStep(this._options.z, 'z');
         };
 
         this._options.z.scale.onchange = () => {
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -5,7 +5,7 @@
 
 import assert from 'assert';
 
-import Plotly, { Axis } from './plotly/plotly-scatter';
+import Plotly from './plotly/plotly-scatter';
 import { Config, Data, Layout, PlotlyScatterElement } from './plotly/plotly-scatter';
 
 import { Property } from '../dataset';
@@ -1143,10 +1143,10 @@ export class PropertiesMap {
         if (bounds.z !== undefined) {
             this._options.z.min.value = bounds.z[0];
             this._options.z.max.value = bounds.z[1];
+        }
 
-            if (!this._is3D()) {
-                this._updateMarkers();
-            }
+        if (!this._is3D()) {
+            this._updateMarkers();
         }
     }
     /**
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -1101,6 +1101,7 @@ export class PropertiesMap {
         console.log(this._options.y.max.value);
         console.log(this._options.z);
         console.log(this._options.z.max.value);
+        console.log();
     }
 
     /** Switch current plot from 3D back to 2D */
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -1255,6 +1255,7 @@ export class PropertiesMap {
     /** Changes the step of the arrow buttons in min/max input based on dataset range*/
     private setScaleStep(axisBounds: number[], axisName: string): void {
         if (axisBounds !== undefined) {
+            // round to 10 decimal places so it does not break in Firefox
             const step = Math.round(((axisBounds[1] - axisBounds[0]) / 20) * 10 ** 10) / 10 ** 10;
             const minElement = getByID<HTMLInputElement>(`chsp-${axisName}-min`);
             const maxElement = getByID<HTMLInputElement>(`chsp-${axisName}-max`);
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -421,9 +421,7 @@ export class PropertiesMap {
                     'xaxis.autorange': true,
                 } as unknown as Layout);
             }
-            console.log('---change of x property---');
-            console.log(this._options.x);
-            this._options.setScaleStep(this._options.x, 'x');
+            this.setScaleStep(this._getBounds().x, 'x');
         };
 
         this._options.x.scale.onchange = () => {
@@ -484,7 +482,7 @@ export class PropertiesMap {
                     'yaxis.autorange': true,
                 } as unknown as Layout);
             }
-            this._options.setScaleStep(this._options.y, 'y');
+            this.setScaleStep(this._getBounds().y, 'y');
         };
 
         this._options.y.scale.onchange = () => {
@@ -531,8 +529,9 @@ export class PropertiesMap {
                 'scene.zaxis.title': this._title(this._options.z.property.value),
                 'scene.zaxis.autorange': true,
             } as unknown as Layout);
-            console.log('---z property change---');
-            this._options.setScaleStep(this._options.z, 'z');
+            if (this._getBounds().z !== undefined) {
+                this.setScaleStep(this._getBounds().z as number[], 'z');
+            }
         };
 
         this._options.z.scale.onchange = () => {
@@ -852,10 +851,11 @@ export class PropertiesMap {
         this._plot.on('plotly_afterplot', () => this._afterplot());
         this._updateMarkers();
 
-        this.setScaleStep(this._options.x, 'x');
-        this.setScaleStep(this._options.y, 'y');
-        if (this._options.z.property !== undefined) {
-            this.setScaleStep(this._options.z, 'z');
+        const bounds = this._getBounds();
+        this.setScaleStep(bounds.x, 'x');
+        this.setScaleStep(bounds.y, 'y');
+        if (bounds.z !== undefined) {
+            this.setScaleStep(bounds.z, 'z');
         }
     }
 
@@ -1096,12 +1096,6 @@ export class PropertiesMap {
             'scene.yaxis.type': this._options.y.scale.value as Plotly.AxisType,
             'scene.zaxis.type': this._options.z.scale.value as Plotly.AxisType,
         } as unknown as Layout);
-        console.log('inside 3D switch');
-        console.log(this._options.y);
-        console.log(this._options.y.max.value);
-        console.log(this._options.z);
-        console.log(this._options.z.max.value);
-        console.log();
     }
 
     /** Switch current plot from 3D back to 2D */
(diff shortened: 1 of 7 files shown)


src/map/options.ts
```diff
@@ -420,4 +420,18 @@ export class MapOptions extends OptionsGroup {
     public colorScale(): Plotly.ColorScale {
         return COLOR_MAPS[this.palette.value];
     }
+
+    public logLinearLabelSwitch = (axis: AxisOptions, axisName: string) => {
```
Luthaf, line 424: Why a member variable with type function instead of a member function? This is not supposed to be changed directly by code using it, right?

```suggestion
    public logLinearLabelSwitch(axis: AxisOptions, axisName: string): void {
```

----

Also, this needs documentation and maybe a better name? What do you think of `setLogLabel` (using action for the function name) and document that this changes between log/linear scale label as appropriate?

src/map/options.ts
```diff
@@ -420,4 +420,18 @@ export class MapOptions extends OptionsGroup {
...
+        const optionsContainer = getByID<HTMLElement>(`chsp-extra-${axisName}`);
```
Luthaf, line 425: HTMLElement is the default

```suggestion
        const optionsContainer = getByID(`chsp-extra-${axisName}`);
```

src/map/options.ts
```diff
@@ -420,4 +420,18 @@ export class MapOptions extends OptionsGroup {
...
+        const minInputLabel = optionsContainer.getElementsByClassName('input-group-text')[1];
```
Luthaf, line 426: this feels really fragile, and would break when we change the HTML layout. How about adding an id to the labels and getting the min/max labels this way?

Also, this would be perfect for a unit test!

jakublala:
image
What do you think of this increased-width layout, @Luthaf ?

Luthaf:
Looks good, except that the color should have min & max input with the same width, and the `reset` button in a smaller space. Did you look at it in a narrow screen/window size?

jakublala:
I have tried in a narrower screen size and to be fair I need to check it more thorougly, because the modal width is not quite responsive as it should be I think. I will change the min/max color sizes and then what exactly do you mean by the `reset` button space? Do you want to place the `reset` button to the left (similar to more options), or do you want to make it wider?

Luthaf:
About the "reset" button, I just meant that it should take a space similar to the "scale" in the x/y/z axis. But instead of being on the left as the scale setting it is on the right

jakublala:
Ah, okay. Or maybe the `reset` button could be moved to the left, so then it will directly align with the scale arrangement (scale with `reset`, min with min, and max with max)

jakublala:
Also, what is the explanation for blocking user input for the z-axis range for some dataset examples and not others? In Arginine-Dipeptide, you can change the z-axis range if you have changed the property, but in others (e.g. Zeolites) it is blocked by default - is that the correct functionality, @Luthaf, or a bug?

Luthaf:
> Ah, okay. Or maybe the reset button could be moved to the left, so then it will directly align with the scale arrangement (scale with reset, min with min, and max with max)

Yeah, that could work.

>  In Arginine-Dipeptide, you can change the z-axis range if you have changed the property, but in others (e.g. Zeolites) it is blocked by default - is that the correct functionality, @Luthaf, or a bug?

That's a bug: applying the settings (which contains a value for the z axis) is missing a call to re-enable editing the settings. It is partially addressed in #144 (which have some overlap with this PR, I'll try to get it merged quickly so that we can rebase this one on top of it)

jakublala:
I have done all suggested changes, but I have issues with the z-axis property change to change the step interval (it's always referring to the min/max range from the previous property), so I will leave it for now and wait until the other PR is done if that possibly interferes with this one.

src/map/map.ts
```diff
@@ -525,10 +529,14 @@ export class PropertiesMap {
                 'scene.zaxis.title': this._title(this._options.z.property.value),
                 'scene.zaxis.autorange': true,
             } as unknown as Layout);
+            if (this._getBounds().z !== undefined) {
```
Luthaf, line 532:
```suggestion
            if (this._options.z.property.value !== '') {
```

or even better 

```suggestion
            if (this._is3D()) {
```

src/map/map.ts
```diff
@@ -842,6 +850,13 @@ export class PropertiesMap {
 
         this._plot.on('plotly_afterplot', () => this._afterplot());
         this._updateMarkers();
+
+        const bounds = this._getBounds();
```
Luthaf, line 854: you should add a short comment here explaining what this is doing and why

src/map/options.ts
```diff
@@ -323,7 +323,7 @@ export class MapOptions extends OptionsGroup {
                 );
 
                 // set width first, since setting position can influence it
-                modalDialog.style.width = `${modalDialog.offsetWidth}px`;
+                modalDialog.style.width = '35%'; //`${modalDialog.offsetWidth / 1.2}px`;
```
Luthaf, line 326: oh, I guess this is the origin of the small modal on small screens =). Why this change?

src/static/chemiscope.css
```diff
@@ -32,7 +32,7 @@
 }
 
 .chsp-extra-options-btn {
-    margin: auto;
+    margin-right: auto;
```
Luthaf, line 35: why this change? Both margin should be auto here if we want the button centred in its column

src/static/chemiscope.css
```diff
@@ -160,6 +160,10 @@
     width: auto;
 }
 
+#chsp-color-reset {
```
Luthaf, line 163: instead of doing this (which will break when the map get a unique id in #132), you should add this bit of styling in the HTML directly

tests/map/options.test.ts
```diff
@@ -30,4 +31,30 @@ describe('MapOptions', () => {
...
+        assertScaleLabel(options.x, 'x');
+        assertScaleLabel(options.y, 'y');
+        assertScaleLabel(options.z, 'z');
+
+        function assertScaleLabel(axisOptions: AxisOptions, axisName: string): void {
+            const min = getByID(`chsp-${axisName}-min-label`);
+            const max = getByID(`chsp-${axisName}-max-label`);
+
+            // change from linear (default) to log scale
+            axisOptions.scale.value = 'log';
```
Luthaf, line 48: instead of manually settings the value & calling the function, could you simulate user interaction here by setting the scale through the HTML select element?

Luthaf (commented):
I think I found a strange behavior: on firefox (not chromium or safari), I can only click once on the arrows to step min/max. I have no idea why =)

Another issue is that on small resolutions (I'm testing in iPad mode, 768x1024) the modal for 
settings is very narrow, which is not the case on the current website:

Capture d’écran 2021-06-30 à 17 15 43

src/map/options.ts
jakublala: Ok, I have reverted it back. The issue is that for narrow screen the modal is then outside of the screen. Also, I have noticed that when in iPad/iPhone mode, I cannot seem to move the modal.

jakublala: So, I have reverted the width change back to modalDialog.offsetWidth but slightly smaller. And then I added a check for the screen.width - so that when it is too narrow, the map setting will place itself rather in the middle and not on the right.

src/static/chemiscope.css
jakublala: I didn't want the button to be centered as then it leaves huge 'gaps' between the button and the rest of the rows.

jakublala: image

It seemed neater to have it on the left.

Luthaf: but then we have a lot of empty space on the right. Maybe you could increase the width of the select boxes to remove the empty space?

tests/map/options.test.ts
```diff
@@ -30,4 +31,33 @@ describe('MapOptions', () => {
...
+        assertScaleLabel(options.y, 'y');
+        assertScaleLabel(options.z, 'z');
+
+        function assertScaleLabel(axisOptions: AxisOptions, axisName: string): void {
+            const min = getByID(`chsp-${axisName}-min-label`);
+            const max = getByID(`chsp-${axisName}-max-label`);
+            const selectElement = getByID<HTMLSelectElement>(`chsp-${axisName}-scale`);
+
+            // change from linear (default) to log scale
+            selectElement.selectedIndex = 1;
```
Luthaf, line 49: why not use

```suggestion
            selectElement.value = "log";
```

tests/map/options.test.ts
```diff
@@ -30,4 +31,33 @@ describe('MapOptions', () => {
...
+
+    it('scale label for min/max switches between linear and log', () => {
+        const root = document.createElement('div');
+        const options = new MapOptions(root, DUMMY_PROPERTIES, DUMMY_CALLBACK);
+
+        assertScaleLabel(options.x, 'x');
+        assertScaleLabel(options.y, 'y');
+        assertScaleLabel(options.z, 'z');
+
+        function assertScaleLabel(axisOptions: AxisOptions, axisName: string): void {
```
Luthaf, line 43:
```suggestion
        function checkScaleLabelChange(axisOptions: AxisOptions, axisName: string): void {
```

src/map/options.ts
```diff
@@ -323,12 +323,17 @@ export class MapOptions extends OptionsGroup {
                 );
 
                 // set width first, since setting position can influence it
-                modalDialog.style.width = `${modalDialog.offsetWidth}px`;
+                modalDialog.style.width = `${modalDialog.offsetWidth / 1.5}px`;
```
Luthaf, line 326: can you add a comment about why we need to do this ` / 1.5`?

src/static/chemiscope.css
jakublala: image
good suggestion, it looks better now.

src/map/options.ts
```diff
@@ -323,12 +323,17 @@ export class MapOptions extends OptionsGroup {
...
                 // unset margins when using position: fixed
                 modalDialog.style.margin = '0';
                 modalDialog.style.position = 'fixed';
                 modalDialog.style.top = `${top}px`;
-                modalDialog.style.left = `${left}px`;
+
+                if (screen.width < 1400) {
```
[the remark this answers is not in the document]
jakublala: I changed this now to work in a slightly different way. It works based on `left` as before, but then if the `window` is smaller than 1400, it will clip to the right edge of the screen. I will commit the changes and let me know if that's alright.

tests/map/options.test.ts
jakublala: I didn't know that's an option, thanks :D

src/map/options.ts
```diff
@@ -323,12 +323,20 @@ export class MapOptions extends OptionsGroup {
...
                 modalDialog.style.margin = '0';
                 modalDialog.style.position = 'fixed';
                 modalDialog.style.top = `${top}px`;
-                modalDialog.style.left = `${left}px`;
+
+                if (window.innerWidth < 1400) {
+                    modalDialog.style.right = `10px`;
+                } else {
+                    modalDialog.style.left = `${left}px`;
+                }
```
Luthaf, line 339: The code in the library should not touch what the user gave in left/top, since the library does not know the overall layout of the different panels. 

Instead this change should live in the app, where we set this layout.

Luthaf (approved):

PR #157, opened 2021-06-23
Added info icons for property table and dataset title

### LINKED ISSUE #128: Make the property table easier to find
This arose from a discussion with @NatalyLo. The property table is hard to find if you don't know it exists, and nothing indicate the users that they can click on "structure XXX" or "atom YYY" to open this table. 

We should make it easier to find, while not being too obnoxious. One suggestion is to add another button on the structure viewer (near duplicate/close/...) that would open this table (and maybe flash the existing buttons to signal the user they can use these as well). Alternatively, clicking on an already selected point on the map could open the table and flash existing buttons (clicking on non-selected point would select it).

----

The alternative would be to create some kind of tutorial for first time users, showing them parts of the page in an interactive way. I'm a bit reluctant to this since website tutorial can be very obnoxious when you already know what you are doing. This could be alleviated by storing a cookie to know if someone already looked at the tutorial. I don't know how this would play out  with different deployments of chemiscope, although only hosting the tutorial and cookie on chemiscope.org might be an OK solution.

jakublala:
Fixes #128 

I have added a clipboard icon onto to the button (below the structure viewer). I am happy to take suggestions on the type of icon. I think this placement of the icon makes the most sense (below the structure viewer), as one wants different buttons for structure and atom.

I also thought it might be a good idea to maybe separate the actual button (made of the clipboard icon) and then the input field, although I am not sure about this one.


ceriottm:
how about an "info" icon of some kind? we could also use this close to the
dataset title to hint that there are details in there if you click
[image: image.png]

jakublala:
@ceriottm Yes, sure I can have a look around for a suitable 'info' icon. The one below seems promising.
image

What do you mean by 'closing the dataset title'?

ceriottm:
If you click on the dataset title (on top of the map) it will show a "description" field. It would make sense to also ad an "i" icon there to hint that there is something to see if you click.

jakublala:
Oh, I see. I will try and think of something neat and clear.

jakublala:
I have updated the info icon used and added a more clear representation of a clickable modal on the dataset title.

Luthaf:
I like the concept of these changes, but they feel very invasive in the current state. These icons should be hints, visible for first time users but not much more. I think the easiest way to do this would be to change the color from black to a light grey.

Capture d’écran 2021-06-28 à 10 16 54

How about moving the icon to before the structure/atom text, and grey them out to ~70%? If we need some separation between the icon and the text, a small vertical bar could do the trick.

-----

Capture d’écran 2021-06-28 à 10 19 28

I don't think we want a border around the title here. The icon should be grey again, and maybe moved to an "exponent" like location? Something like this (making the whole thing grey, not just the letter *i*):


Capture d’écran 2021-06-28 à 10 30 22

src/static/chemiscope.css
```diff
@@ -3,6 +3,19 @@
     cursor: pointer;
 }
 
+.chsp-meta div {
```
Luthaf, line 6: could you use a class here instead of applying the style to all possible div?

Luthaf (commented):
I found a small bug: while the metadata panel opens when clicking on the dataset name, clicking on the `i` icon does nothing. Other than that, this looks good to me!

Luthaf (approved):

PR #160, opened 2021-06-28
Small settings fixes

Luthaf:
This contains fixes to the map settings from #138 and #144, including only the fixes and no additional behavior change. 

I'll rebase both PR once this one is merged.

- First commit fixes loading the dataset Michele shared on Slack, where the z axis value changes but is set to an empty string. 
- Second commit generalize the idea of the first fix to apply it to all settings
- Third commit fixes the "greyed out"/disabled settings when loading settings with size/z axis. It does so by making sure the HTML elements are in the right state when connecting the settings initially. This was already done for the color, I added similar code for z axis and size.

All other changes (name of the "default" setting, ...) are left to be discussed in other PR.

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -388,7 +388,8 @@ export class PropertiesMap {
 
     /** Add all the required callback to the settings */
     private _connectSettings() {
-        // Reads a warning if negative values of log are discared and not shown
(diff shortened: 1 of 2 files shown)


jakublala (approved):
All seems good!

PR #161, opened 2021-06-29
Sending either id of DOM element or the HTML element itself

jakublala:
This is an initial step required for the Jupyter Notebook / Lab integration of Chemiscope. No functionality of regular Chemiscope is changed only that the `Config` of the `DefaultVisualizer` can also take in an HTML element rather that the ID of the HTML element. Remaining code was adapted so that the types in `Config` determine the manipulation with the HTML element/ID of the HTML element.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -387,7 +387,7 @@ function getMapSettings(settings: Partial<Settings> | undefined): SavedSettings 
     }
 }
 /** Validate HTML element */
-function isHTMLElement(o: any) {
+function isHTMLElement(o: unknown) {
     return o instanceof HTMLElement;
 }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -60,20 +60,24 @@ export interface Settings {
  * Check if `o` contains all the expected fields to be a [[Config]].
  */
 function validateConfig(o: JsObject) {
+    if (typeof o !== 'object') {
+        throw Error('the configuration must be a JavaScript object');
+    }
+
     if (!('meta' in o && (typeof o.meta === 'string' || isHTMLElement(o.meta)))) {
-        throw Error('missing "meta" key in chemiscope config');
+        throw Error('missing "meta" key in chemiscope configuration');
     }
 
     if (!('map' in o && (typeof o.map === 'string' || isHTMLElement(o.map)))) {
-        throw Error('missing "map" key in chemiscope config');
+        throw Error('missing "map" key in chemiscope configuration');
     }
 
     if (!('info' in o && (typeof o.info === 'string' || isHTMLElement(o.info)))) {
-        throw Error('missing "info" key in chemiscope config');
+        throw Error('missing "info" key in chemiscope configuration');
     }
 
     if (!('structure' in o && (typeof o.structure === 'string' || isHTMLElement(o.structure)))) {
-        throw Error('missing "structure" key in chemiscope config');
+        throw Error('missing "structure" key in chemiscope configuration');
     }
 
     if ('settings' in o) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -184,7 +184,7 @@ class DefaultVisualizer {
             dataset.structures,
             dataset.environments
         );
-        console.log('after setting up the viewers grid');
+
         if (config.loadStructure !== undefined) {
             this.structure.loadStructure = config.loadStructure;
         }
--- a/src/index.ts
+++ b/src/index.ts
@@ -76,7 +76,12 @@ function validateConfig(o: JsObject) {
         throw Error('missing "info" key in chemiscope configuration');
     }
 
-    if (!('structure' in o && (typeof o.structure === 'string' || o.structure instanceof HTMLElement))) {
+    if (
+        !(
+            'structure' in o &&
+            (typeof o.structure === 'string' || o.structure instanceof HTMLElement)
+        )
+    ) {
         throw Error('missing "structure" key in chemiscope configuration');
     }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -388,13 +388,7 @@ function getMapSettings(settings: Partial<Settings> | undefined): SavedSettings 
 }
 /** Validate HTML element */
 function isHTMLElement(o: any) {
-    return typeof HTMLElement === 'object'
-        ? o instanceof HTMLElement
-        : o &&
-              typeof o === 'object' &&
-              o !== null &&
-              o.nodeType === 1 &&
-              typeof o.nodeName === 'string';
+    return o instanceof HTMLElement;
 }
 
 export {
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,13 +37,13 @@ require('./static/chemiscope.css');
  */
 export interface Config {
     /** Id of the DOM element to use for the [[MetadataPanel|metadata display]] */
-    meta: string;
+    meta: string | HTMLElement;
     /** Id of the DOM element to use for the [[PropertiesMap|properties map]] */
-    map: string;
+    map: string | HTMLElement;
     /** Id of the DOM element to use for the [[EnvironmentInfo|environment information]] */
-    info: string;
+    info: string | HTMLElement;
     /** Id of the DOM element to use for the [[ViewersGrid|structure viewer]] */
-    structure: string;
+    structure: string | HTMLElement;
     /** Settings for the map & structure viewer */
     settings?: Partial<Settings>;
     /** Custom structure loading callback, used to set [[ViewersGrid.loadStructure]] */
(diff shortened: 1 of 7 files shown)


Luthaf:
This branch seems to have conflicts with master, could you rebase it?

src/index.ts
```diff
@@ -60,24 +60,20 @@ export interface Settings {
  * Check if `o` contains all the expected fields to be a [[Config]].
  */
 function validateConfig(o: JsObject) {
-    if (typeof o !== 'object') {
-        throw Error('the configuration must be a JavaScript object');
+    if (!('meta' in o && (typeof o.meta === 'string' || isHTMLElement(o.meta)))) {
+        throw Error('missing "meta" key in chemiscope config');
     }
```
Luthaf, line 65: Looks like you removed this while resolving conflicts, could you put it back please?

src/index.ts
```diff
@@ -60,24 +60,20 @@ export interface Settings {
...
 
-    if (!('meta' in o && typeof o.meta === 'string')) {
-        throw Error('missing "meta" key in chemiscope configuration');
```
Luthaf, line 68: same for the config => configuration change you removed while fixing merge conflicts

src/info/info.ts
```diff
@@ -63,15 +63,24 @@ export class EnvironmentInfo {
 
     /**
      * Create a new [[EnvironmentInfo]] inside the DOM element with given `id`
-     * @param id         HTML id of the DOM element where the sliders and
+     * @param element    HTML id of the DOM element where the sliders and
```
Luthaf, line 66: the documentation needs to be updated to say something like `HTML element or string 'id' of such element ...`

src/info/info.ts
```diff
@@ -63,15 +63,24 @@ export class EnvironmentInfo {
...
      */
-    constructor(id: string, properties: { [name: string]: Property }, indexer: EnvironmentIndexer) {
-        this._root = getByID(id);
+    constructor(
+        element: string | HTMLElement,
+        properties: { [name: string]: Property },
+        indexer: EnvironmentIndexer
+    ) {
+        if (typeof element !== 'string') {
+            this._root = element;
```
Luthaf, line 79:
```suggestion
            assert(element instanceof HTMLElement);
            this._root = element;
```

just in case =)

src/map/map.ts
```diff
@@ -210,13 +210,13 @@ export class PropertiesMap {
      * Create a new [[PropertiesMap]] inside the DOM element with the given HTML
      * `id`
      *
-     * @param id         HTML id of the DOM element where the map should live
+     * @param element   HTML element or HTML id of the DOM element where the name should live
```
Luthaf, line 213: copy-paste error (name => map) and same comment as above ("HTMLElement or string id ...")

src/map/map.ts
```diff
@@ -225,7 +225,12 @@ export class PropertiesMap {
         this.activeChanged = () => {};
         this._selected = new Map<GUID, MarkerData>();
 
-        this._root = getByID(config.id);
+        if (typeof config.element !== 'string') {
+            this._root = config.element;
```
Luthaf, line 229:
```suggestion
            assert(element instanceof HTMLElement);
            this._root = config.element;
```

src/metadata.ts
```diff
@@ -86,17 +86,22 @@ export class MetadataPanel {
      * Create a new [[MetadataPanel]] inside the HTML element with the given
      * `id`.
      *
-     * @param id       HTML id of the DOM element where the name should live
+     * @param element HTML element or HTML id of the DOM element where the name should live
```
Luthaf, line 89:
```suggestion
     * @param element HTML element or HTML id of the DOM element where the name of the dataset should be inserted
```

src/metadata.ts
```diff
@@ -86,17 +86,22 @@ export class MetadataPanel {
...
+    constructor(element: string | HTMLElement, metadata: Metadata) {
         // sanitize HTML, all the other field will go through markdown
         metadata.name = metadata.name.replace(/</g, '&lt;');
         metadata.name = metadata.name.replace(/>/g, '&gt;');
 
         this._guid = `chsp-${generateGUID()}`;
 
-        this._name = getByID(id);
+        if (typeof element !== 'string') {
+            this._name = element;
```
Luthaf, line 100:
```suggestion
            assert(element instanceof HTMLElement);
            this._name = element;
```

src/structure/grid.ts
```diff
@@ -144,15 +144,15 @@ export class ViewersGrid {
      * Create a new [[ViewersGrid]] inside the HTML element with the given
      * `id`
      *
-     * @param id           HTML id of the DOM element where the viewer should live
+     * @param element      HTML element or HTML id of the DOM element where the name should live
```
Luthaf, line 147: copy-paste error (name => structure grid) and same comment as above ("HTMLElement or string id ...")

src/structure/grid.ts
```diff
@@ -181,7 +181,12 @@ export class ViewersGrid {
         this.oncreate = () => {};
         this.activeChanged = () => {};
 
-        const root = getByID(id);
+        let root;
+        if (typeof element !== 'string') {
+            root = element;
```
Luthaf, line 186:
```suggestion
            assert(element instanceof HTMLElement);
            root = element;
```

src/structure/grid.ts
```diff
@@ -513,7 +518,7 @@ export class ViewersGrid {
      */
     private _setupCell(cellGUID: GUID, colNum: number, rowNum: number): string {
         const cellId = `gi-${cellGUID}`;
-        let cell = document.getElementById(cellId);
+        let cell = this._root.querySelector(`#${cellId}`) as HTMLElement;
```
Luthaf, line 521:
```suggestion
        let cell = this._root.querySelector(`#${cellId}`) as HTMLElement | null;
```

src/structure/grid.ts
```diff
@@ -650,7 +655,10 @@ export class ViewersGrid {
 
             // add a new widget if necessary
             if (!this._viewers.has(cellGUID)) {
-                const widget = new MoleculeViewer(`gi-${cellGUID}`, cellGUID);
+                const widget = new MoleculeViewer(
+                    this._root.querySelector(`#gi-${cellGUID}`) as HTMLElement,
```
Luthaf, line 659: why not a getByID here instead?

src/structure/widget.ts
```diff
@@ -168,7 +168,7 @@ export class MoleculeViewer {
      * @param id HTML element id inside which the viewer will be created
```
Luthaf, line 168: the doc needs to be updated

src/structure/widget.ts
```diff
@@ -179,7 +179,14 @@ export class MoleculeViewer {
         this.guid = 'chsp-' + guid;
 
         this._root = document.createElement('div');
-        const root = getByID(id);
+
+        let root;
+        if (typeof element !== 'string') {
+            root = element;
```
Luthaf, line 185:
```suggestion
            assert(element instanceof HTMLElement);
            root = element;
```

src/utils/index.ts
```diff
@@ -52,8 +52,13 @@ export function generateGUID(): GUID {
 }
 
 /** Get an HTML element by id */
```
Luthaf, line 54: could you expand on the doc a bit? Something like
```suggestion
/** 
 * Get an HTML element by id, looking inside the `root` (by default the whole `document`).
 *
 * The generic parameter `HTMLType` can be used to cast the element to a given type (e.g. `getById<HTMLInputElement>("foo")`) . The element type is not checked by this function.
 *
 * @throws if there is not element with the given id. 
 */
```

Luthaf (commented):
Be careful when you resolve merge conflicts, you removed changes from #152

src/structure/grid.ts
jakublala: after adding the `null` assertion, the checks are unhappy about the unnecessary assertion. removing the `| null` still works and if the queryselector cannot get the element, it will still assign `null` to `cell`

src/index.ts
```diff
@@ -390,6 +390,10 @@ function getMapSettings(settings: Partial<Settings> | undefined): SavedSettings
         return {};
     }
 }
+/** Validate HTML element */
+function isHTMLElement(o: unknown) {
+    return o instanceof HTMLElement;
+}
```
Luthaf, line 396:
```suggestion
```

since the function is really small, you can move the code inline in the checks themselves

src/structure/grid.ts
Luthaf: OK, then something like this would be nicer I think (using typescript type system instead of working around it)

```suggestion
        let cell = this._root.querySelector(`#${cellId}`);
```

src/structure/widget.ts
```diff
@@ -179,7 +180,15 @@ export class MoleculeViewer {
...
-        const root = getByID(id);
+
+        let root;
+        if (typeof element !== 'string') {
+            assert(element instanceof HTMLElement);
+            root = element;
+        } else {
+            root = getByID(element);
+        }
+
```
Luthaf, line 191: Since this is repeated multiple times, should we make it into its own function? Something like

```typescript
function getElement(element: string | HTMLElement) {
    if (typeof element !== 'string') {
        assert(element instanceof HTMLElement);
        return element;
    } else {
        return getByID(element);
    }
}
```

Luthaf (commented):
Some final style & readability comments. Feel free to argue if you disagree!

src/structure/widget.ts
jakublala: Good call.

src/utils/index.ts
```diff
@@ -117,3 +128,13 @@ export function arrayMaxMin(values: number[]): { max: number; min: number } {
 export function unreachable(): never {
     throw Error('unreachable code entered, this is a bug');
 }
+
+/** Returns the element based on ID or the element itself*/
+export function getElement<HTMLElement>(element: string | HTMLElement) {
```
Luthaf, line 133:
```suggestion
export function getElement<HTMLElement>(element: string | HTMLElement): void {
```

Luthaf (approved):
Looks good, please add the final change before merging!

PR #166, opened 2021-07-06
Allow to only include a subset of environment in the dataset

ceriottm:
Improve the atom environment navigation functionality by storing it in a sparse format in the JSON,
and making it clear how to navigate a structure/property map when only a few atoms are actual centers. 

These would be features that should be included. Perhaps the PR can actually be merged at intermediate steps. 

- [x] adapt python preprocessor to generate a json in the appropriate format
- [x] fix map to function with sparse environment list
- [x] fix structure to highlight clickable atoms and only allow clicking on valid centers


Luthaf:
> build an UI element to switch between structure and environment view

That's #1, and should be done as a separate PR from the sparse environments.

Luthaf:
Just finished cleaning up and rebasing the Python side of the PR, and implementing the website part of it. Here are some example input files if you want to try it:

Created with the code in `chemiscope/python/examples/molecular_map.py`
[C3H5OH.json.gz](https://github.com/cosmo-epfl/chemiscope/files/6856993/C3H5OH.json.gz)

Using the example zeolite file and removing the NaN values with [this script](https://gist.github.com/Luthaf/0798757483390aae0bb73e245604eeb9) (I tried to make the script re-usable in case there are other NaN-filled files around)
[Zeolites.json.gz](https://github.com/cosmo-epfl/chemiscope/files/6857000/Zeolites.json.gz)

@ceriottm would you have time to review/try to break this? Otherwise I'll find someone else!

Luthaf:
Also, TypeScript really shines in this kind of code change! I basically changed the return type of `EnvironmentIndexer.from_structure_atom` to also return undefined, and then let the compiler tell me where I needed to make changes!

ceriottm:
I don't have time to do a proper review in the near future but I downloaded
and tried it. Works, it seems but I have some comments
1. displaying an error that must be clicked away when trying to select a
missing environment seems overkill. we could think of something less
intrusive (see also 3. below)
2. the environment slider should only run over the existing environments, I
don't see the point of having it index also "missing" environments.
3. I don't know if it's feasible but it'd be ideal if environments center
atoms would display a faint highlight sphere upon hover. then, clicking on
a non-environment center could be silently ignored.

Luthaf:
> 1. displaying an error that must be clicked away when trying to select a
missing environment seems overkill.

I agree, this might be alleviated by #176 as well. I think we have to to tell the user in some way that they last action did nothing because the environment is missing, but a warning might not be the best way.

> 2. the environment slider should only run over the existing environments, I
don't see the point of having it index also "missing" environments.

That makes sense, but the box displaying the numeric id of the atom still needs to deal with the user inputing the id of a missing atom.

> 3. I don't know if it's feasible but it'd be ideal if environments center
atoms would display a faint highlight sphere upon hover. then, clicking on
a non-environment center could be silently ignored.

This should be doable, since 3Dmol exposes a [`setHoverable`](https://3dmol.csb.pitt.edu/doc/$3Dmol.GLViewer.html#setHoverable) function. I'll see how much work it is to make it work!

Luthaf:
Also, still to-do:

- [ ] add user-facing documentation for the new input functionalities
- [ ] add unit tests for the `EnvironmentIndexer`

ceriottm:
Can we resurrect this?

Luthaf:
I won't have time for this until January. If you wanna help, you can write some of the user facing documentation (docstring & something in this section of sphinx doc: https://chemiscope.org/docs/tutorial/index.html)

ceriottm:
There's an issue (IDK if solvable) with hover. The hover selects an atom even if it is behind another (non-clickable) atom. However, the click won't select it unless it is in the foreground. This can get confusing, particularly if the atom in the foreground that is "covering" the one you are trying to select is semi-transparent. Ideally, it should be possible to "click through" just as you can "hover through" non-selectable atoms.

ceriottm:
Another "bug or feature depending how you see it" - you can hover on atoms outside the unit cell, but when you click it'll select the corresponding atom in the unit cell. Personally I'd rather have only atoms in the central box as hoverable/selectable.

Luthaf:
> There's an issue (IDK if solvable) with hover. The hover selects an atom even if it is behind another (non-clickable) atom. However, the click won't select it unless it is in the foreground. This can get confusing, particularly if the atom in the foreground that is "covering" the one you are trying to select is semi-transparent. Ideally, it should be possible to "click through" just as you can "hover through" non-selectable atoms.

This would require changes on 3Dmol.js, so I think we might have to live with it for now

> Another "bug or feature depending how you see it" - you can hover on atoms outside the unit cell, but when you click it'll select the corresponding atom in the unit cell. Personally I'd rather have only atoms in the central box as hoverable/selectable.

I'll have a go!

Luthaf:
Ok, this is good to go on my side. @ceriottm do you want to test it again?

src/structure/widget.ts
```diff
@@ -1007,7 +1007,10 @@ export class MoleculeViewer {
      */
     private _resetView(center: boolean): void {
         if (center && this._highlighted !== undefined) {
-            this._viewer.zoomTo({ serial: this._highlighted.center });
+            // use index rather than serial to specify the selection, to avoid picking also the 
+            // periodic replicas. however we then have to specify the model id, otherwise it'd 
+            // pick the index from the highlighted selection, which does not match the serial ID
+            this._viewer.zoomTo({ index: this._highlighted.center, model: 0 });
```
Luthaf, line 1013: The model number is not guaranteed to be 0, you should use `this._current.model` or something like this

ceriottm: actually you were right. works

PR #168, opened 2021-07-07
Included warning whenever inserted min > max

### LINKED ISSUE #155: Maximum value for range in map settings may be smaller than the minimum
If one manually sets the minimum lower than the maximum, or the maximum higher than the minimum, then the axis becomes flipped. I am not sure if this done on purpose (so that one can easily flip the axis, if one wants to), although at least a warning could be displayed, or max and min values should limit one another (max being always bigger than min).

fgrassel:
Fix #155

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -438,7 +438,7 @@ export class PropertiesMap {
 
         // function creating a function to be used as onchange callback
         // for <axis>.min and <axis>.max
-        const rangeChange = (name: string, axis: AxisOptions) => {
+        const rangeChange = (name: string, axis: AxisOptions, minOrMax: 'min' | 'max') => {
             return (_: number, origin: OptionModificationOrigin) => {
                 if (origin === 'JS') {
                     // prevent recursion: this function calls relayout, which then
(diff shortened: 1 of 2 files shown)


src/map/map.ts
```diff
@@ -447,6 +447,14 @@ export class PropertiesMap {
                 }
                 const min = axis.min.value;
                 const max = axis.max.value;
+                if (min > max) {
+                    sendWarning(
+                        `${name}: inserted min and max values are such that min > max! Values are resetted to the last ones inserted.`
```
jakublala, line 452:
```suggestion
                `Inserted min value is greater than max value for the ${name}!
                         Values are resetted to the last ones inserted.`
```
I would change the warnign message to this so it is more alligned with other error messages and also is not on a single long line in the source code file.

jakublala, line 452:
```suggestion
                        `Inserted min value is greater than max value for the ${name}!`
```
Simplifying the warning message. The 'reset' functionality is obvious from the user using it in my opinion.

jakublala:
1) There seems to be an error when increasing min for x-axis and y-axis with the arrows. The max values start jumping between two different values, if you keep on increasing min.
2) For z-axis, a similar weird behaviour is observed, but the max starts jumping between 0 and the actual maximum.

It's possible these 'bugs' have the same origin.

jakublala (approved):
All good!

PR #170, opened 2021-07-08
Fix tests on node 16

### LINKED ISSUE #165: tests fail on node 16
Some tests in tests/map/ fail when using node 16, but they pass on node 14

Luthaf:
fixes #165 

This also adds node 14 and 16 to the CI to make sure we catch this kind of issue early!


jakublala (approved):
All good!

PR #172, opened 2021-07-08
Feat/shortened properties

serfg:
Adding the possibility to use shortened description of the properties in write_input command of python API.

### DIFF
--- a/python/chemiscope/input.py
+++ b/python/chemiscope/input.py
@@ -57,43 +57,33 @@ def _convert(properties, n_structures, n_atoms):
     total number of atoms in the dataset is 300.
     """
 
-    for key in properties.keys():
-        if type(properties[key]) is not dict:
-            if (type(properties[key]) is not list) and (
-                type(properties[key]) is not np.ndarray
-            ):
+    for key, value in properties.items():
+        if not isinstance(value, dict):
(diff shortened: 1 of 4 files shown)


python/chemiscope/input.py
```diff
@@ -12,6 +12,91 @@
...
+                'target': 'atom',
+                'values': np.zeros((300, 17)),
+           }
+        }
+
+    assuming that number of structures in the dataset is 100 and
+    total number of atoms in the dataset is 300.
+    """
+
+    for key in properties.keys():
```
Luthaf, line 60:
```suggestion
    for key, value in properties.items():
```

Luthaf:
Could you also add some tests checking this functionality as well?

Luthaf:
And could you also add user documentation, changing the example in the docstring of `write_input` and `create_input` to include the new "shortened" properties?

Luthaf:
> How does this interact with sparse_env?

Instead of checking if the size of the array matches the number of atom, we can check if it matches the number of sparse environments.

Luthaf (approved):
I made a few small changes
- moved the test readme together with the main python readme
- small error message improvements

This should be good to go now, I'll merge once CI is happy!

Luthaf:
merged! Thanks a lot @serfg for taking the time to do this!

PR #173, opened 2021-07-08
Integration of Karma testing framework

jakublala:
Integrating `karma` instead of `jsdom` to actually run the tests in a browser so that WebGL works (and one can thus write tests for the Viewer or the Map).


jakublala:
Also, I added a test for the Viewer so that one can actually see that the WebGL stuff works with `karma`.

.github/workflows/tests.yml
```diff
@@ -20,7 +20,8 @@ jobs:
           node-version: '12.x'
       - uses: bahmutov/npm-install@v1
       - run: npm run build
-      - run: npm test
+      - run: sudo apt-get install xvfb
+      - run: xvfb-run --auto-servernum npm test
```
Luthaf, line 24:
```suggestion
      - run: sudo apt-get install xvfb
      # use xvfb to run the tests with a pseudo-display attached
      # so that the browsers started by karma think there is a screen
      - run: xvfb-run --auto-servernum npm test
```

package.json
```diff
@@ -31,36 +31,41 @@
...
+    "@types/tmp": "^0.2.1",
+    "@typescript-eslint/eslint-plugin": "^4.28.2",
+    "@typescript-eslint/parser": "^4.28.2",
     "3dmol": "^1.6.4",
     "assert": "^2.0.0",
     "bootstrap": "^4.6.0",
     "bubleify": "^2.0.0",
     "buffer": "^6.0.2",
-    "canvas": "^2.6.1",
+    "canvas": "^2.8.0",
```
Luthaf, line 50: I don't think we need canvas anymore, it was used to provide a WegGL context inside node for the previous tests.

package.json
```diff
@@ -31,36 +31,41 @@
...
     "ify-loader": "^1.1.0",
     "jquery": "^3.6.0",
-    "jsdom": "^16.4.0",
-    "markdown-it": "^12.0.2",
+    "karma": "^6.3.4",
+    "karma-chrome-launcher": "^3.1.0",
+    "karma-detect-browsers": "^2.3.3",
+    "karma-firefox-launcher": "^2.1.1",
+    "karma-mocha": "^2.0.1",
+    "karma-safari-launcher": "^1.0.0",
```
Luthaf, line 65: did you try this with edge as well?

tests/karma.conf.ts
```diff
@@ -0,0 +1,44 @@
...
+import path from 'path';
+
+module.exports = (config: any) => {
+    config.set({
+        autoWatch: false,
+        browserNoActivityTimeout: 10000,
+        concurrency: 1,
+        client: {
+            mocha: {
+                timeout: 8000,
```
Luthaf, line 11: this should not be needed here, I use it in chemfiles because loading the WASM library takes a while

tests/karma.conf.ts
```diff
@@ -0,0 +1,44 @@
+import { BASE_CONFIG } from '../webpack.config';
+import path from 'path';
+
+module.exports = (config: any) => {
+    config.set({
+        autoWatch: false,
+        browserNoActivityTimeout: 10000,
```
Luthaf, line 7: same, this should not be needed here

tests/karma.conf.ts
```diff
@@ -0,0 +1,44 @@
...
+
+        detectBrowsers: {
+            postDetection: function (availableBrowsers: any) {
+                //Remove IE
+                var result = availableBrowsers;
+                let IEindex = result.indexOf('IE');
+                if (IEindex !== -1) {
+                    result.splice(IEindex);
+                }
+                return result;
```
Luthaf, line 38:
```suggestion
                // Remove IE
                const IEindex = availableBrowsers.indexOf('IE');
                if (IEindex !== -1) {
                    availableBrowsers.splice(IEindex);
                }
                return availableBrowsers;
```

tests/map/options.test.ts
```diff
@@ -1,9 +1,10 @@
 import { MapOptions, AxisOptions } from '../../src/map/options';
 import { getByID } from '../../src/utils';
 
-import { default as setupJSDOM } from '../jsdom';
 import { assert } from 'chai';
 
+let documentHTML: string;
```
Luthaf, line 6:
```suggestion
let KARMA_INSERTED_HTML: string;
```

tests/structure/grid.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    {
+        size: 2,
+        names: ['X', 'Y'],
+        x: [0, 1],
+        y: [0, 1],
+        z: [0, 1],
+    },
+];
+
+const DUMMY_ENVIRONMENTS = {};
```
Luthaf, line 22: This does not seems to be used anywhere

tests/structure/grid.test.ts
```diff
@@ -0,0 +1,45 @@
+import { assert } from 'chai';
+
+import { default as $ } from 'jquery';
+(global as any).$ = $;
+
+import { ViewersGrid } from '../../src/structure/index';
+import { EnvironmentIndexer } from '../../src/indexer';
+import { MoleculeViewer } from '../../src/structure/widget';
+
+let documentHTML: string;
```
Luthaf, line 10:
```suggestion
let KARMA_INSERTED_HTML: string;
```

tests/structure/grid.test.ts
```diff
@@ -0,0 +1,45 @@
+import { assert } from 'chai';
+
+import { default as $ } from 'jquery';
+(global as any).$ = $;
```
Luthaf, line 4: We should fix this properly instead of hacking around it. Maybe we can remove this tests from this PR, and move it to a separate one where we also fix the issue?

tests/karma.conf.ts
```diff
@@ -0,0 +1,44 @@
...
+            },
+            preferHeadless: false,
```
Luthaf, line 40: I'd really like to enable headless testing, I'll check if I can find a way to do it locally

tests/structure/grid.test.ts
jakublala: That would mean we can't write any tests for the structure viewer until we fix this, right?

tests/karma.conf.ts
jakublala: Ok.

Luthaf:
I have an issue when trying to run the tests locally:

```
09 07 2021 14:30:59.759:ERROR [karma-server]: Error: ENOENT: no such file or directory, open '/var/folders/v6/c9ym3wnn4c3c2t3r9sv4v02h0000gn/T/_karma_webpack_713295/../../../../../../Users/guillaume/code/chemiscope/dist/src/structure/grid.d.ts'
```

I'd guess this is related to webpack or karma configuration, I'll check what's going on

package.json
```diff
@@ -31,36 +31,40 @@
...
-    "markdown-it": "^12.0.2",
-    "mocha": "^8.1.2",
-    "mochapack": "^2.1.2",
+    "karma": "^6.3.4",
+    "karma-chrome-launcher": "^3.1.0",
+    "karma-detect-browsers": "^2.3.3",
+    "karma-firefox-launcher": "^2.1.1",
+    "karma-mocha": "^2.0.1",
+    "karma-safari-launcher": "^1.0.0",
+    "karma-webpack": "git+https://github.com/ryanclark/karma-webpack#ef7edb6b6756fb563871eaff88ca876892694896",
```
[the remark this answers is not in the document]
ceriottm: incredibly, works on my system

Luthaf:
I though I already merged this 😄 Well, now it is done!

PR #177, opened 2021-07-09
Unit tests for Map Options

jakublala:
Made a few unit tests for Map Options - it's primarily checking selecting/clicking on the elements. Please, have a look if the unit tests make sense, or if I should do them differently.

Happy to do some more, but I would appreciate some guidance - let's say a list of things we want and can test with this basic framework setup.

### DIFF
--- a/tests/map/map.test.ts
+++ b/tests/map/map.test.ts
@@ -1,0 +1,124 @@
+import { PropertiesMap } from '../../src/map';
+import { EnvironmentIndexer } from '../../src/indexer';
+import { Target } from '../../src/dataset';
+import { getByID } from '../../src/utils';
+
+import { assert } from 'chai';
+
+let KARMA_INSERTED_HTML: string;
+
+const DUMMY_PROPERTIES = {
+    first: {
+        target: 'structure' as Target,
+        values: [1.1, 1.2],
+    },
+    second: {
+        target: 'structure' as Target,
+        values: [2.1, 2.2],
+    },
+    third: {
+        target: 'structure' as Target,
+        values: [3.1, 3.2],
+    },
+};
+
+const DUMMY_MAP_SETTINGS = {
+    x: {
+        max: 0,
+        min: 10,
+        property: 'first',
+        scale: 'linear',
+    },
+    y: {
+        max: 0,
+        min: 10,
+        property: 'first',
+        scale: 'linear',
+    },
+    z: {
+        max: 0,
+        min: 10,
+        property: 'first',
+        scale: 'linear',
+    },
+    color: {
+        max: 0,
+        min: 10,
+        property: 'second',
+        scale: 'linear',
+    },
+    symbol: '',
+    palette: 'hsv (periodic)',
+    size: {
+        factor: 50,
+        property: '',
+    },
+};
+
+const DUMMY_STRUCTURES = [
+    {
+        size: 2,
+        names: ['X', 'Y'],
+        x: [0, 1],
+        y: [0, 1],
+        z: [0, 1],
+    },
+];
+
+describe('Map', () => {
+    before(() => {
+        // store karma's default HTML
+        KARMA_INSERTED_HTML = document.body.innerHTML;
+    });
+
+    it('can remove itself from DOM', () => {
+        const root = document.createElement('div');
+        const indexer = new EnvironmentIndexer('structure', DUMMY_STRUCTURES);
+        const map = new PropertiesMap(
+            { element: root, settings: DUMMY_MAP_SETTINGS },
+            indexer,
+            DUMMY_PROPERTIES
+        );
+
+        assert(root.innerHTML !== '');
+        assert(document.body.innerHTML !== KARMA_INSERTED_HTML);
+
+        map.remove();
+        assert(root.innerHTML === '');
+
+        assert(document.body.innerHTML === KARMA_INSERTED_HTML);
+    });
+
+    it('color range resets when clicked', () => {
+        const root = document.createElement('div');
+        const indexer = new EnvironmentIndexer('structure', DUMMY_STRUCTURES);
+        const map = new PropertiesMap(
+            { element: root, settings: DUMMY_MAP_SETTINGS },
+            indexer,
+            DUMMY_PROPERTIES
+        );
+        const options = map['_options'];
+
+        const minSelectElement = getByID<HTMLSelectElement>(options.getId(`map-color-min`));
+        const maxSelectElement = getByID<HTMLSelectElement>(options.getId(`map-color-max`));
+        const originalMin = options.color.min.value;
+        const originalMax = options.color.max.value;
+
+        minSelectElement.value = '1.0';
+        minSelectElement.dispatchEvent(new window.Event('change'));
+        assert(options.color.min.value === 1.0);
+
+        maxSelectElement.value = '3.0';
+        maxSelectElement.dispatchEvent(new window.Event('change'));
+        assert(options.color.max.value === 3.0);
+
+        const resetButton = map['_colorReset'];
+        resetButton.click();
+        resetButton.dispatchEvent(new window.Event('click'));
+
+        assert(options.color.min.value === originalMin);
+        assert(options.color.max.value === originalMax);
+
+        map.remove();
+    });
+});
--- a/tests/map/options.test.ts
+++ b/tests/map/options.test.ts
@@ -12,6 +12,9 @@ const DUMMY_PROPERTIES = {
     second: {
         values: [],
     },
+    third: {
+        values: [],
+    },
 };
 
 const DUMMY_CALLBACK = () => {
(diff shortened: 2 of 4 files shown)


Luthaf:
Could you rebase this branch now that the karma framework is merged?

jakublala:
Rebased and added/finished a color reset unit test. I am happy to leave this PR open, but I am not sure whether I will be eager to add more unit tests (in the near future), so it might be good just to get it approved and merge it - so that it lives in the repository as a good 'guidance'/template for other people collaborating and adding their own unit tests.

tests/map/map.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        const root = document.createElement('div');
+        const indexer = new EnvironmentIndexer('structure', DUMMY_STRUCTURES);
+        const map = new PropertiesMap(
+            { element: root, settings: DUMMY_MAP_SETTINGS },
+            indexer,
+            DUMMY_PROPERTIES
+        );
+
+        assert(root.innerHTML !== '');
+        assert(document.body.innerHTML !== '');
```
Luthaf, line 85:
```suggestion
        assert(document.body.innerHTML !== KARMA_INSERTED_HTML);
```

tests/map/map.test.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        map.remove();
+        assert(root.innerHTML === '');
+
+        // remove SVG element created by Plotly
```
Luthaf, line 90: this should be part of the map `remove` function

tests/map/map.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        resetButton.dispatchEvent(new window.Event('click'));
+
+        assert(options.color.min.value === originalMin);
+        assert(options.color.max.value === originalMax);
+
+        removeMap(map);
+    });
+});
+
+/** Removes the map and an additional HTML element creted by Plotly */
```
Luthaf, line 129: this function does not seems to be used

tests/map/options.test.ts
```diff
@@ -33,7 +36,7 @@ describe('MapOptions', () => {
         assert(document.body.innerHTML === KARMA_INSERTED_HTML);
     });
 
-    it('scale label for min/max switches between linear and log', () => {
+    it('selected scale label for min/max switches between linear and log', () => {
```
Luthaf, line 39: Not sure why this change is here, the previous test name was clearer IMO

tests/map/options.test.ts
```diff
@@ -61,4 +64,63 @@ describe('MapOptions', () => {
...
+        const options = new MapOptions(root, DUMMY_PROPERTIES, DUMMY_CALLBACK);
+
+        checkPropertySelect(options.x, 'x');
+        checkPropertySelect(options.y, 'y');
+        checkPropertySelect(options.z, 'z');
+        checkPropertySelect(options.color, 'color');
+
+        function checkPropertySelect(axisOptions: AxisOptions, axisName: string) {
+            const selectElement = getByID<HTMLSelectElement>(`chsp-${axisName}`);
+            selectElement.value = 'second';
```
Luthaf, line 79: We want to check that the value changes, so we need to check that the new & previous value are different, with something like this:

```suggestion
            assert(selectElement.value !== 'second');
            assert(axisOptions.property.value !== 'second');
            selectElement.value = 'second';
```

tests/map/options.test.ts
```diff
@@ -61,4 +64,63 @@ describe('MapOptions', () => {
...
+            selectElement.dispatchEvent(new window.Event('change'));
+            assert(axisOptions.property.value === 'second');
+        }
+    });
+
+    it('selected range changes when changed', () => {
```
Luthaf, line 85:
```suggestion
    it('axis range changes when min/max are modified', () => {
```

tests/map/options.test.ts
```diff
@@ -61,4 +64,63 @@ describe('MapOptions', () => {
             assert(max.innerHTML === 'max:');
         }
     });
+
+    it('selected property changes when changed', () => {
```
Luthaf, line 68:
```suggestion
    it('axis property changes when modified by the user', () => {
```

tests/map/options.test.ts
```diff
@@ -61,4 +64,63 @@ describe('MapOptions', () => {
...
+        }
+    });
+
+    it('selected range changes when changed', () => {
+        const root = document.createElement('div');
+        const options = new MapOptions(root, DUMMY_PROPERTIES, DUMMY_CALLBACK);
+
+        checkRangeSelect(options.x, 'x');
+        checkRangeSelect(options.y, 'y');
+        checkRangeSelect(options.z, 'z');
```
Luthaf, line 91: why not the color here?

tests/map/options.test.ts
```diff
@@ -61,4 +64,63 @@ describe('MapOptions', () => {
...
+
+        function checkRangeSelect(axisOptions: AxisOptions, axisName: string) {
+            const minSelectElement = getByID<HTMLSelectElement>(`chsp-${axisName}-min`);
+            const maxSelectElement = getByID<HTMLSelectElement>(`chsp-${axisName}-max`);
+
+            minSelectElement.value = '0.01';
```
Luthaf, line 97: same as above, this needs to check the previous value as well

tests/map/options.test.ts
```diff
@@ -61,4 +64,63 @@ describe('MapOptions', () => {
...
+            assert(axisOptions.max.value === 1.01);
+        }
+    });
+
+    it('selected size changes when changed', () => {
+        const root = document.createElement('div');
+        const options = new MapOptions(root, DUMMY_PROPERTIES, DUMMY_CALLBACK);
+        const selectElement = getByID<HTMLSelectElement>(`chsp-size`);
+
+        selectElement.value = 'second';
```
Luthaf, line 112: same as above, this needs to check the previous value as well

tests/map/options.test.ts
```diff
@@ -61,4 +64,63 @@ describe('MapOptions', () => {
...
+        selectElement.dispatchEvent(new window.Event('change'));
+        assert(options.size.property.value === 'second');
+    });
+
+    it('slider size changes when changed', () => {
+        const root = document.createElement('div');
+        const options = new MapOptions(root, DUMMY_PROPERTIES, DUMMY_CALLBACK);
+        const sliderElement = getByID<HTMLInputElement>(`chsp-size-factor`);
+
+        sliderElement.value = '100';
```
Luthaf, line 122: same as above, this needs to check the previous value as well

tests/map/options.test.ts
```diff
@@ -61,4 +64,63 @@ describe('MapOptions', () => {
...
+            minSelectElement.dispatchEvent(new window.Event('change'));
+            assert(axisOptions.min.value === 0.01);
+
+            maxSelectElement.value = '1.01';
+            maxSelectElement.dispatchEvent(new window.Event('change'));
+            assert(axisOptions.max.value === 1.01);
+        }
+    });
+
+    it('selected size changes when changed', () => {
```
Luthaf, line 107:
```suggestion
    it('property used for size changes when modified by the user', () => {
```

tests/map/options.test.ts
```diff
@@ -61,4 +64,63 @@ describe('MapOptions', () => {
...
+        const root = document.createElement('div');
+        const options = new MapOptions(root, DUMMY_PROPERTIES, DUMMY_CALLBACK);
+        const selectElement = getByID<HTMLSelectElement>(`chsp-size`);
+
+        selectElement.value = 'second';
+        selectElement.dispatchEvent(new window.Event('change'));
+        assert(options.size.property.value === 'second');
+    });
+
+    it('slider size changes when changed', () => {
```
Luthaf, line 117:
```suggestion
    it('size factor changes when the slider is modified by the user', () => {
```

tests/map/options.test.ts
jakublala: i have changed it to 'selected' so that it aligns with the other tests, where a 'selection' is always made on the HTML select element

jakublala: but seeing that you changed the others to something clearer, then i will revert this one

tests/map/options.test.ts
Luthaf: Still an open question =)

tests/map/map.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        minSelectElement.dispatchEvent(new window.Event('change'));
+        assert(options.color.min.value === 1.0);
+
+        maxSelectElement.value = '3.0';
+        maxSelectElement.dispatchEvent(new window.Event('change'));
+        assert(options.color.max.value === 3.0);
+
+        const resetButton = map['_colorReset'];
+        resetButton.click();
+        resetButton.dispatchEvent(new window.Event('click'));
```
Luthaf, line 118: Why do we need both? `click()` does not propagate the event?

jakublala: nope. click() itself does not propagate and one needs to dispatch the event for the callback function to be called

tests/map/options.test.ts
jakublala: sorry, i didn't notice it earlier. i just added a new commit

Luthaf (commented):
Looks good, given that I just merged #132 I'll manually rebase this PR to check it locally before merging

Luthaf (approved):

PR #178, opened 2021-07-12
Environment disable bugfix

### LINKED ISSUE #92: Environment disabling setting is lost when changing structure
If one click on "disable" for the "Environment" setting in the structure visualizer, the setting is reset when loading a new structure.

jakublala:
Fixes #92 

That part of the code seems to me quite confusing, but hopefully this short/simple bugfix does the necessary job. I definitely need someone to properly test it out, but it seemed to function correctly on my side.

### DIFF
--- a/src/structure/widget.ts
+++ b/src/structure/widget.ts
@@ -865,15 +865,9 @@ export class MoleculeViewer {
         }
 
         if (center === undefined) {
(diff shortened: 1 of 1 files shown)


Luthaf:
I pushed a simpler fix =) The code around here should not have to touch the settings in any way.

PR #181, opened 2021-07-14
Jupyter Widget

jakublala:
To check this one out:

1. Clone the branch
2. Run `npm ci` to get/update packages
2. Run `npm run build`
3. Run `cd python && pip install --use-feature=in-tree-build .`
(4.5 Run `jupyter nbextension install --sys-prefix chemiscope` and `jupyter nbextension enable --sys-prefix chemiscope` before we fix this)
4. Open a Jupyter Notebook
5. Import `chemiscope`
5. Use `chemiscope.display(frames, properties)`
6. Observe and be impressed.


jakublala:
Current issues/improvements:

- [ ]  finalise map v. structure layout and ensure it works on different screen sizes
- [ ] adjust icon sizes to reflect the overall size of the widget (i.e. make them smaller)
- [ ] add unique IDs to modals, so more widget instances can be used at once
- [ ] test functionality of more widget instances

ceriottm:
Two issues I found with the current version:
(1) to enable chemiscope widget I had to do
`jupyter nbextension enable chemiscope --user --py`
(2) there is a disappearing marker issue similar to what we saw long ago in mainline chemiscope (after zooming in and out the plotly window, sometimes the active marker doesn't show up anymore) only closing and reopening the structure viewer brings it back

ceriottm:
Idea - might be nice to allow showing only one panel. That way we can also use this to navigate plots where there are no associated structures, or to visualize a set of structures for which we haven't computed projections.

Luthaf:
I just pushed a revamp of the Python build system, which should hopefully be more robust now. Any combination of `pip install .`/`python setup.py sdist && pip install dist/python/chemiscope-*.tar.gz`/`python setup.py bdist_wheel && pip install dist/python/chemiscope-*.whl`/`python setup.py install` should now work. 

This also removes the need for `--use-feature=in-tree-build` and reliance on recent pip version.

>  (1) to enable chemiscope widget I had to do

One file was missing from the install, this should now be fixed.

> Idea - might be nice to allow showing only one panel. That way we can also use this to navigate plots where there are no associated structures, or to visualize a set of structures for which we haven't computed projections.

I'm not sure I see the point of this for users. The whole value proposition of chemiscope is simultaneous exploration of the map & structure. If you want to plot maps, matplotlib is doing it better. If you want to visualize structures, there are a bunch of other jupyter widgets with more configurability than chemiscope.

ceriottm:
> I'm not sure I see the point of this for users. The whole value proposition of chemiscope is simultaneous exploration of the map & structure. If you want to plot maps, matplotlib is doing it better. If you want to visualize structures, there are a bunch of other jupyter widgets with more configurability than chemiscope.

It's the convenience of having a single environment to work with, and the widgets have some functionalities other widgets don't have. For instance, the possibility of loading multiple properties and selecting them dynamically to see how they correlate. On the structure side, the work you did to add periodic structure visualization to 3dmol.js makes it quite useful. My idea would be that if you call chemiscope.display with only structures you get only the structure panel, and if you call it with only properties you get only the map panel.

Luthaf:
Ok, fair enough. That's better left as a separate PR, could you open an issue for it?

tsconfig.json
```diff
@@ -16,6 +16,7 @@
     "excludePrivate": true,
     "excludeNotExported": true,
     "hideGenerator": true,
-    "readme": "none"
+    "readme": "none",
+    "skipLibCheck": true
```
Luthaf, line 20: this should no longer be required

webpack.config.ts
```diff
@@ -45,10 +46,11 @@ const config: ConfigFn = (env, argv) => {
             chemiscope: './src/index.ts',
             'chemiscope-app': './app/app.ts',
             'molecule-viewer': './src/structure/widget.ts',
-        },
+        } as { [key: string]: string },
```
Luthaf, line 49:
```suggestion
        },
```

This should no longer be required

python/nbextension/src/widget.ts
```diff
@@ -0,0 +1,71 @@
...
+ * in the Jupyter Notebook output window when instantiated from
+ * the Chemiscope Python package.
+ */
+export class ChemiscopeView extends DOMWidgetView {
+    private visualizer?: DefaultVisualizer;
+
+    public render(): void {
+        const element = this.el;
+
+        // resize Plotly when inserted
```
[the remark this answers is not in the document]
jakublala: From how I understand this, the Jupyter output window has three width depending on the window width of the browser. For reasonably sized monitors, the three main modes are i) above 1200 pixel width, ii) between 1200 and 1000 pixel width and iii) below 1000 pixel width. The output window changes accordingly once these thresholds are reached, so one could technically compute what the output area width for each of these. Once one gets below 1000 pixel width, the output area essentially follows the width of the window (but this is a very small pixel size and I hope no one has such a small monitor anymore).

Luthaf:
Just rebased & squashed this branch on top of master. Still TBD before we can merge:

- [ ] Figure out why we need "pip install numpy ipywidgets" and why tox does not install these on CI

Luthaf:
> Figure out why we need "pip install numpy ipywidgets" and why tox does not install these on CI

This is fixed by switching to tox 4, which has proper support for PEP418 and build isolation.

Luthaf:
Merged! I'll do a release on PyPI later today! Congratulations @jakublala!

PR #182, opened 2021-07-15
Allow to load the same file multiple time

Luthaf:
If we don't clear the input value, safari and chrome don't call the "onchange" callback again.


jakublala (approved):
All good!

PR #183, opened 2021-07-15
Added configurable maxWidgets option for the ViewesrGrid

jakublala:


src/index.ts
```diff
@@ -48,6 +48,8 @@ export interface Config {
     settings?: Partial<Settings>;
     /** Custom structure loading callback, used to set [[ViewersGrid.loadStructure]] */
     loadStructure?: (index: number, structure: unknown) => Structure;
+    /** Number of maximum allowed structure viewers in [[ViewersGrid]] */
```
Luthaf, line 51:
```suggestion
    /** Maximum number of structure viewers allowed in [[ViewersGrid]] */
```

src/index.ts
```diff
@@ -103,6 +105,16 @@ function validateConfig(o: JsObject) {
     if ('loadStructure' in o && o.loadStructure !== undefined && !isFunction(o.loadStructure)) {
         throw Error('"loadStructure" should be a function in chemiscope config');
     }
+
+    if ('maxWidgets' in o) {
+        if (typeof o.maxWidgets !== 'number' || o.maxWidgets === null) {
```
Luthaf, line 110: why the null check here? `typeof` should already cover this, right? There is a null check when the expected type is `"object"`, since `typeof null === "object"`.

src/structure/grid.ts
```diff
@@ -727,6 +710,30 @@ export class ViewersGrid {
...
+                return { rows: 2, columns: 2 };
+            case 5:
+            case 6:
+                return { rows: 3, columns: 2 };
+            case 8:
+            case 7:
+            case 9:
+                return { rows: 3, columns: 3 };
+            default:
+                throw Error(`Can not create a grid with more than ${this._maxWidgets} elements`);
```
Luthaf, line 734: This should never be reached, so let's make it clear
```suggestion
                throw Error("reached unreachable code: too many viewer in the grid");
```

src/structure/grid.ts
```diff
@@ -157,7 +133,8 @@ export class ViewersGrid {
         element: string | HTMLElement,
         indexer: EnvironmentIndexer,
         structures: Structure[] | UserStructure[],
-        environments?: Environment[]
+        environments?: Environment[],
+        maxWidgets?: number
```
Luthaf, line 137: can you add documentation for this new parameter?

src/structure/grid.ts
```diff
@@ -183,6 +160,12 @@ export class ViewersGrid {
         this.oncreate = () => {};
         this.activeChanged = () => {};
 
+        if (maxWidgets !== undefined) {
+            this._maxWidgets = maxWidgets;
```
Luthaf, line 164: I would validate that maxWidget is lower than 9 here, since this class could be used directly, without going through the validation code in `index.ts`

src/structure/grid.ts
```diff
@@ -727,6 +710,30 @@ export class ViewersGrid {
 
         return newGUID;
     }
+
+    /**
+     * Function to return *optimal* arrangement of n widgets.
+     * Defaults to a 4 x ? grid. Hardcoded for simplicity.
```
Luthaf, line 716:
```suggestion
     * Function to return *optimal* arrangement of n widgets.
```

It is no longer using a 4x? grid, so let's update this

PR #184, opened 2021-07-26
Add meta descrition item in the app

Luthaf:
This will be used by google and other search engine when displaying the search result for chemiscope

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -6,7 +6,7 @@
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <meta
             name="description"
-            content="Interactive explorer for materials and molecular databases, allowing to correlate the atomic structures and their properties. Use it as an online app, as a fully-portable offline viewer, or as a collection of widgets that can be customized for specific applications."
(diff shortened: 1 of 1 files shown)


ceriottm (approved):

app/index.html
```diff
@@ -6,7 +6,7 @@
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <meta
             name="description"
-            content="Interactive explorer for materials and molecular databases, allowing to correlate the structures and properties of the systems"
+            content="Interactive explorer for materials and molecular databases, allowing to correlate the atomic structures and their properties. Use it as an online app, as a fully-portable offline viewer, or as a collection of widgets that can be customized for specific applications."
```
Luthaf, line 9: google will only show the first 155–160 characters, so the second sentence will not be shown

If that's something we want to advertise more, I would put it in the "about" modal instead

Luthaf: cf https://moz.com/learn/seo/meta-description

ceriottm (approved):

PR #196, opened 2021-09-14
Small quality of life improvements

Luthaf:
This leaves more space for the map itself, and make the map less "squished" on the side.

**Before**:
Before

**After**:
After

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -19,20 +19,6 @@ interface Configuration {
     loadStructure?: (index: number, structure: unknown) => Structure;
 }
 
-function cleanOutdatedMessages() {
-    const displayWarning = getByID('warning-display');
-    displayWarning.style.display = 'none';
-    displayWarning.getElementsByTagName('p')[0].innerText = '';
-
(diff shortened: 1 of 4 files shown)


ceriottm:
Makes sense. How does this look like when there are also symbols (e.g. the
NMR dataset)?

app/app.ts
```diff
@@ -131,7 +117,10 @@ export class ChemiscopeApp {
      */
 
     public async load(configuration: Configuration, dataset: Dataset): Promise<void> {
-        cleanOutdatedMessages();
+        // hide any error coming from the previous dataset loading
+        const errors = getByID('error-display');
+        errors.style.display = 'none';
+
```
rosecers, line 123: Will this leave the internals of the element intact? I.e. if you set the display to none and then reset the display, will the innertext be intact?

Luthaf: Yes the inner text stays the same, but will be overwritten before the error box is shown again with `style.display = "block"`

rosecers (approved):

PR #197, opened 2021-09-14
Create partial viewers to show data with only structures or only map

ceriottm:
Objective is to have these both as complement to DefaultVisualizer and as jupyter widgets (which is the real goal for me)


python/nbextension/src/widget.ts
```diff
@@ -79,3 +79,70 @@ export class ChemiscopeView extends DOMWidgetView {
...
+        // this function works by first rendering the widget inside `this.el`,
+        // and then inserting this.el inside the HTML document.
+        const element = this.el;
+
+        // HACK: resize Plotly when inserted. It currently render itself at full
+        // width and then needs to be resized. For more on this, see
+        // https://github.com/cosmo-epfl/chemiscope/pull/181#discussion_r693005307
+        element.addEventListener('DOMNodeInserted', () => {
+            window.dispatchEvent(new Event('resize'));
+        });
```
Luthaf, line 101: This is not needed for a pure structure viewer

python/chemiscope/jupyter.py
```diff
@@ -77,10 +90,22 @@ def show(frames, properties=None, meta=None, cutoff=None):
...
+        if properties is None:
+            properties = {}
+        if not "index" in properties: 
+            # also adds an index property to have something to show in the info panel
+            properties["index"] = list(range(len(frames)))
+                        
+        dict_input = create_input(
+            frames=frames, properties=properties, meta=meta, cutoff=cutoff
+        )
+        return StructureWidget(dict_input)
```
Luthaf, line 108:
```suggestion
        return StructureWidget(dict_input)
    else:
        raise ValueError(f"invalid mode '{mode}' in chemiscope.show, expected one of 'default' or 'structure'")
```

src/index.ts
```diff
@@ -349,7 +349,221 @@ class DefaultVisualizer {
...
+    /**
+     * Removes all the chemiscope widgets from the DOM
+     */
+    public remove(): void {
+        this.meta.remove();
+        this.info.remove();
+        this.structure.remove();
+    }
+
+    /**
```
Luthaf, line 509: I don't think we want to have the `saveSettings`, `applySettings`, or `dataset` for StructureVisualizer. They are purely duplicated code which no-one uses, and would be relatively easy to add back if we need them later.

ceriottm: I agree for the jupyter use case but how about the object that can be used in a webpage? Don't we want to give the possibility of applying settings? Alternative would be to move these to the chemiscope app and out of the viewers.

Luthaf: > how about the object that can be used in a webpage?

Since nobody is using this functionality, I would use YAGNI and remove code. 

Moving the settings functions to the app is also a good idea, I'll do it in a separate PR once this one is merged.

ceriottm:
Ready for review

jakublala (approved):

Luthaf:
Do we really want to have a metadata panel in these new modes? Since they are intended mainly for use in jupyter where the metadata is often not defined, I think that it does not make sense to display them (they only show a ⓘ symbol).

ceriottm:
I think actually that this shows we should not display the (i) when the
metadata is empty. Again, I can see uses of the "one panel chemiscope" in a
website, so to me it made sense to have a meta panel available.
Michele

Luthaf:
Rebased on top of #198. Should be good to merge after some more testing!

PR #202, opened 2021-11-01
Use a real orthographic camera now that 3DMol supports it fully

Luthaf:


ceriottm:
The orthographic view works beautifully, but for some reason the azaphenacene example remains stuck in a neverending loading loop
image

IDK how it can be related, doesn't show up in the deployed website

Luthaf:
Is this happening with a localhost version of chemiscope?

ceriottm:
> Is this happening with a localhost version of chemiscope?

yes

ceriottm:
actually this also happens on master, so it's unrelated.

Luthaf:
Ok, let's merge this and figure out the issue with azaphenacenes separately

PR #204, opened 2021-11-29
Bugfix create_input when passing full properties and no frames

Luthaf:


Luthaf:
There was a second issue where `_expand_properties` would change it's input parameter, meaning that calling `create_input` twice with the same parameter would make the code take two different path. This is now fixed as well.

I'll merge this and release 0.3.3 tomorrow if no one express concerns before that!

PR #206, opened 2021-12-03
Fix slider in map only visualizer

Luthaf:


Luthaf:
Without this fix, the slider in map only visualizer was not initialized to the right number of structures since Python creates an empty structure list in this case.

The alternative would be to generate a structure array with `{size: 0, data: 0}` structures on the Python side; but I fear this will introduce more issues down the line when trying to load such file directly in the "normal" chemiscope.

PR #207, opened 2021-12-07
Fix a bug when calling from an ipywidget callback

ceriottm:
Mini-bugfix for a weird problem that arise when embedding the chemiscope widget in such a way that it gets show()ed from an ipython callback.  WAT!?

See https://github.com/jupyter/jupyter/issues/299


Luthaf (approved):

ceriottm:
Can wait

PR #208, opened 2021-12-08
Simplified the heuristics for propery interpretation

ceriottm:
... and made the index property explicitly structure-based

### DIFF
--- a/python/tests/create_input.py
+++ b/python/tests/create_input.py
@@ -221,7 +221,7 @@ class TestCreateInputProperties(unittest.TestCase):
         self.assertEqual(
             str(cm.warning.args),
             "The target of the property 'name' is ambiguous because there is the same "
-            + "number of atoms and structures. Will assume target=structure. ",
+            + "number of atoms and structures. Will assume target=structure. "
         )
 
     def test_invalid_name(self):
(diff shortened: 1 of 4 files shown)


ceriottm:
@Luthaf I both solved the specific problem (index being created as an array rather than with explicit `target` specifications) and simplified the heuristics to deal with arrays. The only scenario in which we may have list-valued properties that have same size for  atoms and properties is when the dataset is 100% made of single-atom cells. I'd argue that making an executive decision makes more sense than having a separate test and error message thrown if natoms=nstructures, but let's discuss.

python/tests/create_input.py
```diff
@@ -216,13 +216,12 @@ def test_shortened_properties_errors(self):
 
         properties = {"name": ["2", "3"]}
         frames_single_atoms = [ase.Atoms("C"), ase.Atoms("H")]
-        with self.assertRaises(ValueError) as cm:
+        with self.assertWarns(UserWarning) as cm:
             create_input(frames=frames_single_atoms, properties=properties)
         self.assertEqual(
```
Luthaf, line 221:
```suggestion
            data = create_input(frames=frames_single_atoms, properties=properties)

        self.assertEqual(data["properties"]["name"]["target"], "structure")

        self.assertEqual(
```

Luthaf: Just to ensure that this behavior is preserved in the future

Luthaf (approved):

PR #214, opened 2022-01-19
Update to Bootstrap 5

### LINKED ISSUE #162: Bootstrap collapse buttons jump in Map Settings
The collapse buttons from Bootstrap used in 'more options' in map settings are not 'smooth' and jump when closing.

### LINKED ISSUE #194: Update to bootstrap 5 and use isolated bootstrap
We are currently using bootstrap 4, we should update to the latest release at some point. This will need some care to make sure everything looks right. 

We should also use an isolated bootstrap, using something similar to [this blog article](https://formden.com/blog/isolate-bootstrap). This will allow users to embed chemiscope inside other webpage using different version of bootstrap. This also means we should build the bootstrap SASS files to CSS in our own build system.

Both changes could be part of the same PR since both have the opportunity to break styling in a lot of ways, and making a single PR means we only need to manually check the result once.

slietar:
Some elements might appear different now for two reasons:
- The appearance of certain components changed between Bootstrap 4 and 5.
- Others relied on hacks that no longer work or features that no longer exist in Bootstrap 5, therefore I added a replacement solution that might look a little different.

See the [Bootstrap migration guide](https://getbootstrap.com/docs/5.0/migration/) for reference.
Partially fixes #194.

As I understand it, the docs use sphinx-bootstrap-theme which is blocked at Bootstrap 3, so there is no update possible for now.

---

Edit (@Luthaf):
fixes #162

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -20,11 +20,6 @@
             integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo="
             crossorigin="anonymous"
         ></script>
-        <script
-            src="https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js"
-            integrity="sha256-KM512VNnjElC30ehFwehXjx1YCHPiQkOPmqnrWtpccM="
-            crossorigin="anonymous"
-        ></script>
 
         <!-- font-awesome -->
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.11.2/css/all.min.css" />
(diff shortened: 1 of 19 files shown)


app/index.html
```diff
@@ -35,81 +35,79 @@
...
-                                <span class="dropdown-item clickable" onclick="CHEMISCOPE_APP.loadExample('Arginine-Dipeptide');">
-                                    Arginine-Dipeptide
-                                </span>
-                                <span class="dropdown-item clickable" onclick="CHEMISCOPE_APP.loadExample('CSD-1000R');"> Chemical Shieldings </span>
-                                <span class="dropdown-item clickable" onclick="CHEMISCOPE_APP.loadExample('Qm7b');"> Qm7b </span>
-                                <span class="dropdown-item clickable" onclick="CHEMISCOPE_APP.loadExample('Azaphenacenes');"> Azaphenacenes </span>
-                                <span class="dropdown-item clickable" onclick="CHEMISCOPE_APP.loadExample('Zeolites');"> Zeolites </span>
-                            </div>
-                        </li>
+                <div class="container-fluid">
```
Luthaf, line 38: should we make the div above a `container-fluid` instead of adding another one?

Luthaf (commented):
The changes looks good to me overall. One possible improvement would be to increase the size available for "scale" in the axis options for the map:

Capture d’écran 2022-01-19 à 14 53 21

It would be better to be able to read the full name without having to open the select, maybe by reducing the size taken by the min/max elements.

---
 
Another unrelated improvement would be to round the values for min/max (to something with ~6 digits), but this might be a bigger change since it is also linked to the scale of the graph itself, so we can leave it for later.

Luthaf:
CI failure looks weird, maybe something related to popper 2.0 not being compatible with node 12?

slietar:
Thanks for the feedback. I will enlarge the "scale" option as you suggested, I think that the arrow is bigger in Bootstrap 5 which is why it hides the text.

The CI fails because I forgot to update Popper to its version 2 as required by Bootstrap (the package name also changed). But I still don't understand why it didn't fail on my computer.

app/index.html
slietar: I don't think so, that's a breaking change described in [the migration guide](https://getbootstrap.com/docs/5.0/migration/#navbars) and [docs](https://getbootstrap.com/docs/5.0/components/navbar/). I think I should move the `container` class to the `header` element though.

Luthaf:
I found another small issue with the info panel. If you open the chemical shielding example, then click on `atom xxx` to open the table with all properties, and then on `structure xxx`, opening the structure table does not close the atom table like it use to (making the structure table hidden behind the atom table).

slietar:
That last issue should be fixed. I also updated the docs as only 3dmol relies on jQuery now.

Luthaf (approved):
Looks good to me! I've tweaked a couple of settings for the map. If you don't disagree witht the changes, I'll merge this on Monday!

slietar:
Ok, your changes also look good to me. I have a good draft for the map isolation, I will open a PR for that as soon as you merge this one.

PR #217, opened 2022-01-25
Actually change the cutoff in HTML options when loading a new environement

Luthaf:
Previous code was broken for a while, but the reasonable default cutoff hide the issue.

Thanks @agoscinski for finding & reporting this!

### DIFF
--- a/src/structure/widget.ts
+++ b/src/structure/widget.ts
@@ -723,7 +723,8 @@ export class MoleculeViewer {
         // Setup various buttons
         this._resetEnvCutoff = getByID<HTMLButtonElement>(`${this.guid}-env-reset`);
         this._resetEnvCutoff.onclick = () => {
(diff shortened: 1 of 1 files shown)


Luthaf:
@slietar do you want to try and review this?

slietar:
I finally got a chance to look at your changes. I'm not entirely familiar with this part of the code, but it looks good to me!
I was just wondering why is the cutoff always at 5 Å by default?

Luthaf:
The default for the cutoff is 4A, set here: https://github.com/lab-cosmo/chemiscope/blob/83f9929b2d226383fcf57bd1173546033e50fba5/src/structure/options.ts#L86

Where do you see a 5A value? 

---

This default value should never be seen by the user since we only use cutoff when the user gives a list of environments, and these contains their own cutoff. 4A was picked as a default to initialize the data structure because it is a reasonable value, nothing else 😄

PR #218, opened 2022-01-30
Interface to pass settings to chemiscope widgets

ceriottm:
A minimalistic approach to defining settings of the chemiscope widget in a jupyter notebook. 
Basically one just passes the JSON string corresponding to the stored settings. 
Required cleaning up a bit here and there.

PR #219, opened 2022-01-31
Make the settings a proper part of the dataset

Luthaf:
This PR allows specifying the settings in `chemiscope.create_input` and friends, and to specify these settings in a jupyter notebook

### DIFF
--- a/python/chemiscope/jupyter.py
+++ b/python/chemiscope/jupyter.py
@@ -2,29 +2,32 @@
 import json
 import gzip
 import ipywidgets
-from traitlets import Unicode, Bool
+from traitlets import Unicode, Bool, Dict
 
 from .input import create_input
 
 
-@ipywidgets.register
-class ChemiscopeWidget(ipywidgets.DOMWidget, ipywidgets.ValueWidget):
-    _view_name = Unicode("ChemiscopeView").tag(sync=True)
-    _view_module = Unicode("chemiscope-widget").tag(sync=True)
-
+class ChemiscopeWidgetBase(ipywidgets.DOMWidget, ipywidgets.ValueWidget):
     data = Unicode().tag(sync=True)
     has_metadata = Bool().tag(sync=True)
+
+    # synchronized settings from the JS side. You can assign to this field to
+    # change what's being displayed by chemiscope, but you need to assign a full
+    # dictionary (`widget.settings["map"]["x"]["property"] = "foo"` will not
+    # work, but `widget.settings = updated_settings` will).
+    settings = Dict().tag(sync=True)
 
     def __init__(self, data, has_metadata):
         super().__init__()
         self.data = json.dumps(data)
         self.has_metadata = has_metadata
+        self.settings = {}
 
     def save(self, path):
         """
-        Save the dataset displayed by this :py:class:`ChemiscopeWidget` as JSON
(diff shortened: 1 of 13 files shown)


Luthaf:
The main difference with #218 is that the settings are passed as a dictionary instead of a JSON string. The dictionary follows the documentation in https://chemiscope.org/docs/tutorial/input-reference.html, `settings` section.


Here are a couple of examples:
```py
# show unit cell for the initial viewer
chemiscope.show(frames=frames, properties=properties, settings={"structure": [{"unitCell": True}]})

# change the default property for x/y axis
chemiscope.show(frames=frames, properties=properties, settings={"map": {"x": {"property": "<something>"}}})
```

ceriottm:
is this compatible with calling `JSON.parse(thefullstring)` ?

Luthaf:
Yes, this is the same representation

python/chemiscope/jupyter.py
```diff
@@ -2,29 +2,32 @@
...
         self.data = json.dumps(data)
         self.has_metadata = has_metadata
+        self.settings = {}
 
     def save(self, path):
         """
-        Save the dataset displayed by this :py:class:`ChemiscopeWidget` as JSON
-        to the given ``path``. If ``path`` ends with ``.gz``, the file is
-        written as gzip compressed JSON string.
+        Save the dataset displayed by this widget as JSON to the given ``path``.
```
ceriottm, line 28: Shall we stress that this will be the same JSON created when showing the widget, and won't be updated in terms of style?

ceriottm: Also, shall we add a `save_settings` to save the instantaneous settings similar to what can be done with the webpage version?

python/nbextension/src/widget.ts
```diff
@@ -7,7 +7,7 @@ import './chemiscope-bootstrap.less';
 import 'bootstrap/dist/js/bootstrap.min.js';
 
 import { DefaultVisualizer, MapVisualizer, StructureVisualizer } from '../../../src/index';
-import { Dataset } from '../../../src/dataset';
+import { Dataset, Settings } from '../../../src/dataset';
 
 /**
  * The [[ChemiscopeView]] class renders the Chemiscope App as a widget in the
```
ceriottm, line 13: Is there a way to create a base class and reduce a bit the repetition? Only if it is a quick fix though

ceriottm (approved):
Just a couple of optional todos. I let you decide whether to do them, and then merge as you see fit.

python/chemiscope/jupyter.py
Luthaf: > Shall we stress that this will be the same JSON created when showing the widget, and won't be updated in terms of style?

This is no longer the case: https://github.com/lab-cosmo/chemiscope/pull/219/commits/a211d42301bef340218acdc13f2930f6625e7fa8#diff-71fe19e6aa531b95aac2bd2f498dc21dee83eb341831b70c40c887d8bb904d73R39-R43

python/nbextension/src/widget.ts
Luthaf: yes, I though about this as well. It would involve some form of generics though, since the visualizer has a different type every time. I think this should be alright for now, there is not too much repetition.

PR #220, opened 2022-02-08
Isolate the map component

slietar:
Main changes:

- The whole map component is in a shadow root, with the options modal as a child of the root.
- Bootstrap and Chemiscope's CSS are loaded using constructed style sheets, loaded using a Webpack resource query `?sheet` and [polyfilled](https://github.com/calebdwilliams/construct-style-sheets) on browsers that don't support it (i.e. all except Blink browsers).
- Other CSS files are loaded normally using ES6 imports.
- Plotly's `<style>` elements are removed from the `<head>` and transformed into a constructed style sheet adopted by the corresponding shadow root.
- Bootstrap's collapse and modal logic is rewritten, but still uses Bootstrap's classes (e.g. .modal) and attributes (e.g. data-bs-target).

Questions/tasks that remain:

- Should the polyfill be added on the package or the app? I think in the package is ok since it's quite small (5 kB), perfectly replicates the spec and wouldn't cause problems if included twice.
- Adding `dom.iterable` to `compilerOptions.lib` in the typescript config would avoid useless `Array.from()` calls in `for (let el of Array.from(document.querySelectorAll(...))) { ... }`. Is there any argument against this?
- Move Bootstrap replacement files in a specific directory, something like `components`? I think the name components would be better for referring to map and structure, so feel free to propose something else.
- It would be nice for the user to be able to create a map not only by providing an element but also by receiving one. This would allow embedding in tools such as [Observable](https://observablehq.com/) where the user can return an element at the end of a code block to be displayed. For example we could give the shadow root's host as `map.element` instead of asking for its parent in the map's constructor.
- I have opened an issue regarding Plotly's bug: plotly/plotly.js#6108. In the meantime I have created a hack that intercepts mousedown and mouseup events to fix the issue ~(not in the PR yet)~.
- ~I need to update the tests as well.~

Let me know what you think.

### DIFF
--- a/src/collapse.ts
+++ b/src/collapse.ts
@@ -90,8 +90,8 @@ export default class Collapse {
      */
     static initialize(root: HTMLElement = document.body) {
         // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-        for (const el of Array.from(root.querySelectorAll('[data-bs-toggle]')) as HTMLElement[]) {
-            if ('bsTarget' in el.dataset) {
+        for (const el of root.querySelectorAll('[data-bs-toggle]')) {
+            if (el instanceof HTMLElement && 'bsTarget' in el.dataset) {
                 const target: HTMLElement | null = root.querySelector(
                     (el.dataset as { bsTarget: string }).bsTarget
                 );
--- a/src/collapse.ts
+++ b/src/collapse.ts
@@ -1,0 +1,126 @@
+import assert from 'assert';
+
+/**
+ * Component for creating collapsible elements.
+ *
+ * Manipulated classes and styles are summarized here:
+ *  - Hidden:         .collapse
+ *  - Shown:          .collapse.show
+ *  - Hiding/showing: .collapsing, width/height
+ * The .collapse class should be present when calling the Collapse constructor.
+ * See [the Bootstrap docs](https://getbootstrap.com/docs/5.2/components/collapse/)
+ * for details.
+ */
+export default class Collapse {
+    private _horizontal: boolean;
+    private _property: 'width' | 'height';
+    private _visible: boolean = false;
+
+    /**
+     * Create a Collapse instance.
+     * @param _element The element to be made collapsible. It should have the
+     *  .collapse class.
+     */
+    constructor(private _element: HTMLElement) {
+        this._horizontal = this._element.classList.contains('horizontal');
+        this._property = this._horizontal ? 'width' : 'height';
+
+        this._element.addEventListener('transitionend', (event) => {
+            if (event.target === event.currentTarget && event.propertyName === this._property) {
+                this._element.classList.replace('collapsing', 'collapse');
+
+                if (this._visible) {
+                    this._element.classList.add('show');
+                    this._element.style.removeProperty(this._property);
+                }
+            }
+        });
+    }
+
+    hide() {
+        const size = this._element.getBoundingClientRect()[this._property];
+        this._element.style.setProperty(this._property, `${size}px`);
+
+        this._element.classList.replace('collapse', 'collapsing');
+        this._element.classList.remove('show');
+
+        // eslint-disable-next-line no-unused-expressions
+        this._element.offsetHeight;
+        this._element.style.removeProperty(this._property);
+
+        this._visible = false;
+    }
+
+    show() {
+        const scrollProperty = this._horizontal ? 'scrollWidth' : 'scrollHeight';
+
+        this._element.classList.replace('collapse', 'collapsing');
+        this._element.style.setProperty(this._property, '0');
+
+        const size = this._element[scrollProperty];
+        this._element.style.setProperty(this._property, `${size}px`);
+
+        this._visible = true;
+    }
+
+    toggle(value: boolean = !this._visible) {
+        if (value) {
+            this.show();
+        } else {
+            this.hide();
+        }
+    }
+
+    /**
+     * Set an element as a trigger for the collapsible element.
+     * @param el The element that should be set as a trigger.
+     */
+    addTrigger(el: HTMLElement) {
+        el.addEventListener('click', (event) => {
+            event.preventDefault();
+            this.toggle();
+        });
+    }
+
+    /**
+     * Scan the DOM for elements with the [data-bs-toggle] attribute to make
+     * them interactive with regards to this component.
+     *
+     * @param root The root element in which to scan.
+     */
+    static initialize(root: HTMLElement = document.body) {
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+        for (const el of Array.from(root.querySelectorAll('[data-bs-toggle]')) as HTMLElement[]) {
+            if ('bsTarget' in el.dataset) {
+                const target: HTMLElement | null = root.querySelector(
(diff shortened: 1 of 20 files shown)


slietar:
The map component should now look good on both Jupyter Notebook and Jupyter Lab. A few changes I made:

- Jupyter Notebook has a `font-size` of 11px on `<html>` which causes all text to be very small as Bootstrap relies on rems. I overrode this rule and it didn't cause problems with Jupyter itself but theoretically could be unexpected to some other plugins.
- The updated Webpack configuration is not comatible with the Jupyter build process as it is not possible to override the rule for CSS files. Two possible solutions:
  - Use inline imports specifying the configuration, e.g. `!css-loader?exportType=css-style-sheet!bootstrap/dist/css/bootstrap.min.css';`. That's what I did for now.
  - Add another Webpack build process upstream.
- I moved modals to a separate shadow root in the body to fix z-index issues.

Luthaf:
> The map component should now look good on both Jupyter Notebook and Jupyter Lab.

Amazing news! 

> The updated Webpack configuration is not compatible with the Jupyter build process as it is not possible to override the rule for CSS files.

We already use somehow separate webpack configuration for the jupyter build (https://github.com/lab-cosmo/chemiscope/blob/5d73aff6ed78a2fa3796dfcec88683edb31a481e/python/webpack.config.labextension.ts and https://github.com/lab-cosmo/chemiscope/blob/5d73aff6ed78a2fa3796dfcec88683edb31a481e/python/webpack.config.nbextension.ts in the jupyterlab support branch), so I think we can use custom webpack loader for css there as well (potentially renaming the file from `.css` to `.css.in` or something so that the default rules don't match). I already do this for `.html` files in these custom config.

slietar:
Indeed, I encountered an error with the default rules. Renaming input files would work, but how would you deal with CSS files that belong to dependencies such as Bootstrap, as in `bootstrap/dist/css/bootstrap.min.css`? Copy them as `.css.in` beforehand?

Another solution would be to import (using inline loader options) and re-export all CSS in a single file, something like that:

```js
import bootstrap from '!css-loader?exportType=css-style-sheet!bootstrap/dist/css/bootstrap.min.css';

export { bootstrap };
```

This way we keep all the weird imports in the same place.

Luthaf:
Yes, inline loader might be the best solution here, with a nice comment to not have people too confused when looking at it.

slietar:
The PR should be ready to be merged, although I'm not sure what that failed Python test is about.

Luthaf:
I just fixed the tox4 issue in #232, could you rebase this PR on top of master to make sure tests are fixed in this PR as well?

slietar:
It should be okay now!

src/collapse.ts
```diff
@@ -0,0 +1,95 @@
+import assert from 'assert';
+
+export default class Collapse {
```
Luthaf, line 3: This class needs some documentation: what is it doing, & the fact that it was adapted (copied?) from bootstrap

src/collapse.ts
```diff
@@ -0,0 +1,95 @@
...
+
+                if (target) {
+                    getCollapse(target).addTrigger(el);
+                }
+            }
+        }
+    }
+}
+
+const map = new WeakMap<HTMLElement, Collapse>();
```
Luthaf, line 81: could you explain why a weak map is required here? and what the function below does/why it is required?

src/map/options.ts
```diff
@@ -154,12 +157,15 @@ export class MapOptions extends OptionsGroup {
...
         root.appendChild(this._openModal);
-        document.body.appendChild(this._modal);
 
         this._bind(properties);
         this.applySettings(settings);
     }
 
+    public getById<T extends HTMLElement = HTMLElement>(id: string): T {
+        return this._modal.shadow.getElementById(id) as T;
+    }
```
Luthaf, line 167: The previous implementation of `getById` also checked if the returned element is not `null`, which caught a few bugs in, the past. 

It also takes the root in which to search the element as a second argument, so you could implement this function using `getById(this.getId(id), this._modal.shadow)` where needed.

Since the function is only looking inside the modal, how about calling it `getModalElement` or something similar? And we can also merge this function with `getId`.

src/map/plotly/fix-plot.ts
```diff
@@ -0,0 +1,107 @@
+import { PlotlyScatterElement } from './plotly-scatter';
+
+export default function fixPlot(plot: PlotlyScatterElement) {
```
Luthaf, line 3: Please add a comment explaining what is the issue and why this function is required, ideally linking to the plotly issue

src/modal.ts
```diff
@@ -0,0 +1,113 @@
+export default class Modal {
```
Luthaf, line 1: Same as Collapse, this should have a short explanation of the code, what it is doing and where it comes from

src/structure/widget.ts
```diff
@@ -239,9 +239,9 @@ export class MoleculeViewer {
         this._root.addEventListener(
             'wheel',
             (event) => {
-                // Avoid an infinite loop by only intercepting the original event
-                // and not synthetic ones.
-                if (event instanceof WheelEvent) {
+                // Avoid an infinite loop by only intercepting the original
+                // (= trusted) event and not synthetic ones.
+                if (event.isTrusted) {
```
Luthaf, line 244: nice!

Luthaf (commented):
Looks really good overall! I'd like to see some more documentation on some tricky parts of the code before merging

src/map/options.ts
slietar: Yes I agree it's better to use the global `getByID` method which I actually had updated to work with shadow roots (although I'm a little skeptical to the idea of using the second argument when not looking inside a shadow root, given that ids are unique).

I also ended up removing all guids in this file so there is no more need for `getId()`.

slietar:
Thanks for the feedback, I have added comments as you suggested. Let me know if you'd like to see other changes.

src/collapse.ts
slietar: The point of `getCollapse()` is to access existing `Collapse` instances without the reference to that instance but instead one to the corresponding element.

```js
// Initially
let collapse = getCollapse(element);

// Later (same collapse)
let collapse = getCollapse(element);

// Removing the element
element.remove(); // -> `collapse` will eventually be garbage-collected as the WeakMap does not count as a reference
```

Setting `element.collapse = new Collapse()` works fine, but pollutes the element object (using symbols would be an alternative though).

slietar: Also `getFromMap()` underneath is just a way of generalizing this solution to any key and value given a `WeakMap` and a function to generate new values when missing from the map.

src/map/options.ts
Luthaf: > although I'm a little skeptical to the idea of using the second argument when not looking inside a shadow root, given that ids are unique

The use case for this is that the element might not be inserted in the document yet, which is the case when creating a new jupyter widget

src/collapse.ts
Luthaf: ok, that sounds good!

Luthaf (approved):
This is good to go!

src/map/options.ts
slietar: Ok this makes sense indeed!

PR #232, opened 2022-05-04
Simple fix to allow for non-bonding species in 3dmol

rosecers:
Allows users to designate non-bonding items as type 'X', which is typically reserved for generalized entities in ase without known bonding nature (so it's appropriate)

### DIFF
--- a/src/structure/3dmol/assignBonds.ts
+++ b/src/structure/3dmol/assignBonds.ts
@@ -87,6 +87,7 @@ function bondLength(elem: string): number {
 // return true if atom1 and atom2 are probably bonded to each other
 // based on distance alone
 function areConnected(atom1: AtomSpec, atom2: AtomSpec) {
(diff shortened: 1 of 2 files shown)


src/structure/3dmol/assignBonds.ts
```diff
@@ -87,6 +87,7 @@ function bondLength(elem: string): number {
 // return true if atom1 and atom2 are probably bonded to each other
 // based on distance alone
 function areConnected(atom1: AtomSpec, atom2: AtomSpec) {
+    if (atom1.elem === "X" || atom2.elem === "X") return false;
```
Luthaf, line 90:
```suggestion
    if (atom1.elem === "X" || atom2.elem === "X") {
        return false;
    }
```

should fix the prettier issue

Luthaf (approved):

PR #233, opened 2022-05-18
Fixed an infinite-loop condition due to the auto-updating settings

ceriottm:
* Added a flag to disable auto-update
* Use the flag to avoid trigger an update when changes happen on the JS side
* Also streamlined a bit the widgets using a base class

### DIFF
--- a/python/nbextension/src/widget.ts
+++ b/python/nbextension/src/widget.ts
@@ -9,15 +9,69 @@ import 'bootstrap/dist/js/bootstrap.min.js';
 import { DefaultVisualizer, MapVisualizer, StructureVisualizer } from '../../../src/index';
 import { Dataset, Settings } from '../../../src/dataset';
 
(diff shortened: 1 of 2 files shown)


python/nbextension/src/widget.ts
```diff
@@ -9,15 +9,69 @@ import 'bootstrap/dist/js/bootstrap.min.js';
 import { DefaultVisualizer, MapVisualizer, StructureVisualizer } from '../../../src/index';
 import { Dataset, Settings } from '../../../src/dataset';
 
+
+class CSBaseView extends DOMWidgetView {
```
Luthaf, line 13:
```suggestion
class ChemiscopeBaseView extends DOMWidgetView {
```

removes the need to guess what `CS` means here

Luthaf (approved):

PR #236, opened 2022-06-13
Isolate the structure component

slietar:
Not much new here, the isolation logic is similar to that of #220. It should work correctly on Jupyter Notebook and Lab.


Luthaf (approved):
This looks good and works fine!

PR #238, opened 2022-06-21
Isolate the metadata panel

Luthaf:
This PR contains two commits: the first one removes the guid from the structure viewer (as far as I understand, it is no longer required since everything moved to a shadow root, and different roots can have the same id inside); and the second moves the metadata panel to a shadow root.

@slietar would you have some time to review this?

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -47,7 +47,7 @@ export class ChemiscopeApp {
             <div class="row">
                 <div class="col-md-6" style="padding: 0">
                     <div class="ratio ratio-1x1">
-                        <div id="chemiscope-meta"></div>
+                        <div id="chemiscope-meta" style="z-index: 10"></div>
                         <div id="chemiscope-map" style="position: absolute"></div>
                     </div>
                 </div>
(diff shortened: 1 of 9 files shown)


slietar:
Sure, I'll review this as soon as I can!

slietar:
The metadata panel works well except for removal (e.g. by selecting another example in the application). I get this error: `NotSupportedError: Failed to execute 'attachShadow' on 'Element': Shadow root cannot be created on a host which already hosts a shadow tree.`, caused by the fact that you're only removing the shadow root's contents and not the shadow root itself.

Since shadow roots cannot be removed, I think you need to add one more `<div>` element as a parent of the shadow root (I called it `hostElement`, [see here](https://github.com/lab-cosmo/chemiscope/blob/421835df8d9ce1196fb3c4aebce007d5bac5afd3/src/map/map.ts#L240-L247)) and remove that element. There might be a more straightforward way of doing it though.

slietar:
It looks good to me this time! I have the PR for the touchscreen fix and the one for the info isolation ready, I'll create them as soon as you merge this one.

PR #240, opened 2022-07-28
Fix map click offset on touch screens

slietar:


Luthaf (approved):
That works great, thanks a lot!

PR #241, opened 2022-08-18
Isolate the info component

slietar:
This PR isolates the last component of Chemiscope which is the info panel. The `Collapse` class is also updated to support the `data-bs-parent` attribute of [Bootstrap](https://getbootstrap.com/docs/5.2/components/collapse).


Luthaf:
Sorry, this fell off my radar for some reason. Thanks a lot!

slietar:
Thanks!

PR #247, opened 2022-09-03
Fixed and tweaked the styling of widgets in a jupyter notebook

ceriottm:
There were several quirks with the widgets - invisible scrollbar, invisible meta label, ...
This should fix them.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -24,6 +24,7 @@ import {
     WarningHandler,
     addWarningHandler,
     getNextColor,
+    sendWarning,
 } from './utils';
 
 import './static/chemiscope.css';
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -613,7 +613,16 @@ class MapVisualizer {
...
-        const n_structure = dataset.properties[Object.keys(dataset.properties)[0]].values.length;
+
+        // unless all properties have atom targets, we assume the map is meant to represent structures
+        let n_structure = 0;
+        for (const key of Object.keys(dataset.properties)) {
+            n_structure = dataset.properties[key].values.length;
+            if (dataset.properties[key].target === 'structure') {
+                break;
+            }
+        }
```
Luthaf, line 624: Why is this? The default for the other viewer is to use per-atom properties if environments are present, otherwise use per-structure properties.

ceriottm: I had this problem where a dataset of structures that had stray environment properties in it was completely broken when showing a map-only viewer. When we do a combined view it can use info on whether environments are in, here it relies only on counting the number of entries in the first property. For sure it was a bug, whether this is the right way to fix it, I'm open to alternative strategies.

Luthaf: Do you have the notebook around? It might be possible to fix this by detecting the presence/absence of the `environments` parameter instead

PR #248, opened 2022-09-12
Stop automatically extracting properties from frames

Luthaf:
Instead, we provide functions to decide what properties one wants to extract

Code like this

```py
chemiscope.input(
    frames=my_frames,
    properties=my_properties,
   ...
)

# only frame properties
chemiscope.input(
    frames=my_frames,
   ...
)
```

Should now be like this

```py
frame_properties = chemiscope.extract_properties(my_frames)
my_properties.update(frame_properties)

chemiscope.input(
    frames=my_frames,
    properties=my_properties,
   ...
)

# only frame properties
chemiscope.input(
    frames=my_frames,
    properties=chemiscope.extract_properties(my_frames),
   ...
)
```

----

I think this makes the API less surprising/less magical, and make it easier for users to select a subset of properties out of the frames.

It also helps on the developer side by reducing the explosion of parameters, for example `composition` is now a separate function that users can decide to use or not. Similar function could compute a basic SOAP representation and project it in 2 dimensions for nicer integration with lib-rascal-ine, etc.

However, this is breaking enough to warrant a bit of discussion and releasing it in a new version (0.5) altogether.


python/chemiscope/input.py
```diff
@@ -245,46 +159,38 @@ def create_input(
...
+                if len(properties_list) != 0:
+                    properties_list += " and "
+                properties_list += "[" + ", ".join(atom_properties) + "]"
+
+            if len(properties_list) != 0:
+                warnings.warn(
+                    "chemiscope behavior changed to no longer include properties "
+                    "from the structure objects. Use `chemiscope.extract_properties` "
+                    f"to also visualize these properties ({properties_list})"
+                )
```
Luthaf, line 193: this warning will inform users of the change in behavior once they update to the new version. Hopefully it is clear enough!

ceriottm: I am against having error messages that refer to previous behavior. Require keeping track to remove them at the next version bump, which I think we don't have the scale to do. I suggest that we don't show a message if properties are provided explicitly, and if someone only passes frames and asks for a map-containing visualization we could write something like "Structure or atom properties have to be provided using the `properties` keyword. If you are using an atoms frames that also contain properties, you can use `chemiscope.extract_properties` to retrieve them in a format suitable to be used in this context."

ceriottm:
Also, I suggest we aim to have jupyter-lab integration for 0.5 ^_^

Luthaf:
> Also, I suggest we aim to have jupyter-lab integration for 0.5 ^_^

Yes, that's my plan: this PR and jupyter-lab will go into 0.5

python/chemiscope/input.py
ceriottm: Yeah I'm worried about too many warnings (that BTW don't disappear automatically in jupyter as far as I could tell). 
I don't have very strong feelings overall, I'd happily wait for you to converge on something that makes sense to you and then test it to give feedback.

PR #263, opened 2022-12-13
Updated plausible tracker code to only report lab/notebook

ceriottm:
Previously this was sending the name and path of the notebook, which seems like too personal information

### DIFF
--- a/python/jupyter/src/widget.ts
+++ b/python/jupyter/src/widget.ts
@@ -9,8 +9,8 @@ import './widget.css';
 import { DefaultVisualizer, MapVisualizer, StructureVisualizer } from '../../../src/index';
 import { Dataset, Settings } from '../../../src/dataset';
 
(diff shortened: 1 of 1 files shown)


python/jupyter/src/widget.ts
```diff
@@ -83,7 +83,7 @@ export class ChemiscopeView extends ChemiscopeBaseView {
 
     public render(): void {
         PlausibleTracker.trackEvent('jupyter-show', undefined, {
```
Luthaf, line 85: We might want to use [`trackPageView`](https://plausible-tracker.netlify.app/globals#trackpageview) instead of `trackEvent`

ceriottm: uhm. I just modified what you had there - IDK what actually changes, the result is good enough.
you mean that given that we changed to a fake subdomain now we can ignore the "conversion target" thing and directly record pages?

PR #264, opened 2022-12-31
Changed structure viewer to use playback delay as a savable option

ceriottm:
It's nice to be able to set (and save) the playback delay, which was impossible because it was not part of the structure options. This adds a playbackDelay option that is saved in the dataset json, and is accessible from Python as well, and removes some (IMO) unnecessary special handling of playbackDelay inside ViewersGrid. 

Also, moves from using tens of second to milliseconds as units, which seems more natural.


Luthaf (approved):
looks good!

PR #266, opened 2023-01-18
Improve error message when the user forgets to specify environments

### LINKED ISSUE #216: Wrong error message when cutoff is missing for plotting atomic properties
When the cutoff is missing in
```python
a = 5
GaAs_cubic = ase.Atoms("8Ga", positions=[[0,0,0], [a,0,0], [0,a,0], [0,0,a], [a,a,0], [a,0,a], [0,a,a], [a,a,a]], cell=[a,a,a])
properties = {}
properties["name1"] = {
                "target": "atom",
                "values": list(range(len(GaAs_cubic))),
                "units": "foo",
                "description": "foo"
            }

properties["name2"] = {
                "target": "atom",
                "values": list(range(len(GaAs_cubic))),
                "units": "foo",
                "description": "foo"
            }

cs = chemiscope.show([GaAs_cubic], properties=properties, mode="default")
display(cs)
```
The error message is 

> Error: we need at least two properties to plot in the map, we have none

While it should be something like

> Error: Cutoff must be given

or even set to a default to no cutoff, if this is possible.

Error message comes from 
https://github.com/lab-cosmo/chemiscope/blob/83b85b3ada8708508c5d93821da6c231b1ccf43b/src/map/options.ts#L95
Checks are done here
https://github.com/lab-cosmo/chemiscope/blob/83b85b3ada8708508c5d93821da6c231b1ccf43b/src/dataset.ts#L294

Luthaf:
Fixes #216

### DIFF
--- a/src/indexer.ts
+++ b/src/indexer.ts
@@ -178,14 +178,9 @@ export class EnvironmentIndexer {
         }
     }
 
-    /** Get the total number of environments currently being displayed */
-    public environmentsCount(): number {
-        if (this.mode === 'atom') {
-            assert(this._environments !== undefined);
-            return this._environments.length;
-        } else {
-            return this._structures.length;
-        }
+    /** Does this Indexer knows about atom-centered environments? */
+    public hasEnvironments(): boolean {
+        return this._environments !== undefined;
     }
 
     /** Get the total number of structures we know about */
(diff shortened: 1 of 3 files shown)


agoscinski:
Verified new error message is now
>Error: could not find enough structure properties to display, but there are atom properties. Please provide the environment list to display them

providing an environment list is enforcing to provide a cutoff

Luthaf:
Thanks for checking this!

PR #273, opened 2023-03-20
Fix CI job checking google collab

Luthaf:


Luthaf:
The notebook was edited, @rosecers did you change it? It would be nice to check that CI still works here next time you need to do this.

rosecers:
> The notebook was edited, @rosecers did you change it? It would be nice to check that CI still works here next time you need to do this.

Hey, sorry about that -- just making a skcosmo -> skmatter change. Do you have any thoughts on a better live-notebook than colab? Would love to change the interfacing if possible.

Luthaf:
I don't know, my ideal solution would have the notebook in this repo, so we are sure it always passes CI. There should be solutions to run live notebooks from inside a git repo

PR #278, opened 2023-04-24
Fix for log/inverse/sqrt color scale in map plot

angerossi:
- Fixed broken interaction with color axis limits
- Added warnings for user selecting properties that have **some** values <= 0
- Ensured that the user cannot select a property with **only** values <= 0 with the log/sqrt scale.

### DIFF
--- a/docs/src/tutorial/input-reference.rst
+++ b/docs/src/tutorial/input-reference.rst
@@ -36,15 +36,42 @@ contains the following fields and values:
                 // the property target: is it defined per atom or for the full
                 // structures
                 "target": "atom" | "structure",
-                // values of the properties can either be numbers or strings.
+                // values of the properties can either be numbers, strings
+                // or array of numbers.
+                //
                 // string properties are assumed to represent categories of
                 // data.
-                "values": [1, 2, 3, ...] | ["first", "second", "first", ...],
+                //
+                // the first dimension of the multidimensional property corresponds
+                // to the number atoms/structures, the second dimension corresponds
+                // to the size of the array of values per atom/structure.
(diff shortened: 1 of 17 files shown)


ceriottm:
OK, I'm testing this. Two first observations:
1. it looks like proper logarithmic colorscale labels _can_ be generated see 
https://plotly.com/python/colorscales/#customizing-tick-text-on-logarithmic-color-bars
2. the error message says that negative values are mapped to gray, but actually they are shown with the color corresponding to the minimum of the scale

Luthaf:
> it looks like proper logarithmic colorscale labels can be generated see
https://plotly.com/python/colorscales/#customizing-tick-text-on-logarithmic-color-bars

This is possible, but mainly works if the data spans multiple orders of magnitude. If all values are between 1 and 3, creating log-spaced ticks is a bit harder.

angerossi:
> it looks like proper logarithmic colorscale labels can be generated see
https://plotly.com/python/colorscales/#customizing-tick-text-on-logarithmic-color-bars

I did have a look at this. While it is possible, the [`ticktest`](https://plotly.com/python/reference/layout/coloraxis/#layout-coloraxis-colorbar-ticktext) argument requires that the [`tickmode`](https://plotly.com/python/reference/layout/coloraxis/#layout-coloraxis-colorbar-tickmode) is set to `array`, such that the `tickvals` are set in the layout.  Also, the alternative implementation with the [`labelalias`](https://plotly.com/python/reference/layout/coloraxis/#layout-coloraxis-colorbar-labelalias) argument allows for any `tickmode`, but the `tickvals` are still required to set the aliases. Since we set the min and max for the colorbar, Plotly automatically sets the tickvalues and extracting them is non-trivial.

PR #281, opened 2023-05-01
Considering the pyodide_kernel interpreter in the _is_running_in_notebook

agoscinski:
jupyterlite is using pyodide_kernel interpreter to run python code in the browser, we add this so chemiscope can be  run within jupyterlite

From jupyterlite
image
try out https://agoscinski.github.io/jupyterlite-playground/lab/index.html (adding example currently, might need 20 minutes till it is on the homepage) EDIT: okay works now, use the cosmo.ipynb

I added this change to the code and it worked out for me

### DIFF
--- a/python/chemiscope/jupyter.py
+++ b/python/chemiscope/jupyter.py
@@ -194,7 +194,9 @@ def _is_running_in_notebook():
 
     try:
         shell = get_ipython().__class__.__name__
-        if shell == "ZMQInteractiveShell":
+        # ZMQInteractiveShell is the standard Jupyter Kernel
+        # Interpreter is used by pyiodide
+        if shell in ["ZMQInteractiveShell", "Interpreter"]:
             return True
         elif shell == "TerminalInteractiveShell":
             return False


python/chemiscope/jupyter.py
```diff
@@ -194,7 +194,7 @@ def _is_running_in_notebook():
 
     try:
         shell = get_ipython().__class__.__name__
-        if shell == "ZMQInteractiveShell":
+        if shell == "ZMQInteractiveShell" or "Interpreter":
```
Luthaf, line 197: that does not do what you think it does ^_^. This code evaluate "Interpreter" to a bool (`True`), and then evaluate the `or`.

```suggestion
        # ZMQInteractiveShell is the standard Jupyter Kernel
        # Interpreter is used by pyiodide
        if shell in ["ZMQInteractiveShell", "Interpreter"]:
```

Luthaf:
This looks good, although I could not understand where the javascript part of the extension is loaded from. Removing most of the code in your example notebook, I just patched this function with `chemiscope.jupyter._is_running_in_notebook = lambda : True`, and it worked, but I'm not sure if it loads the JS from github-pages, through `piplite` (and some kind of in-memory FS) or from npm. Do you know how this work?

agoscinski:
I think its on the github server of the page, because if I don't install chemiscope on the requirements.txt
https://github.com/agoscinski/jupyterlite-playground/actions/runs/4849978532/jobs/8642471828 (see Install dependencies), it does not work.

Luthaf:
ok, but then why do you need the `piplite.install` call?

agoscinski:
I don't fully understand this either.

My impression is that you need to install chemiscope in the building to install the javascript part of chemiscope (otherwise you get a javascript error running the widget). And you need to install again the python part of chemiscope for the pyodide kernel, because in the building it is just installed as a regular python package. The pyodide package installation happens when it is run in the browser, but I think it could be also installed during building by putting the dependency in some other file, but I haven't figured it out yet.

PR #289, opened 2023-08-23
Utility functions to draw shapes that visualise tensors and vectors

ceriottm:
Added utility functions to visualize positive-definite tensors (polarizability, thermal ellipsoids, ...) as well as vectors (forces, dipoles, ....). The latter required changing (switching off) the centering of the shapes. I think this is better, because you can always center (and add a utility function for that) a custom shape, but you can't draw an off-center shape if vertices get centered automatically. 

Also fixed `chemiscope.show` so it takes a `shapes` argument.

### DIFF
--- a/python/chemiscope/structures/_ase.py
+++ b/python/chemiscope/structures/_ase.py
@@ -1,6 +1,7 @@
 import warnings
 from collections import Counter
 
+from ._shapes import ellipsoid_from_tensor, arrow_from_vector
 import numpy as np
 
 try:
@@ -302,6 +303,66 @@ def _is_convertible_to_property(value):
                 return True
             except Exception:
                 return False
+
+
+def extract_vectors_from_ase(frames, key="forces", **kwargs):
+    """
+    Extract a vectorial atom property from a list of ase.Atoms
+    objects, and returns a list of arrow shapes. Besides the specific
+    parameters it also accepts the same parameters as
+    `array_from_vector`, which are used to define the style of the
+    arrows.
+
+    :param frames: list of ASE Atoms objects
+    :param key: name of the ASE atom property. Should contain
+       three components corresponding to x,y,z
+    """
+
+    vectors = []
+
+    for f in frames:
+        if key not in f.arrays:
+            raise IndexError(f"Key {key} not found in `Atoms.arrays`")
+        values = f.arrays[key]
+        if len(values.shape) != 2 or values.shape[1] != 3:
+            raise ValueError(
+                f"Property array {key} has not the shape of a list of 3-vectors"
+            )
+
+        # makes a list of arrows to visualize the property
+        vectors.append([arrow_from_vector(v, **kwargs) for v in values])
+
+    return vectors
+
+
+def extract_tensors_from_ase(frames, key="tensor", **kwargs):
+    """
+    Extract a 3-tensor atom property from a list of ase.Atoms
+    objects, and returns a list of arrow shapes. Besides the specific
+    parameters it also accepts the same parameters as
+    `ellipsoid_from_tensor`, which are used to draw the shapes
+
+    :param frames: list of ASE Atoms objects
+    :param key: name of the ASE atom property. Should contain
+       nine components corresponding to xx,xy,xz,yx,yy,yz,zx,zy,zz or
+       six components corresponding to xx,yy,zz,xy,xz,yz
+    """
+
+    tensors = []
+
+    for f in frames:
+        if key not in f.arrays:
+            raise IndexError(f"Key {key} not found in `Atoms.arrays`")
+        values = f.arrays[key]
+        if len(values.shape) != 2 or (values.shape[1] != 6 and values.shape[1] != 9):
+            raise ValueError(
+                f"Property array {key} has not the shape of a list of 6 or 9-vectors"
+            )
+
+        # makes a list of arrows to visualize the property
+        tensors.append([ellipsoid_from_tensor(v, **kwargs) for v in values])
+
+    return tensors
 
 
 def extract_lammps_shapes_from_ase(frames, key="shape"):
--- a/python/chemiscope/structures/_shapes.py
+++ b/python/chemiscope/structures/_shapes.py
@@ -1,0 +1,167 @@
+import numpy as np
+
+
+def center_shape(shape):
+    """
+    Takes a dictionary that describes a custom shape, and centers it, subtracting the
+    mean of the vertex positions. Returns a shallow copy, with a new list for the
+    vertices.
+
+    :param shape: A dictionary describing a custom shape
+    """
+
+    if shape["kind"] != "custom":
+        raise ValueError("Only `custom` shapes can be centered")
+
+    new_shape = {k: shape[k] for k in shape}
+    points = np.array(shape["vertices"])
+    points -= points.mean(axis=0)
+    new_shape["vertices"] = points.tolist()
+
+    return new_shape
+
+
+def _oriented_circle(radius, vec, n_points=20):
+    """
+    Generates a circle in 3D centered at the origin ands oriented according to the
+    given vector
+    """
+    # makes sure the normal is unit
+    nvec = vec / np.linalg.norm(vec)
+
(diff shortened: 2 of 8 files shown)


python/chemiscope/structures/_shapes.py
```diff
@@ -0,0 +1,180 @@
...
+            ]
+        )
+    # Compute the eigenvalues and eigenvectors
+    eigenvalues, eigenvectors = np.linalg.eigh(matrix)
+
+    if force_positive:
+        eigenvalues = np.abs(eigenvalues)
+
+    # The lengths of the principal axes are the square roots of the eigenvalues
+    old_settings = np.seterr(invalid="raise")
```
rosecers, line 155: Why are you bypassing this?

python/jupyter/src/widget.ts
```diff
@@ -48,6 +48,7 @@ class ChemiscopeBaseView extends DOMWidgetView {
             },
             this
         );
+        this._updatePythonSettings();
```
rosecers, line 51: Why this change?

python/chemiscope/structures/_shapes.py
```diff
@@ -0,0 +1,180 @@
...
+    points -= points.mean(axis=0)
+    new_shape["vertices"] = points.tolist()
+
+    return new_shape
+
+
+def _oriented_circle(radius, nvec, n_points=20):
+    """Generates a circle in 3D centered at the origin and oriented according to the given normal"""
+    # makes sure the normal is unit
+    nvec = nvec / np.linalg.norm(nvec)
```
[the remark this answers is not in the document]
ceriottm: Won't do - it changes nvec in place which is problematic. I will rather edit the code to have a different variable name so there's no risk of confusion.

python/chemiscope/structures/_shapes.py
```diff
@@ -0,0 +1,180 @@
...
+
+    # generate n_points in the plane defined by nvec, centered at vec
+    angles = np.linspace(0, 2 * np.pi, n_points, endpoint=False)
+    circle_points = np.outer(np.cos(angles), u) + np.outer(np.sin(angles), v)
+    circle_points = radius * circle_points
+
+    return circle_points
+
+
+def arrow_from_vector(
```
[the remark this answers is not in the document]
ceriottm: shading is OK, if not perfect. can we provide normals?

python/chemiscope/structures/_shapes.py
```diff
@@ -0,0 +1,180 @@
...
+        if m00 < -m11:
+            t = 1 - m00 - m11 + m22
+            q = np.array((m20 + m02, m12 + m21, t, m01 - m10))
+        else:
+            t = 1 + m00 + m11 + m22
+            q = np.array((m12 - m21, m20 - m02, m01 - m10, t))
+
+    q /= np.sqrt(q @ q)  # makes sure the quaternion is normalized
+
+    return q
```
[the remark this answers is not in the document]
ceriottm: OK. Didn't want to add a dependency on scipy just for this, but I see we have already one for the convex hull, so OK.

python/chemiscope/structures/_shapes.py
ceriottm: I'm not bypassing it. By default it's a warning, and I'm turning it to an error.

python/jupyter/src/widget.ts
ceriottm: as per commit comment - otherwise on the python side the settings are uninitialized unless accessed from the notebook, so when you show and then chemiscope.save, you'll get empty settings otherwise.

ceriottm: ok, my bad. I dug deeper and this was a red herring. problem was elsewhere, it's now fixed and this line has gone.

ceriottm: code review done right, thank you!

python/chemiscope/structures/_shapes.py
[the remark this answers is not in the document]
ceriottm: I think arrows look decent - perhaps we can take care of normals in a PR in which we revamp the shape definition format.

ceriottm:
@rosecers  I think I addressed all your comments. Nod if you're OK with merging now so we can move on to the next PR!

python/chemiscope/structures/_shapes.py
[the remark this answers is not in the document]
Luthaf: > shading is OK, if not perfect. can we provide normals?

in modern graphics API, normals are implicitly provided by the order of the indices/simplices. They are assumed to be in clockwise or counter-clockwise order, and the screw axis of the rotation is taken as the normal. The only thing I'm not sure about is if 3dmol is using clockwise or counter-clockwise ...

Luthaf:
Regarding arrows: should we have these as predefined shape type inside chemiscope? This will save a lot of space by not requiring to store all the vertices/simplices in the file, but instead generate them on the JS side like we do for ellispoids.

python/chemiscope/structures/_ase.py
```diff
@@ -304,6 +305,66 @@ def _is_convertible_to_property(value):
...
+    :param frames: list of ASE Atoms objects
+    :param key: name of the ASE atom property. Should contain
+       three components corresponding to x,y,z
+    """
+
+    vectors = []
+
+    for f in frames:
+        if key not in f.arrays:
+            raise IndexError(f"Key {key} not found in Atoms frame")
```
Luthaf, line 325:
```suggestion
            raise IndexError(f"Key {key} not found in `Atoms.arrays`")
```

this will avoid some confusion if people thing we are looking for stuff in `info`

python/chemiscope/structures/_shapes.py
```diff
@@ -0,0 +1,157 @@
...
+        ax = np.sqrt(eigenvalues[0])
+        ay = np.sqrt(eigenvalues[1])
+        az = np.sqrt(eigenvalues[2])
+    except FloatingPointError:
+        print(
+            "Non-positive definite tensor found with eigenvalues ",
+            eigenvalues,
+            ".\nIf this is acceptable, set `force_positive=True` to take the absolute values.",
+        )
+        raise
```
Luthaf, line 138:
```suggestion
    except FloatingPointError as e:
        raise ValueError(
            f"Non-positive definite tensor found with eigenvalues {eigenvalues}.\n"
            "If this is acceptable, set `force_positive=True` to take the absolute values.",
        ) from e
```

let's not use print as a way to handle/propagate errors

python/chemiscope/structures/_shapes.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # makes sure the rotation is proper
+    if np.linalg.det(eigenvectors) < 0:
+        eigenvectors[:, 0] *= -1
+
+    # Form the rotation matrix from eigenvectors
+    rotation = eigenvectors.T
+
+    # converts to quaternion
+    from scipy.spatial.transform import Rotation
```
rosecers, line 149: Move this to the top of the file

rosecers (commented):
I like @Luthaf's suggestion of an arrow as, essentially, a primitive type!

ceriottm:
> Regarding arrows: should we have these as predefined shape type inside chemiscope? This will save a lot of space by not requiring to store all the vertices/simplices in the file, but instead generate them on the JS side like we do for ellispoids.

I agree, but I would like to do this as part of a separate PR in which we address the input format too. In the spirit of doing small incremental PRs.

python/chemiscope/structures/_shapes.py
ceriottm: This is imported in place as for the part that computes simplices. Let's avoid a global dependency, particularly if we're going to make the arrows a JS-side primitive.

ceriottm: I added an exception check

Luthaf (approved):

Luthaf:
> I agree, but I would like to do this as part of a separate PR in which we address the input format too. In the spirit of doing small incremental PRs.

This is tracked in #291!

PR #290, opened 2023-08-25
Allow displaying multiple shapes simultaneously

ceriottm:
These are mostly changes to the UI, using multi-select as HTML widget.

Required some hacking of the HTMLOptions part which is why I'm also tagging @Luthaf


ceriottm:
Example .json demonstrating the functionality 

[shape_example.zip](https://github.com/lab-cosmo/chemiscope/files/12442765/shape_example.zip)

PR #292, opened 2023-08-29
Better input format for shapes

ceriottm:
The purpose of this PR is to slim down the format for storing structures in the input file. I think the functionality is sufficiently new that we don't need to worry about being backward compatible, but we should move relatively fast before it becomes an issue. Main point we want to address here is that storing vertices for each atom is a huge overhead if all that happens is that a structure is rotated or displaced. 

I suggest to proceed as follows. Shapes are still stored as entries in a dictionary. Each entry contains

```
export interface BaseShapeData {
    kind: string;
    settings:  {};
    frame_settings: undefined | {};
    atom_settings: undefined | {};
}
```

Each shape is generated by passing the settings to a function. If `frame_settings` or `atom_settings` are defined, then they must be a dictionaries where each parameter is an array containing the values of the parameters for each frame, or each atom in each frame. 

Then, shapes are created by looping over the appropriate entities, and calling the appropriate function with an argument list created by combining settings, frame_settings and atom_settings. 

For instance, one could have:

`{kind: "arrow", settings={'radius': 0.1}, frame_settings: {direction: [[0,0,1],[0,2,0]]} }`

which will plot an arrow in the specified direction, centered on the origin, for frames 1 and 2. 

`{kind: "arrow", frame_settings={'radius': [0.1, 0.2]}, atom_settings: {direction: [ [[0,0,1],[0.0,2]], [[0,2,0],[1,1,1]]]} }` 

will plot arrows with radius 0.1 in frame 1 (one for each of the two atoms) and arrows with radius 0.2 in frame 2. 

Incidentally, this format also addresses the use case in which one wants to have shapes that are not directly attached to atoms, e.g. polyhedra or isosurfaces.


Luthaf:
I agree with the idea here, it would be nice for smaller files by not repeating identical values. 

A couple of things:

- I would use `parameters` instead of `settings`
- frame_parameters and atoms_parameters should be mutually exclusive. If we need to use the same parameters across all frames, we can put them in the `parameters`. Having something different from one structure to the other, but the same for all atoms feels like a corner case that we don't need to support.
- if not provided, the `position` parameter will be `[0, 0, 0]`  for `frame_parameters` or the atom position for `atoms_parameter`. 

Another consideration: should `atoms_parameter` be named `particles_parameters`?

Overall, I would do something like this:

```ts
interface SomeShapeData {
    orientation: ...
    color: ...
    ...
}

export interface ShapeParameters<T> {
    kind: string;
    parameters:  {
        global: Partial<T>,
        structures?: Array<Partial<T>>,
        particles?: Array<Partial<T>>,
    }
}
```

-----

What is not covered by this proposal is the ability to have a handful of "template" shapes to pick between (e.g. have 3 different ellispoids, where only the position & orientation vary between particles). Not sure how much this would be worth it, but we should consider it now.

ceriottm:
> What is not covered by this proposal is the ability to have a handful of "template" shapes to pick between (e.g. have 3 different ellispoids, where only the position & orientation vary between particles). Not sure how much this would be worth it, but we should consider it now.

Not sure I understand what you mean here. You mean building a "template shape" composing multiple primitive and then placing them at different positions and orientations?

One major issue I see is that RN shapes are deeply associated with atoms - they are stored in the frames and not as a separate section of the dataset. This is quite a deep change and my re-implementation kind of stalled trying to disentangle shapes from frames.

Luthaf:
> Not sure I understand what you mean here. You mean building a "template shape" composing multiple primitive and then placing them at different positions and orientations?

something like this: we store could parameters once per template and then refer to a template in the shapes

```
templates = {
    A = {kind = "...", parameters: ...}
    B = {kind = "...", parameters: ...}
    C = {kind = "...", parameters: ...}
}

shapes = {
    "name" = {
        parameters = {
            global = {
                ...
            }
            particles = [
                { template: "A", position: ... },
                { template: "A", position: ... },
                { template: "B", position: ... },
                { template: "A", position: ... },
                { template: "A", position: ... },
                { template: "C", position: ... },
                { template: "C", position: ... },
                { template: "B", position: ... },
                { template: "B", position: ... },
            ]
        }
    }
}
```


---

> One major issue I see is that RN shapes are deeply associated with atoms - they are stored in the frames and not as a separate section of the dataset.

That's fine? This way we can easily display shapes in a standalone viewer as well. Why would we want to move the shapes outside of the structures? (I agree they should not be inside the atoms, but we don't have a concept of atom, cf https://chemiscope.org/docs/api/interfaces/Structure.html).

ceriottm:
OK, this is starting to get somewhere. @rosecers can you look at LAMMPS parsing and the validation code - that I have largely commented out ATM - while I fix the custom shapes and perhaps add a builtin arrow?

python/tests/shapes.py
```diff
@@ -42,7 +42,8 @@ def test_custom_shapes(self):
             numbers=[1, 1, 1], positions=[[0, 0, 0], [1, 1, 1], [2, 2, 5]]
         )
 
-        shapes = {
+        # TODO re-enable after having fixed custom shapes
```
rosecers, line 45: What is this?

src/dataset.ts
```diff
@@ -439,7 +485,7 @@ export function checkStructure(s: JsObject): string {
                 } else {
                     return `Chemiscope currently only supports custom, ellipsoid, or sphere shapes, got ${shape.kind}`;
                 }
-            }
+            } MCCOMMENT*/
```
rosecers, line 488: ?

rosecers (changes requested):
In general, I am very happy with this update. There is still some messiness and commented-out-things-that-shouldn't-be-commented-out in the code that needs to be fixed pre-merge, but all-in-all I'm very happy with the results. I've tried to break it a few times, to no avail, which says something. I've also added some joyful mischief.

docs/_static/chemiscope-icon.png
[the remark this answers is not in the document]
ceriottm: Fixed documentation glitch while working on the rest. Indulge me.

docs/src/conf.py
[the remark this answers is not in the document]
ceriottm: Same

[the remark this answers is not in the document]
ceriottm: Can we still be friends even if I add random shit to PRs?

python/chemiscope/input.py
```diff
@@ -165,56 +165,85 @@ def create_input(
...
+    and the parameters for each shape are obtained by combining the parameters from the
+    most general to the most specific - the parameters for atom `k` that is part of
+    structure `j` are obtained as
+
+    .. code-block:: python
+
+        global_parameters.update(structure_j).update(atom_k)
+
+    If given, the `structure` parameters list should contain one entry per structure, and
+    the `atom` parameters list should be a flat list with one entry per atom and structure.
```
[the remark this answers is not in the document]
ceriottm: Fine with me

python/chemiscope/structures/_ase.py
```diff
@@ -305,7 +305,38 @@ def _is_convertible_to_property(value):
...
+    if target == "structure":
+        try:
+            values = frame.info[key]
+        except IndexError:
+            raise IndexError(f"Key {key} not found in `Atoms.info`")
+
+    return values, target
+
+
+def extract_vectors_from_ase(frames, key="forces", target=None, **kwargs):
```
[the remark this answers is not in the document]
ceriottm: I agree, feel free to change it here and in the examples!

python/chemiscope/structures/_ase.py
```diff
@@ -320,22 +351,38 @@ def extract_vectors_from_ase(frames, key="forces", **kwargs):
...
 
-    return vectors
+    if target == "atom":
+        return {"kind": "arrow", "parameters": {"global": globs, "atom": vectors}}
+    else:
+        return {"kind": "arrow", "parameters": {"global": globs, "structure": vectors}}
 
 
-def extract_tensors_from_ase(frames, key="tensor", **kwargs):
+def extract_tensors_from_ase(frames, key="tensor", target=None, **kwargs):
```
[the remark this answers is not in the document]
ceriottm: Agree!

docs/_static/chemiscope-icon.png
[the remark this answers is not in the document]
rosecers: -_-

docs/src/conf.py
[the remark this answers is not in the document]
rosecers: ... within reason

python/chemiscope/input.py
```diff
@@ -165,56 +165,85 @@ def create_input(
...
```
rosecers, line 201:
```suggestion
    the `atom` parameters list should be a flat list corresponding to the atoms of each consecutive structure.
```

python/chemiscope/input.py
```diff
@@ -165,56 +165,85 @@ def create_input(
...
-            "orientation" [float, float, float, float],
         }
 
-    where ``orientation`` is an optional parameter corresponding to a quaternion in
-    ``x, y, z, w`` format. For ``custom`` shapes, ``simplices``, referring to the
-    *indices* of the facets, is also optional, and will be determined by convex
-    triangulation when not provided.
+    ``orientation`` is provided as a quaternion in ``x, y, z, w`` format.
+    For ``custom`` shapes, ``simplices``, refer to the *indices* of the vertices. If omitted, the mesh will be
+    determined by convex.
```
rosecers, line 246:
```suggestion
```

python/chemiscope/input.py
```diff
@@ -165,56 +165,85 @@ def create_input(
...
+        parameters = {
+            "global" : global_parameters,
+            "structure" : [ structure_1, structure_2, .... ],
+            "atom"" : [ atom_1, atom_2, .... ]
+        }
+
+    Each of these can contain some or all of the parameters associated with each shape,
+    and the parameters for each shape are obtained by combining the parameters from the
+    most general to the most specific - the parameters for atom `k` that is part of
+    structure `j` are obtained as
```
rosecers, line 194:
```suggestion
    Each of these can contain some or all of the parameters associated with each shape,
    and the parameters for each shape are obtained by combining the parameters from the
    most general to the most specific, i.e., if there is a duplicate key in the  `global` and `atom` 
    fields, the value within the `atom` field will supercede the `global` field for that atom. 
    The parameters for atom `k` that is part of structure `j` are obtained as
```

python/examples/shapes.py
```diff
@@ -0,0 +1,244 @@
...
+            },
+            "structure": [
+                {"vector": frames[0].info["dipole_ccsd"].tolist()},
+                {"vector": frames[1].info["dipole_ccsd"].tolist()},
+            ],
+        },
+    ),
+)
+
+dipoles_auto = chemiscope.extract_vectors_from_ase(frames, "dipole_ccsd", scale=0.5)
```
rosecers, line 189:
```suggestion
dipoles_auto = chemiscope.ase_vectors_to_arrows(frames, "dipole_ccsd", scale=0.5)
```

python/examples/shapes.py
```diff
@@ -0,0 +1,244 @@
...
+    properties=chemiscope.extract_properties(frames, only=["alpha"]),
+    shapes={
+        # cubes with smooth shading, centered on atoms
+        "smooth_cubes": smooth_cubes,
+        # demonstrates showing a "global" shape for each structure
+        "cube": sharp_cubes,
+        # (molecular) electric dipole
+        "dipole": dipoles_auto,
+        # atomic decomposition of the polarizability as ellipsoids. use utility to extract from the ASE frames
+        "alpha": chemiscope.extract_tensors_from_ase(
```
rosecers, line 210:
```suggestion
        "alpha": chemiscope.ase_tensors_to_ellipsoids(
```

python/examples/trajectory.py
```diff
@@ -0,0 +1,62 @@
...
+        "authors": ["The chemiscope developers"],
+        "references": [
+            'G. Fraux, R. Cersonsky, and M. Ceriotti, "Chemiscope: interactive structure-property explorer for materials and molecules," JOSS 5(51), 2117 (2020).'
+        ],
+    },
+    frames=frames,
+    properties=properties,
+    # visualize forces as vectors
+    shapes={
+        "forces": chemiscope.extract_vectors_from_ase(
```
rosecers, line 49:
```suggestion
        "forces": chemiscope.ase_vectors_to_arrows(
```

python/tests/shapes.py
rosecers: ^^

src/dataset.ts
rosecers: ^^

rosecers (commented):
Almost there! @ceriottm there are a few outstanding comments of mine to address, then passing all tests, and we should be almost there.

@Luthaf any objection to removing lammps port? Finding we don't need it with the new format

ceriottm:
Looking at current version. Can you @rosecers update the test? I'll go through the rest and polish up.

Luthaf:
> @Luthaf any objection to removing lammps port? Finding we don't need it with the new format

You mean the Python code to extract shapes from ase Atoms in LAMMPS format? Happy to remove it if it is no longer needed!

Luthaf:
> [I should really learn how to run eslint locally](https://github.com/lab-cosmo/chemiscope/pull/292/commits/c3e0518da14f123722e5fc735011c2c085d8016d)

You can set it up to automatically run inside vscode (https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) or other editors, or run it with `npx eslint` on command line.

ceriottm:
Hi @Luthaf  I think this is ready to merge. Will wait for you to have a last look. 
FYI - I didn't take in your suggestion to use `particles` for the shape parameters to stay consistent with the nomenclature we used for the properties `target`.

rosecers (approved):
Still going to make a stink that there are unrelated changes in this PR, but otherwise looks good to me.

python/chemiscope/structures/_ase.py
```diff
@@ -320,25 +351,41 @@ def extract_vectors_from_ase(frames, key="forces", **kwargs):
...
 
-    return vectors
+    if target == "atom":
+        return {"kind": "arrow", "parameters": {"global": globs, "atom": vectors}}
+    else:
+        return {"kind": "arrow", "parameters": {"global": globs, "structure": vectors}}
 
 
-def extract_tensors_from_ase(frames, key="tensor", **kwargs):
+def ase_tensors_to_ellipsoids(frames, key="tensor", target=None, **kwargs):
```
Luthaf, line 385:
```suggestion
def ase_tensors_to_ellipsoids(frames, key, target=None, **kwargs):
```

I would not set a default, since it will depend a lot on the application.

python/chemiscope/structures/_ase.py
```diff
@@ -320,25 +351,41 @@ def extract_vectors_from_ase(frames, key="forces", **kwargs):
...
     """
-    Extract a 3-tensor atom property from a list of ase.Atoms
-    objects, and returns a list of arrow shapes. Besides the specific
+    Extract a 2-tensor atom property from a list of ase.Atoms
+    objects, and returns a list of ellipsoids shapes. Besides the specific
     parameters it also accepts the same parameters as
     `ellipsoid_from_tensor`, which are used to draw the shapes
```
Luthaf, line 391: This also needs documentation for the `target` parameter

python/chemiscope/structures/_ase.py
Luthaf:
```suggestion
def ase_tensors_to_ellipsoids(frames, key="tensor", target=None, scale=1.0, force_positive=False):
```

Making the parameters explicit (and documented, even if just to point to the documentation for `ellipsoid_from_tensor`) will be nicer.

python/examples/shapes.py
```diff
@@ -0,0 +1,244 @@
+"""
+Extended shape visualization in chemiscope
+==========================================
+
+This example demonstrates how to visualize structure and atomic 
+properties in the structure panel, using different types of 
+visualizations (ellipsoids for tensors, arrows for vectors).
```
Luthaf, line 7:
```suggestion
properties in the structure panel, using different types of 
predefined shapes (ellipsoids for tensors, arrows for vectors).
```

python/examples/shapes.py
```diff
@@ -0,0 +1,244 @@
...
+The example also shows how to define custom shapes. 
+
+Note that the same parameters can be used with `chemiscope.show`
+to visualize an interactive widget in a Jupyter notebook.
+"""
+
+import ase.io as aseio
```
Luthaf, line 14:
```suggestion
import ase.io
```

python/examples/shapes.py
```diff
@@ -0,0 +1,244 @@
...
+import chemiscope
+import numpy as np
+from scipy.spatial.transform import Rotation
+
+# loads a dataset of structures
+frames = aseio.read("data/alpha-mu.xyz", ":")
```
Luthaf, line 20:
```suggestion
frames = ase.io.read("data/alpha-mu.xyz", ":")
```

python/examples/shapes.py
```diff
@@ -0,0 +1,244 @@
...
+# one can always update the defaults created by these automatic functions
+dipoles_auto["parameters"]["global"] = {
+    "base_radius": 0.2,
+    "head_radius": 0.3,
+    "head_length": 0.5,
+    "color": 0xFF00B0,
+}
+
+chemiscope.write_input(
+    "alphamu-shapes.json.gz",
```
Luthaf, line 199:
```suggestion
    "shapes-example.json.gz",
```

docs/src/tutorial/input-reference.rst
```diff
@@ -94,37 +94,44 @@ contains the following fields and values:
...
             },
             // other structures as needed
             ...
         ],
 
+        // OPTIONAL: shapes to display in the structure viewer. 
+        // Multiple shapes groups with different names are supported.
+        //
+        // Each shape is defined by parameters that can be specified globally,
+        // at the structure level, or for individual atoms
```
Luthaf, line 106: This should also contain an example/a description of all options for all shapes. It is a reference of the input format, not just an example.

src/structure/shapes.ts
```diff
@@ -15,6 +15,7 @@ import assert from 'assert';
 
 import { default as $3Dmol } from '3dmol';
 import { Quaternion } from '3dmol';
+type ColorSpec = number | string | { r: number; g: number; b: number; a?: number }; // mysterious compile errors if I import from 3dmol
```
Luthaf, line 18:
```suggestion
import { ColorSpec, Quaternion } from '3dmol';
```

this worked fine for me, are you still having compilation errors?

src/structure/shapes.ts
```diff
@@ -23,30 +24,71 @@ export interface XYZ {
...
+    head_radius?: number;
+    head_length?: number;
+}
+
+export interface ArrowParameters extends BaseShapeParameters<ArrowData> {
+    kind: 'arrow';
 }
 
 // Interface for polytope data, where
 // orientation is stored in the (w, x, y, z) convention
```
Luthaf, line 75:
```suggestion
// orientation is stored in the (x, y, z, w) convention
```

src/structure/shapes.ts
```diff
@@ -23,30 +24,71 @@ export interface XYZ {
...
+export interface EllipsoidParameters extends BaseShapeParameters<EllipsoidData> {
+    kind: 'ellipsoid';
+}
+
+// Interface for arrow data
+export interface ArrowData extends BaseShapeData {
+    vector: [number, number, number];
+    base_radius?: number;
+    head_radius?: number;
+    head_length?: number;
```
Luthaf, line 67:
```suggestion
    baseRadius?: number;
    headRadius?: number;
    headLength?: number;
```

We should use JS style here

src/structure/shapes.ts
ceriottm: yes still getting compilation errors

Issue #276, opened 2023-04-21
Include chemiscope in docs using sphinx


ceriottm:
It would be very nice to be able to include "passive" chemiscope viewers into autogenerated documentations, using sphinx or perhaps nbsphinxgallery.

ceriottm:
I think that generating a static image is the best way to proceed, if it's easy to do. The level of interactivity of the embedded ipywidgets is very limited in all cases because you can't really use the evenfullness outside individual widgets.

Luthaf:
The sphinx-gallery seems to be the easiest one to fix: we mainly need to add a `__repr_html__` function to `ChemiscopeWidget` and friends, 

https://github.com/lab-cosmo/chemiscope/blob/9b5c9b12b1b005a599488bf867a69d02f8c823e0/python/chemiscope/jupyter.py#L60

containing the same kind of HTML as the one we are using from the JS side:

https://github.com/lab-cosmo/chemiscope/blob/9b5c9b12b1b005a599488bf867a69d02f8c823e0/python/jupyter/src/widget.ts#L108-L134

The remaining question is how to insert the chemiscope javascript in there. We could load it from `https://chemiscope.org`, or one of the CDN based on top of npm?

sandipde:
This would be very useful. Not sure if you already know this ..ngl  viewers are possible to be included in the sphinxdoc directly from.notebook render without any additional steps.

ceriottm:
This is now fully implemented cf #340  and #341

PR #294, opened 2023-09-14
Display a warning when using chemiscope.show inside a sphinx-gallery example

ceriottm:
chemiscope.show() won't crash when processed by sphinx-gallery, just issue a warning. 

Also, returns an HTML error message to be displayed e.g. by sphinx-gallery

A stop-gap solution to allow including chemiscope in example noteboox while we see how to implement #276

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,14 +22,17 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
+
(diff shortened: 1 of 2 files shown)


ceriottm:
@Luthaf if you agree to merge this, can you also make a release so we can start incorporating chemiscopes into the software-cookbook?

Luthaf:
I think we should solve #292 before we can do a release. I'll comment over there as well

PR #297, opened 2023-10-12
Allow input of list of lists in the properties for python bindings

### LINKED ISSUE #296: Passing a list of lists to write_input gives an error.
The python bindings have the possibility of giving property values that are 2D numpy arrays, and automatically unpack that to `property[i]` - which is very useful for e.g. PCA maps. However, the function does not accept a list of lists, which is kind of confusing and inconsistent with the documentation.

ceriottm:
This fixes #296 . Not much to report, simple and to the point. @rosecers appreciate please I'm not bundling it in another PR I'll open shortly.

### DIFF
--- a/python/chemiscope/input.py
+++ b/python/chemiscope/input.py
@@ -704,22 +704,15 @@ def _linearize(name, property, n_structures, n_centers, multidimensional=False):
     data = {}
     values = property["values"]
     if not multidimensional:
+        if isinstance(values, list) and isinstance(values[0], list):
+            # convert to ndarray so we can use the parser below
+            values = np.array(values)
         if isinstance(values, list):
-            if isinstance(values[0], list):
-                for i in range(len(values[0])):
-                    try:
-                        values_i = [v[i] for v in values]
-                    except IndexError:
-                        raise IndexError("List property has inconsistent size")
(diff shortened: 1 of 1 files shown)


python/chemiscope/input.py
```diff
@@ -702,31 +702,41 @@ def _linearize(name, property, n_structures, n_centers, multidimensional=False):
...
+            if isinstance(values[0], list):
+                for i in range(len(values[0])):
+                    try:
+                        values_i = [v[i] for v in values]
+                    except IndexError:
+                        raise IndexError("List property has inconsistent size")
+                    data[f"{name}[{i + 1}]"] = {
+                        "target": property["target"],
+                        "values": _typetransform(values_i, name),
+                    }
```
Luthaf, line 717: Why not use `np.array(values)` here, letting numpy do all the types & size checks for us?

ceriottm: ok so you suggest to check if it's a list of lists, then convert to ndarray and go through the ndarray path? seemed hacky but you're right it simplifies the code a lot. will try this

ceriottm: Have a look I've streamlined this

PR #298, opened 2023-10-12
New examples - files

ceriottm:
Seems to me that we need to do one PR on gh-pages to add the example files, and one to master to do update the app creation scripts. This is the former.


Luthaf:
ROY example has a couple of issues:

- `warning: the 'name' property contains more than 50 different values, it can not be interpreted as categories` 
- Title should not use abbreviations: `GCH for ROY polymorphs` should expand on what GCH and ROY are
- in metatdata, `author` fields renders wrongly
- in metatdata, `references` fields renders wrongly
- PCA property should be constructed as a 2D property, letting chemiscope name it `pca [1], `pca [2]`, …

The dipole example looks clean, the only suggestion I'd make is to have slightly bigger arrow heads. Right now for large dipoles the arrow looks a lot like a cylinder.

ceriottm:
> ROY example has a couple of issues:
> 
> * `warning: the 'name' property contains more than 50 different values, it can not be interpreted as categories`
> * Title should not use abbreviations: `GCH for ROY polymorphs` should expand on what GCH and ROY are
> * in metatdata, `author` fields renders wrongly
> * in metatdata, `references` fields renders wrongly
> * PCA property should be constructed as a 2D property, letting chemiscope name it `pca [1]`, `pca [2]`, …
> 
> The dipole example looks clean, the only suggestion I'd make is to have slightly bigger arrow heads. Right now for large dipoles the arrow looks a lot like a cylinder.

Done most of these. Didn't spell out ROY because it's used colloquially  that way (see the info panel for the "real name" ... )

Issue #288, opened 2023-07-25
Example file with shapes


Luthaf:
Following up with #282, we would need an example of dataset with multiple shapes to add to the existing set of example.

ceriottm:
I think it would be nice to revamp a bit all the examples - we have much nicer applications done with chemiscope in the recent past - and to also include examples on how to make those visualizations from a jupyter notebook.

Luthaf:
Closed by #299

PR #299, opened 2023-10-12
New examples - master

ceriottm:
Seems to me that we need to do one PR on gh-pages to add the example files, and one to master to do update the app creation scripts. This is the latter.


ceriottm:
BTW this Fixes #288

ceriottm:
OK checked locally and all seems to work. Fingers crossed and I'll merge!

PR #303, opened 2023-10-16
Color the 3D structure by a property

### LINKED ISSUE #118: Feature request: allow to visualize tensorial properties on structures

MaximeScope:
It should finally be possible to merge the coloring of the 3d structure with master.

There is an additional option to color the background atoms by property
=> (Environments -> more options -> color: prop, style: ball and stick)

-----

EDIT:

Fixes #118


MaximeScope:
I should note that I've implemented a _color_function_ that includes a check for undefined values (such as when taking the square root of a negative value of the selected property). In such cases, the atom is assigned a gray color.

ceriottm:
I think you should start by adding a minimal example to the `python/examples` folder. I tested it on the chemical shielding dataset and my impression is that it doesn't really work but it's hard to make a reproducible example with this much data. 
I have it loaded with these parameters 
image
and I'm looking at structure 276, atoms 75 and 78. They have a CS_total of 24 and 26 (quite close on a range between 15 and 30) but they end up colored in bright red and bright blue as if they were pretty much extreme.

ceriottm:
Another observation is that IMO we REALLY need to provide a rudimentary color scale, otherwise it's very hard to read the figures. I read a bit and it seems the best way to do this would be to add an HTML entity outside the widget, a bit like the floating buttons in the top-right cornier.

Luthaf:
> I read a bit and it seems the best way to do this would be to add an HTML entity outside the widget, a bit like the floating buttons in the top-right cornier.

The idea was to us a custom label, for which we can set the background to an arbitrary HTML canvas, for which doing  a gradient would be fairly easy. 

But let's make sure the core of the functionality works first, merge it and then work on the color bar in a separate PR to reduce the size of the individual PRs.

src/structure/viewer.ts
```diff
@@ -1037,17 +1217,74 @@ export class MoleculeViewer {
...
                 radius: 0.15,
-            };
+                colorfunc: this.color_function(),
+            } as unknown as $3Dmol.StickStyleSpec;
         }
 
         return style;
     }
 
+    private color_function() {
```
Luthaf, line 1233:
```suggestion
    private colorFunction() {
```

src/structure/viewer.ts
```diff
@@ -1037,17 +1217,74 @@ export class MoleculeViewer {
...
+            grad = new $3Dmol.Gradient.RWB(max, min);
+        } else if (this._options.color.palette.value === 'Roygb') {
+            grad = new $3Dmol.Gradient.ROYGB(max, min);
+        } else if (this._options.color.palette.value === 'Sinebow') {
+            grad = new $3Dmol.Gradient.Sinebow(max, min);
+        }
+
+        if (property !== 'element') {
+            return (atom: $3Dmol.AtomSpec) => {
+                if (isNaN(Number(properties?.[property]?.[Number(atom.serial)]))) {
```
Luthaf, line 1279: Instead of re-accessing the full array everytime, you could create a variable for it & capture this variable in the function

Luthaf: Also, instead of `Number(atom.serial)` I would do an assert it is not undefined (since it should never be)

src/structure/viewer.ts
```diff
@@ -1310,4 +1566,16 @@ export class MoleculeViewer {
             }),
         };
     }
+
+    /** Changes the step of the arrow buttons in min/max input based on dataset range*/
+    private _setScaleStep(axisBounds: number[], name: 'x' | 'y' | 'z' | 'color'): void {
```
Luthaf, line 1571:
```suggestion
    private _setColorScaleStep(axisBounds: number[]): void {
```

src/structure/grid.ts
```diff
@@ -478,6 +489,41 @@ export class ViewersGrid {
         return this._resolvedStructures[index];
     }
 
+    private _getAtomProperties(): Record<string, Property> {
+        const numberProperties = filter(this._properties, (p) =>
+            Object.values(p.values).every((v) => typeof v === 'number')
+        );
+        const atomProperties = filter(numberProperties, (p) => p.target === 'atom');
+        return atomProperties;
+    }
```
Luthaf, line 498: Instead of storing all properties & then looking for properties containing number & with `target =  atom`, I would only store the ones we need in `this->_properties` and do this filtering in the constructor

MaximeScope:
> I think you should start by adding a minimal example to the `python/examples` folder. I tested it on the chemical shielding dataset and my impression is that it doesn't really work but it's hard to make a reproducible example with this much data. I have it loaded with these parameters image and I'm looking at structure 276, atoms 75 and 78. They have a CS_total of 24 and 26 (quite close on a range between 15 and 30) but they end up colored in bright red and bright blue as if they were pretty much extreme.

I have addressed the issue, and the code should now work correctly. Initially, the code was selecting the first atoms independently from the structure. With the help of the environnements, the visualizers should now accurately assign colors based on the correct atoms property value.

src/structure/grid.ts
MaximeScope: Fixed

src/structure/viewer.ts
MaximeScope: Fixed

src/structure/viewer.ts
MaximeScope: Fixed

src/structure/viewer.ts
MaximeScope: Fixed

ceriottm:
OK I have added an example that generates a simple color-coded figure. I have also renamed the RWB color scheme because it's blue first. I think one important thing is we can do to make it possible to have a reference for the colors is to make sure we have color schemes that match the plotly one. BWR is very similar to seismic (and this I have used in the example) but it'd be nice to have exactly the same schemes - at least viridis and magma for instance. 
Looks like you just need to sample the plotly/matplotlib palettes and hardcode say 16 intermediate points. Could you do this?

ceriottm:
Just commenting that this Fixes #118  for good

src/structure/grid.ts
```diff
@@ -120,11 +128,11 @@ export class ViewersGrid {
     /// List of structures in the dataset
     private _structures: Structure[] | UserStructure[];
     /// List of properties in the dataset
-    private _properties: { [name: string]: Property };
+    private _properties: Record<string, Property>;
```
Luthaf, line 131: Since you already filter out the string and structure properties

```suggestion
    private _properties: Record<string, number[]>;
```

and changing the comment to `List of per-atom properties in the dataset`

src/structure/grid.ts
```diff
@@ -28,6 +28,10 @@ import CLOSE_SVG from '../static/close.svg';
 import DUPLICATE_SVG from '../static/duplicate.svg';
 import PNG_SVG from '../static/download-png.svg';
 
+interface SelectiveEnvironment extends Environment {
```
Luthaf, line 31:
```suggestion
/// Extension of `Environment` adding the global index of the environment
interface NumberedEnvironment extends Environment {
```

src/structure/options.ts
```diff
@@ -116,8 +117,12 @@ export class StructureOptions extends OptionsGroup {
             palette: new HTMLOption('string', 'Rwb'),
         };
 
+        // Handling undefined propertiesName:
+        if (propertiesName === undefined) {
+            propertiesName = [];
+        }
```
Luthaf, line 123:
```suggestion
```

src/structure/options.ts
```diff
@@ -67,7 +66,11 @@ export class StructureOptions extends OptionsGroup {
     // Callback to get the initial positioning of the settings modal.
     private _positionSettingsModal: PositioningCallback;
 
-    constructor(root: HTMLElement, positionSettings: PositioningCallback, properties?: MapData) {
+    constructor(
+        root: HTMLElement,
+        positionSettings: PositioningCallback,
+        propertiesName?: string[]
```
Luthaf, line 72:
```suggestion
        propertiesName: string[] = []
```

you can setup different default values than undefined

src/structure/viewer.ts
```diff
@@ -810,6 +797,12 @@ export class MoleculeViewer {
         // ======= color settings
         // setup state when the property changes
         const ColorPropertyChanges = () => {
+            const properties = JSON.parse(JSON.stringify(this._properties)) as Record<
+                string,
+                (number | undefined)[]
```
Luthaf, line 802: I don't think the values can be undefined here, all the sparsity should be handled by the `environments` list

```suggestion
                number
```

src/structure/viewer.ts
```diff
@@ -891,23 +879,29 @@ export class MoleculeViewer {
...
-                properties[property] = properties[property].map((value) =>
-                    !isNaN(Number(value)) ? Math.sqrt(Number(value)) : NaN
-                );
-            } else if (mode === 'inverse') {
-                properties[property] = properties[property].map((value) =>
-                    !isNaN(Number(value)) ? 1 / Number(value) : NaN
-                );
+            if (property !== 'element' && mode === ('log' || 'sqrt' || 'inverse')) {
+                properties[property] = properties[property].map((value) => {
+                    if (value !== null) {
```
Luthaf, line 884: where would a `null` value comes from?

Luthaf:
> Just commenting that this Fixes #118 for good

This kind of comment needs to be in the first message of the PR for github to automatically close the issue

python/examples/colors.py
```diff
@@ -0,0 +1,84 @@
...
+                a.arrays["axy"],
+                a.arrays["axz"],
+                a.arrays["ayz"],
+            )
+        ]
+    )
+
+    a.arrays["anisotropy"] = (
+        a.arrays["alpha_eigenvalues"][:, 2] - a.arrays["alpha_eigenvalues"][:, 0]
+    )
```
Luthaf, line 48: Do we really need this in the "atomic coloring" example?

src/structure/options.ts
```diff
@@ -131,7 +131,7 @@ export class StructureOptions extends OptionsGroup {
             );
         }
         this.color.mode.validate = optionValidator(['linear', 'log', 'sqrt', 'inverse'], 'mode');
-        this.color.palette.validate = optionValidator(['Rwb', 'Roygb', 'Sinebow'], 'palette');
+        this.color.palette.validate = optionValidator(['bwr', 'Roygb', 'Sinebow'], 'palette');
```
Luthaf, line 134: These should have similar casing. Either `bwr`, `roygb` and `sinebow` or `Bwr`, `Roygb` and `Sinebow`. I could also see a point for `BWR`, `ROYGB` and `sinebow` since the first two refer to colors names.

src/structure/options.ts
MaximeScope: I don't see the problem here

src/structure/viewer.ts
MaximeScope: The null value came when there were no property value assigned on the atom. As for some atoms and properties in the Zeolite example.

python/examples/colors.py
MaximeScope: I don't know what you mean by that.

src/structure/options.ts
MaximeScope: Actually, there is a line in the code that Angelo was using to inverse max & min in the gradient. That's probably why it is BWR instead of RWB.

python/examples/colors.py
ceriottm: well, I thought it was better to avoid adding new data files, but we need scalar properties to display. 
I'll add a comment so it's clear it's not like one needs to do this in order to use the coloring.

src/structure/options.ts
ceriottm: agree. I'd vote for BWR, ROYGB and sinebow

Luthaf (approved):
Should be good to go once CI passes

Luthaf:
merged! 🎉 Thanks @MaximeScope and @angerossi for your work implementing this!

PR #306, opened 2023-10-20
chemiscope-input had been lost due to removal of setup.cfg

ceriottm:
We may say that the fact nobody noticed for a year means we should drop the tool altogether, but I think it makes sense to have it just for the old-school people who like the CLI. 

This simply adds it back to setup.py


Luthaf:
pyproject.toml is the modern replacement for setup.cfg, I would rather put the entry_point definition there

PR #309, opened 2023-10-25
Load function

bananenpampe:
Adds a load function to load the .json or .json.gz file after saving a jupyter widget

### DIFF
--- a/python/chemiscope/jupyter.py
+++ b/python/chemiscope/jupyter.py
@@ -101,6 +101,7 @@ class MapWidget(ChemiscopeWidgetBase):
     def __init__(self, data, has_metadata):
         super().__init__(data, has_metadata)
 
+
 def load(path, mode="default"):
     """
     Loads a dataset from given ``path``.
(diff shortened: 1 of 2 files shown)


ceriottm:
Looks good but I fear the semantics is a bit confusing. `save` is a method associated with a chemiscope widget, that saves its status. `load` should be similarly a method that loads a ".json" into an existing chemiscope widget. 
IDK if this is something that makes sense, and/or something you want to implement in this PR but I would rename the current `load` method as something like `read_input` or probably `show_input`

python/chemiscope/jupyter.py
```diff
@@ -101,6 +101,64 @@ class MapWidget(ChemiscopeWidgetBase):
     def __init__(self, data, has_metadata):
         super().__init__(data, has_metadata)
 
+def load(path, mode="default"):
```
ceriottm, line 105: I would rename this as show_input

python/chemiscope/jupyter.py
```diff
@@ -101,6 +101,64 @@ class MapWidget(ChemiscopeWidgetBase):
...
+    """
+    Loads a dataset from given ``path``.
```
ceriottm, line 107: I would indicate "Loads and show as a chemiscope widget" ...

ceriottm (changes requested):
I would also add something to the documentation.

bananenpampe:
It makes sense, yet it would require first initializing an empty widget and then loading the saved widget. I dont really see a usecase for it. I changed the names.

ceriottm:
> It makes sense, yet it would require first initializing an empty widget and then loading the saved widget into the empty one. I dont really see a usecase for it. I changed the names.

I am thinking of having a loaded CS, and then changing the file it has loaded. Might be interesting in interactive notebooks. I think however it's a corner case, and we could cover it in another PR if the need ever arises.

ceriottm (approved):

PR #315, opened 2023-11-05
A few bugfixes and cleanups for the Jupyter widgets

ceriottm:
Stumbled on a bunch of bugs and deprecation warnings  in the widgets and decided to fix them. 
Mainly:

* handle the case of partial environments definitions also for StructureWidget
* solve a deprecation warning and handles better the resizing of the plotly panel

### DIFF
--- a/python/chemiscope/jupyter.py
+++ b/python/chemiscope/jupyter.py
@@ -31,7 +31,6 @@ class ChemiscopeWidgetBase(ipywidgets.DOMWidget, ipywidgets.ValueWidget):
 
     def __init__(self, data, has_metadata):
         super().__init__()
-
         self.value = json.dumps(data)
         self.has_metadata = has_metadata
         if "settings" in data:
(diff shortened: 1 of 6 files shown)


python/chemiscope/jupyter.py
```diff
@@ -58,6 +57,19 @@ def save(self, path):
...
         file.close()
 
+    def __repr__(self, max_length=32):
+        # string representation of the chemiscope widget, truncating too large outputs
+        class_name = self.__class__.__name__
+        attributes = vars(self)
+        string_repr = ""
+
+        for key, value in attributes.items():
+            value_repr = repr(value)
```
Luthaf, line 67: I don't see how this helps with memory consumption: we are computing the full repr unconditionally here.

python/chemiscope/structures/_ase.py
```diff
@@ -303,7 +303,13 @@ def _is_convertible_to_property(value):
...
                 return True
             except Exception:
-                return False
+                try:
+                    if isinstance(np.array(value).flatten()[0], str):
+                        return True
+                    else:
+                        return False
+                except Exception:
+                    return False
```
Luthaf, line 312: I'm not sure what this is trying to detect. Could you add a test for it? Also, since it seems to be testing for string compatibility, it should not be in the "everything convertible to float" else block.

python/chemiscope/jupyter.py
ceriottm: the problem is that in some cases (e.g. when loading a previously-saved notebook) the whole string representation would be shown in the browser, killing the tab. this makes sure that the string representation of the widget remains sane.

ceriottm: ok this was indeed generating pretty nonensical representation, missing all the important bits. I have cleaned it up, now it does IMO the right thing and avoids generating a huge string with the whole dataset in it.

Luthaf (approved):

Luthaf:
I cleaned up the commits history, since they are unrelated I would prefer not to squash all of them in a single one.

PR #316, opened 2023-11-13
Python-side selection

### LINKED ISSUE #244: Changing active molecule from within API
Hi all,

I'm just having a first play around with chemiscope so apologies if there is a very simple answer I'm missing. I've currently got the library running through Flask and I'm trying to find a way to use the API to change the selected molecule in the `EnvironmentInfo`, the `PropertiesMap` and the `ViewersGrid` simultaneously, such that the server can say "now show structure no. 24" and it can push that to the client.

I've got this working already for the `EnvironmentInfo` and `ViewersGrid` by directly firing the `DefaultVisualizer.map.onselect()` callback with the structure ID I want to change to, but this doesn't update the active point in `PropertiesMap`. From what I can tell, I want to be using `DefaultVisualizer.map.activeChanged()` to do this all in one, but that requires a GUID as well as a structure ID.

Is this the correct way to go about this? If so, how do I find the GUID of the point within the Plotly plot to change to? I've been combing through the source code but so far haven't figured out how GUIDs are mapped to their respective points.

Thanks in advance,
Joe

ceriottm:
Traitlet-based synchronization of the selected structure. Facilitates interaction with other ipywidgets.
We should also use this to Fix #244  for good.


Luthaf:
What's the status on this? Should we close & reopen later when someone has time to implement it?

ceriottm:
well, it's actually kind of working. there are some problems handling multiple viewers and the synchronization with the map. I think it however might be good to keep this on hold until we have done something about #326, as I think one underlying issue is that the settings synchronization mechanism is not well documented and so there's a lot of guesswork.

Luthaf:
Sure, I'm just trying to clear PR no-one is working on. Should I close this PR and keep the branch around?

ceriottm:
I'd leave it on if you don't mind. I think it might take half an hour if we look at it together.

Luthaf:
We'll re-open this whenever we have time for it!

PR #319, opened 2023-12-04
A bug with the environment's coloring persists.

MaximeScope:
When a environment is enabled and further disabled, changing the color property while keeping the environments disabled indeed updates the color for every atom, although a ghost of the previous environment's coloring remains on top of the current coloring. If the environment is then re-enabled and further re-disabled by the user, the "ghost" will disappear. But, changing the color property again will result in the same problem.

I tried to fix this issue today but I can't understand where the problem comes from. I need help @Luthaf 

Although, I found another bug wwhere the color bar appeared when selecting a coloring per element again and fixed it!

### DIFF
--- a/src/structure/viewer.ts
+++ b/src/structure/viewer.ts
@@ -1065,7 +1065,7 @@ export class MoleculeViewer {
         // main style
         if (!this._environmentsEnabled()) {
             if (this._options.color.property.value !== 'element') {
-                alert('no environments to highlight');
+                console.log('no environments to highlight');
             }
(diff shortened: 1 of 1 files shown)


ceriottm:
I also noticed this problem, but have no clear idea how to fix it

Luthaf:
The issue was that we did not remove the GLModel from 3Dmol renderer when switching off the environments highlighting. It was not caught previously because both models had the same color anyway. This should be fixed by #322

PR #320, opened 2023-12-04
Updating the documentation.

MaximeScope:
Adding and explaining the new option "property" to "bgColor" in input-reference.rst from the docs.

Is there anything more to update?


ceriottm (approved):
Looks good as far as documentation goes. We should also add an example, but that requires two separate PRs

PR #321, opened 2023-12-08
Added cylinder shape type

ceriottm:
Very simple, self-contained PR to include a cylinder shape - started off to switch to cylinder when headLength was zero, but it seems cleaner to have a separate object even if it was a bit more coding.

### DIFF
--- a/python/chemiscope/structures/_ase.py
+++ b/python/chemiscope/structures/_ase.py
@@ -506,7 +506,7 @@ def ase_merge_pi_frames(bead_frames):
     Takes a list of lists of ase.Atoms objects, corresponding to
     the trajectories of the beads of a path integral calculation.
     Returns a consolidated trajectory in which all beads are in the
-    same frame, together with propertie and shapes that help visualize
+    same frame, together with properties and shapes that help visualize
     the path integral. The various pieces are returned as a dictionary
     and can be passed directly to `create_input` and related functions.
 
--- a/python/chemiscope/structures/_ase.py
+++ b/python/chemiscope/structures/_ase.py
@@ -501,7 +501,7 @@ def _get_shape_params_atom(prefix, shape_kind, dictionary, atom_i):
     return shape
 
 
-def ase_make_paths(bead_frames):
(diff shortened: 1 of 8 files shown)


ceriottm:
Wanted to make a path integral visualization and got a bit dragged away. Overall these are just a few improvements to the shapes functionality:
* I changed the way shapes are displayed so it is much faster to add many - basically all the vertices are added at once rather than one shape at a time
* I set things up so that when there are atom-based shapes with no explicit color definition the shapes will take the color of the corresponding atom.
* I added a utility to make path integral visualizations.
Ready to go, overall.

python/chemiscope/structures/_ase.py
```diff
@@ -499,3 +499,84 @@ def _get_shape_params_atom(prefix, shape_kind, dictionary, atom_i):
         )
 
     return shape
+
+
+def ase_make_paths(bead_frames):
```
Luthaf, line 504: Not a fan of this name, it's a bit too generic for my taste 😄. How about 

```suggestion
def ase_merge_pimd_frames(bead_frames):
```

Or something like it?

Luthaf (commented):
Could you also add the new options to the input reference?

ceriottm:
OK all done. @Luthaf  - since Rose has things to take care of these days, if you're happy I'd merge and move on.

python/chemiscope/structures/_ase.py
```diff
@@ -499,3 +499,84 @@ def _get_shape_params_atom(prefix, shape_kind, dictionary, atom_i):
...
 
     return shape
+
+
+def ase_merge_pi_frames(bead_frames):
+    """
+    Takes a list of lists of ase.Atoms objects, corresponding to
+    the trajectories of the beads of a path integral calculation.
+    Returns a consolidated trajectory in which all beads are in the
+    same frame, together with propertie and shapes that help visualize
```
rosecers, line 509:
```suggestion
    same frame, together with properties and shapes that help visualize
```

PR #322, opened 2023-12-11
Atoms color fixes

Luthaf:
Fix #319 and some related code and layout cleanup.

@MaximeScope could you give this a look and check that everything works as intended?

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -157,9 +157,13 @@ export class ChemiscopeApp {
 
         this.visualizer.structure.positionSettingsModal = (rect) => {
             const structureRect = getByID('chemiscope-structure').getBoundingClientRect();
+            let left = structureRect.left - rect.width + 25;
+            if (left < 25) {
+                left = 25;
+            }
             return {
                 top: structureRect.top,
-                left: structureRect.left - rect.width - 25,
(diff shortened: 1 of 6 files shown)


ceriottm:
Ghost environments are fixed, but the color scale remains even after switching to element coloring

Luthaf:
> had some issue with the color scale remaining even after switching to element coloring, but might have been down to non-updated widgets

If you manage to reproduce let me know I'll try to fix it.

Luthaf:
> when switching environments on and off and then on, it doesn't switch to environment highlighting until an atom is clicked on, IDK if that is intended or a bug

Should be fixed now

ceriottm:
> > when switching environments on and off and then on, it doesn't switch to environment highlighting until an atom is clicked on, IDK if that is intended or a bug
> 
> Should be fixed now, we just go back to the previously selected center

Works.

ceriottm:
fixed legend layout. @Luthaf  only thing I can't seem to manage to fix is the position of the modal, that at least on chrome seem to ignore completely what goes into positionsettingsrectangle

ceriottm:
Fixed also the alignment I think

PR #327, opened 2024-01-25
Fix a bug in atom-centered shape layout

ceriottm:
Somehow this slipped through the cracks. Arrows (and any atom-centered shape) 
were positioned at the origin rather than at the atomic sites. 

Just to upset @rosecers I also included a few unrelated changes to avoid datafile bloat,
and to make her happy I added some extra comments to the shape code.


Luthaf (approved):

rosecers:
-_-

PR #330, opened 2024-03-18
Modifying docs to simplify the discussion of basic chemiscope usage

ceriottm:
- [x] Move around some of the code section, separating more clearly the Python package and API
- [x] Make an examples section, ideally including (links to?) the generated chemiscopes


ceriottm:
It's not perfect, but - as the branch name says - it's "better" ^_^

Luthaf:
This looks good overall. I'll reduce the PNG file sizes before merging, but it's not urgent. 

Should we split the example into multiple pages instead of having all of them one after the other on the same page?

docs/src/examples.rst
```diff
@@ -0,0 +1,65 @@
...
+
+
+.. raw:: html
+
+    <h2>Trajectory data</h2>
+
+.. figure:: img/example_pca.png
+    :align: center
+
+    Trajectory data from a MD simulation of allyl alcohol, including
```
rosecers, line 21: This PCA is very sparse -- it might be good to add some description of what's going on, especially because it's the first example.

docs/src/manual/input.rst
```diff
@@ -18,63 +18,356 @@ that can help!
...
-`scikit-matter`_.
-The `ASAP`_ structural analysis package is another tool that can directly
-generate an output in chemiscope format.
 
+Input file reference
+--------------------
 
-``chemiscope`` functions reference
-----------------------------------
+If you can not or do not want to use the ``chemiscope`` python package to create
```
rosecers, line 42: Can we add a snarky tooltip here?

python/examples/colors.py
```diff
@@ -22,6 +22,8 @@
 anisotropy = []
 
 for frame in frames:
+    # center in the box
+    frame.positions += frame.cell.diagonal() * 0.5
```
rosecers, line 26: What's the purpose of this here?

rosecers (commented):
In general looks good, left a few q's and comments throughout, but nothing super serious

ceriottm:
> This looks good overall. I'll reduce the PNG file sizes before merging, but it's not urgent.
> 
> Should we split the example into multiple pages instead of having all of them one after the other on the same page?

I'm ambivalent. I do not want to have sub-headings in the menu (that's also why I use these weird raw HTML headings, thanks ChatGPT) but it's true that this is very long. Maybe we could use a style that automatically folds and unfolds the navbar? We could also have a list of examples.

python/examples/colors.py
ceriottm: For reasons I didn't investigate, if the molecule is at the origin it'll be shown folded into the box (i.e. split into pieces). this was to avoid it without figuring out this bug in this PR, because I'm scared Rose, very scared, by what you'd do to me otherwise.

docs/src/manual/input.rst
ceriottm: Yes we can. Please do it!

docs/src/examples.rst
ceriottm: Actually I was thinking of adding a truly basic example, load-from-ase-and-visualize as the first example. Would that be fine with you?

rosecers:
I honestly do not know how — @luthaf?

docs/src/examples.rst
ceriottm: Added a "simple example" does it help?

ceriottm:
@rosecers forgive me - I also touched the code, but I think that this falls within the "better docs"

Luthaf:
> can you also make a point release, so that the docs correspond to the version most users will see?

We don't need a release for this, docs are built from `main`

docs/src/manual/input.rst
Luthaf: @rosecers this looks relevant: https://sphinx-hoverxref.readthedocs.io/en/latest/usage.html

python/chemiscope/input.py
```diff
@@ -581,6 +583,62 @@ def _normalize_environments(environments, structures):
...
+    return {
+        "map": {
+            "x": {"property": x},
+            "y": {"property": y},
+            "z": {"property": z},
+            "color": {"property": color},
+            "size": {"property": size},
+            "symbol": symbol,
+        }
+        | map_settings,
```
Luthaf, line 631: this syntax requires Python 3.9 (https://peps.python.org/pep-0584/), could you update the minimal required Python version in pyproject.toml & CI?

ceriottm:
> > can you also make a point release, so that the docs correspond to the version most users will see?
> 
> We don't need a release for this, docs are built from `main`

I meant that since most users will install from pip, it'd be nice for them to get the `quick_settings()` from there. Seriously, this was one of the most serious "small hassle" in using CS so hopefully it'll help.

python/chemiscope/input.py
ceriottm: This is top-notch PR-review! Thanks, and done.

Luthaf:
right, I commented before seeing you added a function. I'll do a release!

rosecers (approved):

Luthaf:
Released in 0.6.1 !

PR #332, opened 2024-04-23
GLViewer.replicateUnitCell is no longer a no-op for 1x1x1

### LINKED ISSUE #331: Periodic neighbors of molecules are shown
I want to verify a dataset with my favorite visualization tool, but I obtain confusing view.

I have a test structure with one single water molecule but chemiscope is showing me my molecule plus 3 "ghost" atoms. 

chemiscope-structure
[h2o_mic.xyz.txt](https://github.com/lab-cosmo/chemiscope/files/15011264/h2o_mic.xyz.txt)

 I already talked to @Luthaf and I know that this is intended behavior and makes looking and crystals much nicer.

But, for molecules it can be very confusing - at least it is for me. I know that I can get rid of it if I move my atoms away from the origin or disable the PBC but this is not really useful if one wants to verify something and it is required to first manipulate them. 

I rather would have a bool viewer option like there is already one for `cell` etc or an addition to the already existing option to show the neighboring cells.

Luthaf:
So we should skip calling it in this case. 

Fix #331

### DIFF
--- a/src/structure/viewer.ts
+++ b/src/structure/viewer.ts
@@ -444,12 +444,17 @@ export class MoleculeViewer {
         };
 
         setup3DmolStructure(this._current.model, structure);
-        this._viewer.replicateUnitCell(
-            this._options.supercell[0].value,
-            this._options.supercell[1].value,
-            this._options.supercell[2].value,
-            this._current.model
-        );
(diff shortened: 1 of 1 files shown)


ceriottm (approved):
Tested, works, merge

PR #334, opened 2024-05-17
Fix of Sphinx Gallery display of Chemiscope Visualizer

sofiia-chorna:
TODO
# Opening the draft PR for the moment


docs/_static/js/chemiscope.min.js
Luthaf, line 1: This file should not be committed to the repo long term (since it is pretty big and auto-generated). It's fine for now, I'll help replace it with the generated version of the file later

docs/src/conf.py
```diff
@@ -33,6 +39,13 @@
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 
 
+sphinx_gallery_conf = {
+    "examples_dirs": os.path.join(ROOT, "python", "examples"),
+    "gallery_dirs": "examples",
+    "filename_pattern": ".*",
```
Luthaf, line 45: I guess this is required to include the JSON files?

docs/src/scraper.py
```diff
@@ -0,0 +1,95 @@
...
+        return f"""
+            <!-- Load all dependencies -->
+            <!-- JQuery -->
+            <script
+                src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.4.1/jquery.min.js"
+                integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo=" crossorigin="anonymous"
+            ></script>
+
+            <!-- JS scripts -->
+            <script src="../../../_static/js/chemiscope.min.js"></script>
```
Luthaf, line 53: You should load these from `gallery_dirs`, not the root!

docs/src/scraper.py
```diff
@@ -0,0 +1,95 @@
...
+            <script async="async" src="../../../_static/js/chemischope-sphinx-gallery.js"></script>
+
+            <!-- Font-awesome -->
+            <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.11.2/css/all.min.css" />
+            <link rel="icon" type="image/png" href="../../../_static/chemiscope-icon.png" sizes="32x32" />
```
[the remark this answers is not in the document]
sofiia-chorna: There is an icon for loading

docs/_static/js/chemiscope.min.js
sofiia-chorna: I see, thanks!

docs/src/scraper.py
```diff
@@ -0,0 +1,95 @@
...
+                "./_static/js/chemiscope.min.js",
+                "./_static/js/chemischope-sphinx-gallery.js"
+            ]
+            self.copy_chemiscope_files(target_dir, source_files)
+
+            # Use the custom html content
+            widget._repr_html_ = lambda: self.generate_html_content(widget.value)
+
+        # Use sphinx-gallery standard scraper
+        return matplotlib_scraper(block, block_vars, gallery_conf)
```
Luthaf, line 28: do we need to do this? since the matplotlib scrapper is already in the list of scrappers in the conf

ceriottm:
So this has made great progress. There are a few issues that need to be fixed IMO:

- [ ]  styling of the modals is messed up
- [ ] using an autogenerated chemiscope.min.js rather than explicitly uploading it
- [ ] the `loadChemiscopeSphinxGallery` uses always the default layout even if the chemiscope.show is using one of the simplified modes
- [ ] we need to mention in the documentation somewhere how to add chemiscope scrapers to a sphinx-gallery project

ceriottm:
One should check whether the CSS for the jupyterlab layout can be reused here, to avoid defining specialized layout code. Maybe however it's better to use something different, particularly if we can make it adapt to the layout and styling used for a given sphinx preset

setup.py
```diff
@@ -30,6 +30,7 @@ def run(self):
     # source files
     *glob.glob("src/**/*", recursive=True),
     *glob.glob("python/jupyter/src/**/*", recursive=True),
+    *glob.glob("python/chemiscope/**/*", recursive=True),
```
Luthaf, line 33: this is only listing TS files that enter the java script build, and there should not be any of these in `python/chemiscope`. Adding this line will trigger a rebuild of the javascript code too often (any time one of the Python file is modified) even if this rebuild is not necessary.

```suggestion
```

setup.py
```diff
@@ -41,6 +42,8 @@ def run(self):
 NPM_BUILD_OUTPUT = [
     "python/jupyter/nbextension/chemiscope.min.js",
     "python/jupyter/labextension/package.json",
+    "python/jupyter/nbextension/chemiscope.min.js",
```
Luthaf, line 45:
```suggestion
```

It's already above

utils/clean-python.sh
```diff
@@ -14,6 +14,7 @@ rm -rf MANIFEST
 
 rm -rf python/jupyter/labextension
 rm -rf python/jupyter/nbextension
+rm -rf python/chemiscope/
```
Luthaf, line 17: Why? This will remove the code! This file is intended to cleanup generated files only

```suggestion
```

sofiia-chorna:
- [ ] add a doc on how to set it up

ceriottm:
Works great and I think this only needs some polish before we can merge and release. Amazing work @sofiia-chorna  I'd say this is worthy of a point release at the very least, @Luthaf and I will also help starting on Tuesday. Meanwhile, something that clearly needs some polish is the documentation of this new feature. If I understand correctly, this can now be used in plain sphinx too - which is fantastic - but there's no example or explanation of how. @sofiia-chorna could you perhaps write down some docs? I'd rename the "use in sphinx-gallery" to a more generic "use with spinx" and discuss both direct inclusion and the generation from a python file using chemiscope.show.

ceriottm:
OK, it works - and so does the software cookbook so all is really nice, great job @sofiia-chorna . 
I'd just like to understand a bit better how it works - for instance, it looks like I should be able to add a chemiscope to an arbitrary rst file by writing

```
.. chemiscope::
:filepath: /home/michele/lavoro/code/chemiscope/docs/src/examples/datasets/fig_trajectory_005.json.gz
            :mode: default
```

in it, however this does not work. Could you add a documentation of how this is supposed to be used wit (and without) sphinx-gallery?

PR #337, opened 2024-05-30
Fix the version of bootstrap to fix the styling of modals

sofiia-chorna:


ceriottm:
wat!? is this really enough to fix it? I'm trying it on my end and it doesn't seem to make a difference.

ceriottm:
Checked - works!

ceriottm (approved):
Let's hope this won't break in the future, but for the moment bs 5.2 works

PR #340, opened 2024-06-11
Sphinx directive for chemiscope

sofiia-chorna:
Implemented a sphinx extension for the chemiscope python module that allows to include a chemiscope visualizer in a sphinx documentation file. 

Changes:
-  Created a chemiscope directive
-  Added `chemiscope.sphinx` extension to be included in `docs/conf.py`
-  Updated documentation to include usage examples and instructions


ceriottm (approved):
Ok! I  made already the changes I wanted to make. Good for me and ready for GF

ceriottm (approved):

docs/src/chemiscope-sphinx.rst
```diff
@@ -0,0 +1,74 @@
...
+
+.. code-block:: python
+
+   extensions = ["chemiscope.sphinx"]
+
+Chemiscope directive
+^^^^^^^^^^^^^^^^^^^^
+
+In order to include a chemiscope viewer into a RST documentation page one can use a 
+``chemiscope`` directive. When compiled with ``sphinx-builder``, the directive will
```
Luthaf, line 25:
```suggestion
``chemiscope`` directive. When compiled with ``sphinx-build``, the directive will
```

docs/src/chemiscope-sphinx.rst
```diff
@@ -0,0 +1,74 @@
...
+(``default``, ``structure``, or ``map``).
+
+For instance, to show a viewer combining structure and property panels you can 
+include the directive
+ 
+.. code-block:: rst
+
+    .. chemiscope::
+        :filepath: datasets/showcase.json.gz
+        :mode: default
```
Luthaf, line 39: Regarding syntax: I would take the path as the main argument to the directive, i.e.

```
.. chemiscope:: datasets/showcase.json.gz
    :mode: default 
```

and make `mode` default to "default" if not given. So this would be used like

```
.. chemiscope:: datasets/showcase.json.gz


.. chemiscope:: datasets/showcase.json.gz
    :mode: structure
```

docs/src/chemiscope-sphinx.rst
```diff
@@ -0,0 +1,74 @@
...
+and the ``map`` mode only the property map
+
+.. chemiscope::
+    :filepath: datasets/showcase.json.gz
+    :mode: map 
+    
+Viewing a documentation locally
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Because of CORS restrictions in most browsers, it is not possible to load the
```
Luthaf, line 63: is this CORS? I though CORS was about getting stuff from another webserver; while here the issue is more that we don't have such a server

python/chemiscope/sphinx/nodes.py
```diff
@@ -0,0 +1,66 @@
...
+
+from docutils import nodes
+
+
+class chemiscope(nodes.Element):
+    pass
+
+
+def visit_chemiscope_latex(self, node):
+    pass
```
Luthaf, line 13: should we at least leave some indication in the LaTeX document that we ignored a chemiscope file?

python/chemiscope/sphinx/static/css/chemiscope-sphinx.css
```diff
@@ -0,0 +1,86 @@
...
+    from {
+        transform: rotate(0deg);
+    }
+
+    to {
+        transform: rotate(360deg);
+    }
+}
+
+.visualizer-container {
```
Luthaf, line 46: I would scope these inside a chemiscope class, to prevent CSS pollution

python/chemiscope/sphinx/utils.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Parameters:
+    - app (Sphinx.application.Sphinx): The sphinx application object
+    - exception (Exception): The exception that occurred during Sphinx build, if any
+    """
+    if exception:
+        return
+    current_file_dir = os.path.dirname(__file__)
+    static_dir = os.path.join(current_file_dir, "static")
+    build_static_dir = os.path.join(app.outdir, "static")
```
Luthaf, line 30: I think there is an `html_static_dir` setting in sphinx (or with a similar name) that we should use here

Luthaf (commented):
Look great overall! I would tweak the directive syntax a bit to make it more convenient to use manually

python/chemiscope/sphinx/directive.py
```diff
@@ -0,0 +1,101 @@
...
+            rst_file_path = source_path + self.options.get("filepath")
+
+            # Copy dataset to the docs/build/html/_datasets folder
+            filename = os.path.basename(rst_file_path)
+            build_file_path, rel_file_path = self.get_build_file_path(filename)
+            copy_file(rst_file_path, build_file_path)
+
+            # Create the chemiscope node
+            node = self.create_node(
+                rel_file_path, source not in ChemiscopeDirective.pages_with_headers
```
Luthaf, line 52:
```suggestion
                rel_file_path, include_headers=source not in ChemiscopeDirective.pages_with_headers
```

docs/src/chemiscope-sphinx.rst
ceriottm: ok I agree this would be really neat - IDK how harder it'd be to write the custom directive and document it but yes if it's simple let's do this

docs/src/chemiscope-sphinx.rst
ceriottm: I _think_ that because of CORS we can't load the file from disk but right now the issue is that we use absolute paths relative to the server root for the includes so OK we can just say that one must use a server and that's it.

python/chemiscope/sphinx/nodes.py
ceriottm: Nice to have

python/chemiscope/sphinx/static/css/chemiscope-sphinx.css
sofiia-chorna: You mean like this?
```
<div class="chemiscope">
  <div class="visualizer-container"> ...</div>
</div>
```
And in css
```
.chemiscope {
  .visualizer-container { ... }
}
```

Luthaf: yes, something like this. Maybe the outer class can be `chemiscope-sphinx`?

python/chemiscope/sphinx/utils.py
sofiia-chorna: But I'm wondering that user can specify any folders here but in out html we rely exactly on the "static". Do you think it is worth passing static folder name and replacing it in the html template?

Luthaf: Ok, so I've been checking the docs for `html_static_path` https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_static_path, and it is actually a list of path that will be copied in full in the build output. It looks like the files are copied to `_static`, and this output path is not user-configurable.

Maybe a better solution here is to add our own static path (`chemiscope/sphinx/static`) to this variable (in the extension's `setup`); and load from sphinx's `_static` directory in the output (i.e. load from `/_static/js/chemiscope.min.js`). Does this make sense?

python/chemiscope/sphinx/directive.py
```diff
@@ -29,7 +28,7 @@ class ChemiscopeDirective(Directive):
     required_arguments = 0
```
Luthaf, line 28: I think we need `required_arguments = 1` for the file path

python/chemiscope/sphinx/directive.py
```diff
@@ -40,7 +39,8 @@ def run(self):
             source_path = os.path.dirname(source) + "/"
 
             # Path to the saved dataset in the .rst files folder
-            rst_file_path = source_path + self.options.get("filepath")
+            dataset_rel_path = self.content[0].strip()
```
Luthaf, line 42: what happens if the path is absolute? I think sphinx interpret absolute path as relative to the root of the documentation

ceriottm (approved):
I've a couple of mini-tweaks to text and error messages that I'm pushing RN, then we're good to go.

python/chemiscope/sphinx/directive.py
Luthaf: I'm fine with this, but then we need a check in the code that raise an exception if the user is giving an absolute path here.

sofiia-chorna: Ok, let me add it quickly !

Luthaf: well, you'll have to add it in the follow-up PR adding sphinx-gallery support, since this one is already merged =)

PR #341, opened 2024-06-12
Integration with sphinx-gallery

sofiia-chorna:
Uses `chemiscope.sphinx` functionalities to generate output for `chemiscope.show` calls in a sphinx-gallery example.


ceriottm (commented):
Looks good to me but let's wait @Luthaf as usual

sofiia-chorna:
I see those missing files errors are currently present on the https://chemiscope.org/docs/. Probably, we could fix it in the separate PR as I suggest it might not be related to the new development.

image

python/examples/shapes.py
```diff
@@ -212,8 +236,12 @@
...
     d["position"] = [11, 11, 11]
 
-chemiscope.write_input(
-    "shapes-example.json.gz",
+# %%
+#
+#
+# Uses a chemiscope widget to visualize structures including the custom shapes.
+# Even though we have property data, we show a `structure` view of the data
+cs = chemiscope.show(
```
Luthaf, line 244:
```suggestion
chemiscope.show(
```

python/examples/shapes.py
```diff
@@ -261,4 +289,11 @@
...
     },
     environments=chemiscope.all_atomic_environments(frames),
+    mode="structure",
 )
+cs
+
+# %%
+#
+# Save as a file that can be viewed at chemiscope.org
+cs.save("shapes-example.json.gz")
```
Luthaf, line 299:
```suggestion
```

python/examples/structure_map.py
```diff
@@ -11,27 +11,39 @@
 widget in a Jupyter notebook.
 """
 
+# %%
+#
+
```
Luthaf, line 16: why is this needed?

Luthaf: (and similarly for all other examples)

python/examples/structure_map.py
```diff
@@ -56,9 +70,12 @@
...
 
-# write the chemiscope input file.
-chemiscope.write_input(
-    path="trajectory-pca.json.gz",
+
+# %%
+#
+# Create and display the chemiscope widget
+
+cs = chemiscope.show(
```
Luthaf, line 78:
```suggestion
chemiscope.show(
```

python/examples/structure_map.py
```diff
@@ -83,3 +100,9 @@
         "structure": [{"keepOrientation": True, "playbackDelay": 100}]
     },
 )
+cs
+
+# %%
+#
+# Save as a file that can be viewed at chemiscope.org
+cs.save("trajectory-pca.json.gz")
```
Luthaf, line 108:
```suggestion
```

python/examples/trajectory.py
```diff
@@ -30,9 +35,13 @@
...
-# write the chemiscope input file.
-chemiscope.write_input(
-    path="trajectory-md.json.gz",
+
+# %%
+#
+# Uses a chemiscope widget to visualize structures and properties.
+# Note how to display the forces as vectorial properties
+
+cs = chemiscope.show(
```
Luthaf, line 44:
```suggestion
chemiscope.show(
```

python/examples/trajectory.py
```diff
@@ -67,3 +76,9 @@
         ]
     },
 )
+cs
+
+# %%
+#
+# Save as a file that can be viewed at chemiscope.org
+cs.save("trajectory-md.json.gz")
```
Luthaf, line 84:
```suggestion
```

Luthaf (changes requested):

docs/_templates/sidebar-toc.html
```diff
@@ -1 +1 @@
-{{ toctree(maxdepth=2|toint, collapse=False, includehidden=theme_globaltoc_includehidden|tobool) }}
+{{ toctree(maxdepth=2|toint, collapse=True, includehidden=True) }}
```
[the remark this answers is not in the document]
ceriottm: The doc TOC had become too long to include all the level 2 headings without collapsing. In particular for the examples, and despite much trying and hacking I could not find a way with this theme to only show level 2 headings for some of the pages.

docs/src/python/reference.rst
```diff
@@ -0,0 +1,28 @@
...
+
+.. autofunction:: chemiscope.ellipsoid_from_tensor
+
+.. autofunction:: chemiscope.arrow_from_vector
+
+.. autofunction:: chemiscope.ase_vectors_to_arrows
+
+.. autofunction:: chemiscope.ase_tensors_to_ellipsoids
+
+.. _chemiscope-input-cli:
```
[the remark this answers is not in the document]
ceriottm: Well spotted

python/chemiscope/jupyter.py
```diff
@@ -321,3 +322,31 @@ def _is_running_in_notebook():
...
+
+    if not sphinx_build:
+        try:
+            import sphinx_gallery
+
+            sphinx_build = hasattr(sphinx_gallery, "gen_gallery")
+        except ImportError:
+            sphinx_build = False
+
+    return sphinx_build
```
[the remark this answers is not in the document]
ceriottm: Because otherwise we have warnings when including the widgets in pure sphinx.

python/chemiscope/sphinx/file_path_iterator.py
```diff
@@ -0,0 +1,44 @@
+import uuid
+
+from sphinx.errors import ExtensionError
+
+
+class FilePathIterator:
```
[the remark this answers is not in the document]
ceriottm: @sofiia-chorna

python/examples/structure_map.py
ceriottm: To separate the header from the first cell

python/examples/structure_map.py
ceriottm: I think it is nice to also show how to export the file, which requires defining a cs object and then displaying it.
This also means that when a user runs the examples from the CLI they'll get some output otherwise they'll just get an error.

Luthaf: Yes, but we don't have to do this in all examples, it is already showed in `base.py`.

Also, the examples can now be dowloaded as jupyter notebooks, which is what I expect most users will be doing. And finally, when running on the CLI, users should be using `chemiscope.write_input` directly and not go through jupyter just for fun 😃

ceriottm: I think people can legitimately want to run these example from python. I agree we could use write_input but that would mean copying and pasting all the arguments to show() that seems silly to me. maybe we could write_input and then have a show_input() call instead?

python/chemiscope/jupyter.py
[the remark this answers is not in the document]
Luthaf: But this code is never executed from pure sphinx? Sphinx does not run any Python code by itself.

python/examples/structure_map.py
Luthaf: I think that examples should be focused on one thing, and not everything at the same time.

> maybe we could write_input and then have a show_input() call instead?

I don't understand what you mean here. What would show_input do?

ceriottm: The examples are there to demonstrate chemiscope functionalities. The fact that they work both to output a static .json.gz and an interactive viewer seems to me a good thing. 

`show_input` loads an input file from disk and shows it. so we would first write_input the .json.gz and then load it to view it in a notebook.

python/chemiscope/jupyter.py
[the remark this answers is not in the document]
ceriottm: the chemiscope directive is run through sphinx. I mean matter-of-fact without this code the "chemiscope in a doc" pages contain the warning box

python/examples/structure_map.py
Luthaf: ok, fair enough

python/chemiscope/sphinx/scraper.py
```diff
@@ -0,0 +1,84 @@
...
+        - _block: Unused parameter, a .rst code block
+        - block_vars (dict): Variables from the executed code block
+        - _gallery_conf: Unused parameter, sphinx-gallery config
+
+        Returns:
+        - str: .rst directive for embedding the chemiscope visualization
+        """
+        # Retrieve the chemiscope widget from block variables
+        widget = block_vars.get("example_globals", {}).get("___")
+        mode = self.get_widget_mode(widget)
```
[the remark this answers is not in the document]
Luthaf: Not sure, the way I did this in the scrapper for metatensor docs (that will be replaced by this one ^_^) was to loop over all global variables, and render all the new ones. https://github.com/lab-cosmo/metatensor/blob/5ebb12f42a9df081fd6add2bde82c55696b0b916/docs/generate_examples/conf.py#L34-L37

python/examples/colors.py
```diff
@@ -81,11 +82,11 @@
     # in map panel we must indicate the list of environments (all atoms in this case)
     environments=chemiscope.all_atomic_environments(frames),
 )
-cs
 
 
 # %%
 #
-# Save as a file that can be viewed at chemiscope.org
+# The file can also be viewed in a notebook. Use `chemiscope.show` to bypass the creation
```
Luthaf:
```suggestion
# The file can also be viewed in a notebook. Use `chemiscope.show` above to bypass the creation
```

python/examples/shapes.py
```diff
@@ -289,11 +289,12 @@
...
-    mode="structure",
 )
-cs
+
 
 # %%
 #
-# Save as a file that can be viewed at chemiscope.org
-cs.save("shapes-example.json.gz")
+# The file can also be viewed in a notebook. Use `chemiscope.show` to bypass the creation
```
Luthaf:
```suggestion
# The file can also be viewed in a notebook. Use `chemiscope.show` above to bypass the creation
```

python/examples/structure_map.py
```diff
@@ -100,9 +101,10 @@
         "structure": [{"keepOrientation": True, "playbackDelay": 100}]
     },
 )
-cs
 
 # %%
 #
-# Save as a file that can be viewed at chemiscope.org
-cs.save("trajectory-pca.json.gz")
+# The file can also be viewed in a notebook. Use `chemiscope.show` to bypass the creation
```
Luthaf:
```suggestion
# The file can also be viewed in a notebook. Use `chemiscope.show` above to bypass the creation
```

python/examples/trajectory.py
```diff
@@ -76,3 +76,10 @@
         ]
     },
 )
+
+# %%
+#
+# The file can also be viewed in a notebook. Use `chemiscope.show` to bypass the creation
```
Luthaf:
```suggestion
# The file can also be viewed in a notebook. Use `chemiscope.show` above to bypass the creation
```

python/chemiscope/sphinx/file_path_iterator.py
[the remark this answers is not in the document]
sofiia-chorna: Added, thanks

Luthaf (approved):

PR #342, opened 2024-06-12
Fix paths for the chemiscope spinx on the website

sofiia-chorna:
Added `html_baseurl` to the `conf.py` but it is relevant only for the website ! In the website there is one more folder `doc` but if we run it locally there is no additional folder.

Here is the script  to check how it will work on the website:
```
tox -e docs
rm -rf tmp && mkdir tmp/
cp -r docs/build/html tmp/doc/
cd tmp 
python -m http.server
```


ceriottm:
Problem has been fixed by #343 and #344

PR #343, opened 2024-06-12
Fix the location of CS header files using relative paths

ceriottm:
Determines relative path of `_static` to find the chemiscope headers


ceriottm:
Note to self: if @Luthaf tells you to fix the paths , fix the paths.

PR #344, opened 2024-06-12
Fix relative path loading of chemiscope jsons

ceriottm:
More fixing of paths...


ceriottm:
Last time I merge like this promised

PR #346, opened 2024-06-19
chemiscope.explore

sofiia-chorna:
A new method offering a quick overview through computed properties and dimensionality reduction,
allowing to rapidly gain insights into the composition and structure of data.


ceriottm:
Hey @sofiia-chorna @bananenpampe  I've many concerns about t-SNE cf https://distill.pub/2016/misread-tsne/

bananenpampe:
I am aware of these shortcomimgs, but my counterargument would be, that the user will still inspect the dimensionality reduction and decide for themselves if they can deduce something from it. If the input was purely noise then nothing can be deduced from the clusters either.

Luthaf:
So there a couple of things we need to decide here:

1) what should be the default featurization scheme
2) how can the user customize the featurization scheme
3) how can the user give additional properties to display

For (1), my main requirement is to find something that is easy to install on all platforms we support (this includes Windows) and does not break. I'm a bit worried using MACE for this, since it is relying on torch and looks a bit unstable to me. Rascaline + SOAP has the advantage of not requiring a pre-trained model, but fails when trying to handle many atomic types simultaneously, and is currently a pain to install. DScribe could be an alternative option as well.

I'm not a bit fan of t-SNE either because (a) it is stochastic and can produce different results over multiple runs; and (b) adding new data can completely change the clustering. For the default, I feel like something simple and stupid like PCA/KPCA would be better. This does not prevent us from giving an example of how to use the code with t-SNE for people who like this dimensionality reduction better, but PCA/KPCA strikes me as a much safer default.

----

For (2) I would go with `chemiscope.explore(frames, featurize=some_function)`, where `some_function` takes the frames and returns a ndarray. The function would handle both feature calculation and dimensionality reduction. I think this is what's currently called `reducer`.

For (3) I would use the same kind of API as the rest of the function, passing an explicit `properties` argument. Then users who also want to extract properties from ASE Atoms can use `chemiscope.extract_properties`

sofiia-chorna:
Questions:
1. chemiscope.show has some optional arguments (settings, shape, enviroment, meta). Should it be possible to pass them through chemiscope.explore? Or the new method is rather a simplification of .show still available if more flexibility is needed
2. The examples should be inside software-cookbook?

sofiia-chorna:
Hello @ceriottm, thanks a lot for the feedback! I will address your points.

Regarding the docstrings of the function, we were experimenting a bit with the `IncrementalPCA` as the default featurizer, so I put writing the description of the function "on-hold" ^^

sofiia-chorna:
I added a small diagram but it's fine for me to delete it, if it makes sense.
The remaining datasets are relatively small (< 30 kb each), datasets with visualisation (to showcase the work of the featurizers) are fetched now from the `zenodo`.

And currently SOAP + PCA are used by default if no `featurize` argument is provided. The dependencies for it can be installed with `pip install chemiscope[explore]`, it was added to the script for `tox -e docs`.

docs/src/img/chemiscope-explore-algorithm.png
Luthaf, line 1: I don't think a diagram for this is worth it, it can be described in a single sentence in the docs

pyproject.toml
```diff
@@ -67,3 +67,16 @@ line_length = 88
...
+[project.optional-dependencies]
+explore = [
+    "dscribe",
+    "scikit-learn",
+]
+
+mace = [
+    "scikit-learn",
+    "mace-torch"
+]
```
Luthaf, line 82:
```suggestion
```

I think we decided we did not want other dependencies in chemiscope than the minimal ones

python/chemiscope/explore.py
```diff
@@ -0,0 +1,145 @@
...
+        :py:func:`extract_properties` or manually defined by the user.
+
+    :param str mode: optional. Visualization mode for the chemiscope widget. Can be one
+        of "default", "structure", or "map". The default mode is "default".
+
+    :return: a chemiscope widget for interactive visualization
+
+    Here is an example of usage with and without providing a function to do the
+    reprensentation and reduction. The frames are obtained by reading the structures
+    from a file that `ase <ase-io_>`_ can read, and performing PCA using `sklearn`_ on
```
Luthaf, line 37:
```suggestion
    from a file that `ase <ase-io_>`_ can read, and performing Kernel PCA using `sklearn`_ on
```

python/chemiscope/explore.py
```diff
@@ -0,0 +1,145 @@
...
+
+    .. code-block:: python
+
+        import chemiscope
+        from ase.io import read
+        from dscribe.descriptors import SOAP
+        from sklearn.decomposition import KernelPCA
+
+        # Read the structures from the dataset
+        frames = read("trajectory.xyz", ":")
```
Luthaf, line 48:
```suggestion
        frames = ase.io.read("trajectory.xyz", ":")
```

python/chemiscope/explore.py
```diff
@@ -0,0 +1,145 @@
...
+    reprensentation and reduction. The frames are obtained by reading the structures
+    from a file that `ase <ase-io_>`_ can read, and performing PCA using `sklearn`_ on
+    a descriptor computed with SOAP with `dscribe`_ library.
+
+    .. code-block:: python
+
+        import chemiscope
+        from ase.io import read
+        from dscribe.descriptors import SOAP
+        from sklearn.decomposition import KernelPCA
```
Luthaf, line 45:
```suggestion
        import ase.io
        import dscribe.descriptors
        import sklearn.decomposition
```

and then using explicit path makes it clearer where things come from

python/chemiscope/explore.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Read the structures from the dataset
+        frames = read("trajectory.xyz", ":")
+
+        # Basic usage with the default featurizer (SOAP + PCA)
+        chemiscope.explore(frames)
+
+        # Using a custom featurizer
+        chemiscope.explore(frames, featurize=soap_kpca)
```
Luthaf, line 54: this should be done after defining the function

python/chemiscope/explore.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+        def soap_kpca(frames):
+            # Compute descriptors
+            soap = SOAP(
```
Luthaf, line 59:
```suggestion
            soap = dscribe.descriptors.SOAP(
```

python/chemiscope/explore.py
```diff
@@ -0,0 +1,145 @@
...
+                species=["C"],
+                r_cut=4.5,
+                n_max=8,
+                l_max=6,
+                periodic=True,
+            )
+            descriptors = soap.create(frames)
+
+            # Apply KPCA
+            transformer = KernelPCA(n_components=2, gamma=0.05)
```
Luthaf, line 69:
```suggestion
            transformer = sklearn.decomposition.KernelPCA(n_components=2, gamma=0.05)
```

python/chemiscope/explore.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Use default featurizer
+    else:
+        X_reduced = soap_pca(frames)
+
+    # Add dimensionality reduction results to properties
+    settings = {"map": {}}
+    axis = ["x", "y", "z"]
+    for i in range(X_reduced.shape[1]):
+        properties[f"Component {i + 1}"] = X_reduced[:, i].tolist()
```
Luthaf, line 96: you don't need to manually separate the different dimensions, the other functions in chemiscope will do it for you

```suggestion
    properties["features"] = X_reduced
```

Also, this should check that there is not already a `features` key in the properties

Luthaf: we can also pick another name than `features` or `components`. I like features, since the function is called `featurize`

python/examples/explore.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+# %%
+#
+# Example with SOAP and KPCA
+# ++++++++++++++++++++++++++
+#
+# This part illustrates how to create a custom function for dimensionality reduction
+# as an argument (`featurize`) to `chemiscope.explore`. Inside this function, we perform
+# descriptor calculation using SOAP and then reduce the dimensionality with `Kernel PCA
```
Luthaf, line 72: This should also link to dscribe!

python/examples/explore.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+# %%
+#
+# Example with MACE-OFF and t-SNE
+# +++++++++++++++++++++++++++++++++++++
+#
+# In this part, we are going to define another `featurize` function that runs
+# calculation of desciptors with MACE-OFF and uses t-SNE for the dimensionality
+# reduction.
```
Luthaf, line 155: can you link to both MACE and t-SNE documentation here? Not everyone will be familiar with these

python/examples/explore.py
```diff
@@ -0,0 +1,295 @@
...
+# %%
+#
+# Provide the created function to `chemiscope.explore`.
+
+cs = chemiscope.explore(frames, featurize=soap_kpca)
+
+# %%
+#
+# Note: It is possible to add parallelization when computing the SOAP descriptors and
+# performing dimentionality reduction with KernelPCA by providing the `n_jobs`
```
Luthaf, line 121: does KernelPCA have a `n_jobs` parameter? I though this was only for dscribe

python/examples/explore.py
```diff
@@ -0,0 +1,295 @@
...
+#
+# In this part, we are going to define another `featurize` function that runs
+# calculation of desciptors with MACE-OFF and uses t-SNE for the dimensionality
+# reduction.
+#
+# The dependencies for this example can be installed with the following command:
+#
+# .. code:: bash
+#
+#     pip install chemiscope[mace]
```
Luthaf, line 161:
```suggestion
#     pip install mace-torch scikit-learn
```

python/examples/explore.py
```diff
@@ -0,0 +1,295 @@
...
+    # Apply t-SNE
+    perplexity = min(30, descriptors.shape[0] - 1)
+    reducer = TSNE(n_components=2, perplexity=perplexity, n_jobs=n_jobs)
+    return reducer.fit_transform(descriptors)
+
+
+# %%
+#
+# We can also provide the additional properties inside `chemiscope.explore`.
+# For example, let's extract diapole moment from the frames using the related helper
```
Luthaf, line 210:
```suggestion
# For example, let's extract dipole moment from the frames using the related helper
```

tox.ini
```diff
@@ -29,6 +29,7 @@ deps =
     ase==3.22.1
 
 commands =
+    pip install chemiscope[explore]
```
Luthaf, line 32: ideally we should not use pip in the commands section of tox. If you need to use the extra dependencies, you should be able to use `extras = explore`: https://tox.wiki/en/latest/config.html#extras

tox.ini
```diff
@@ -69,6 +70,8 @@ deps =
 allowlist_externals =
     npm
 commands =
+    pip install chemiscope[explore]
+    pip install chemiscope[mace]
```
Luthaf, line 74: Same as above

python/examples/explore.py
```diff
@@ -0,0 +1,295 @@
+"""
+Chemiscope.explore example
```
Luthaf, line 2: Do we want a single example with all different featurization schemes, or do we want one example with the default and then other examples with other schemes?

@bananenpampe @ceriottm

Luthaf (commented):
Could you add a README in the `python/examples/data/` directory explaining where the different xyz files come from?

pyproject.toml
sofiia-chorna: Yes, sure, they are only installed with the [tox.docs] right now.
Where should I add those dependencies then? To `docs/requirements` ?

python/examples/explore.py
sofiia-chorna: Yes, KernelPCA also has this parameter
https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.KernelPCA.html

python/chemiscope/explore.py
sofiia-chorna: Okay okay, I thought this way it would be more readable ^^ I will move it up

pyproject.toml
Luthaf: yes!

python/examples/explore.py
sofiia-chorna: In the previous paragraph there were already a link to dscribe.SOAP.
https://github.com/lab-cosmo/chemiscope/blob/a50c792fdfc56b4d809e2b2cd7b7ffb231a07190/python/examples/explore.py#L57

I could add it here too if it makes sense

Luthaf: yes, more links is good in this context

python/examples/explore.py
```diff
@@ -0,0 +1,296 @@
...
+==========================
+This example illustrates usage of different methods of dimensionality reduction and its
+visualisation with `chemiscope.explore`. The primary goal of this function is to enable
```
Luthaf, line 5:
```suggestion
visualisation with :py:func:`chemiscope.explore`. The primary goal of this function is to enable
```

Luthaf: And same thing below!

python/examples/explore.py
```diff
@@ -0,0 +1,296 @@
...
+# properties to be used in visualization.
+
+
+# %%
+#
+# Example with SOAP and KPCA
+# ++++++++++++++++++++++++++
+#
+# This part illustrates how to create a custom function for dimensionality reduction
+# as an argument (`featurize`) to `chemiscope.explore`. Inside this function, we perform
```
Luthaf, line 67:
```suggestion
# as an argument (``featurize``) to `chemiscope.explore`. Inside this function, we perform
```

RestructuredText uses double backtick for inline code (while markdown uses a single one). Could you update everything to with the right syntax?

python/examples/explore.py
```diff
@@ -0,0 +1,296 @@
...
+cs = chemiscope.explore(qm9_frames, featurize=mace_off_tsne, properties=properties)
+
+# %%
+#
+# Here we display the visualisation of the pre-computed data using the described
+# function for 6k structures taken from the `QM9
+# <https://jla-gardner.github.io/load-atoms/index.html>`_ dataset.
+
+fetch_dataset("mace-off-tsne-qm9.json.gz")
+chemiscope.show_input("data/mace-off-tsne-qm9.json.gz")
```
Luthaf, line 229: why is the map zoomed in here?

python/examples/explore.py
ceriottm: I think the idea here is to demonstrate the usage of chemiscope.explore more than to show cool visualizations - so I'd have enough examples to show how to use it in the basic mode, and how to create custom featurizers.

Luthaf (commented):
The example is a nice demonstration of the capabilities of `chemiscope.explore`, but could be made even better by explaining **why** someone would want to use it!

python/examples/explore.py
Luthaf: In this case we don't need to use `fetch_dataset` and instead always use a small set of frames. We also don't have to make an example with MACE/t-SNE, since custom featurizers can be demonstrated with the existing dscribe + KPCA example.

I would only keep two examples:

- basic, do nothing special
- custom featurizer + custom properties

----

There was a discussion of having more featurizers such as MACE/t-SNE in example as an easy to copy/paste version of the code for people who aren't familiar with this kind of things, should we move these to a separate (set of) examples?

python/examples/explore.py
bananenpampe: here we wanted to highlight a specific cluster more closely. specifically a cluster of zwitterions which are all nicely grouped.

python/examples/explore.py
bananenpampe: Where else would be a good place to show these examples?
Should there be a chemiscope-gallery  that shows/explains them?

python/examples/explore.py
Luthaf: Why not, this should be made clear in the accompanying text then!

python/examples/explore.py
Luthaf: These could still be in the same place IMO, but split into "explaining how to use chemiscope.explore in general" and "specific applications of chemiscope.explore" (one example file per application)

python/examples/explore.py
sofiia-chorna: Done

pyproject.toml
```diff
@@ -67,3 +67,11 @@ line_length = 88
 indent = 4
 include_trailing_comma = true
 known_first_party = "chemiscope"
+
+### ======================================================================== ###
+
+[project.optional-dependencies]
+explore = [
```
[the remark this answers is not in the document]
sofiia-chorna: Done

python/examples/explore.py
sofiia-chorna: Done

python/examples/explore.py
sofiia-chorna: Okay, I have separated it to generic and specific applications, let me know if I need change anything else

Luthaf (approved):

PR #348, opened 2024-06-25
Fix appling map settings x and y properties

sofiia-chorna:
Fixed bug: apply setting of min and max values for x and y properties of `map` :

```
settings = {
  'map': {
    'x': {
      'property': 'PCA1',
      "min": 42,    // to be applied
      "max": 180,   // to be applied
    },
    'y': {
      'property': 'PCA2',
      "min": 46,    // to be applied
      "max": 170,   // to be applied
    }
  }
}
```

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -666,18 +666,33 @@ export class PropertiesMap {
         if (this._options.hasColors()) {
             // setup initial color range (must do before setting up events)
             const determineColorRange = (optionMin: number, optionMax: number) => {
-                // By default, range values are set to zeros
-                if (optionMin !== 0 || optionMax !== 0) {
+                const minProvided = !isNaN(optionMin);
+                const maxProvided = !isNaN(optionMax);
+
+                const getMinMaxFromValues = () => {
+                    const values = this._colors(0)[0] as number[];
+                    return arrayMaxMin(values);
+                };
+
+                if (minProvided && maxProvided) {
                     if (optionMin <= optionMax) {
                         return { min: optionMin, max: optionMax };
                     }
                     sendWarning(
                         `The inserted min and max values in color are such that min > max! The default values will be used.`
                     );
+                } else if (minProvided) {
+                    // Calculate max value from values
+                    const { max } = getMinMaxFromValues();
+                    return { min: optionMin, max: max };
+                } else if (maxProvided) {
+                    // Calculate min value from values
+                    const { min } = getMinMaxFromValues();
+                    return { min: min, max: optionMax };
                 }
-                // calculate default range
-                const values = this._colors(0)[0] as number[];
-                return arrayMaxMin(values);
+
+                // Calculate default range
+                return getMinMaxFromValues();
             };
             const { min, max } = determineColorRange(
                 this._options.color.min.value,
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -666,22 +666,16 @@ export class PropertiesMap {
         if (this._options.hasColors()) {
             // setup initial color range (must do before setting up events)
             const determineColorRange = (optionMin: number, optionMax: number) => {
-                const minProvided = !isNaN(optionMin);
-                const maxProvided = !isNaN(optionMax);
+                const [min, max] = this.getAxisRange(optionMin, optionMax, 'map.color');
+                const minProvided = min !== undefined;
+                const maxProvided = max !== undefined;
 
                 const getMinMaxFromValues = () => {
                     const values = this._colors(0)[0] as number[];
                     return arrayMaxMin(values);
                 };
 
-                if (minProvided && maxProvided) {
-                    if (optionMin <= optionMax) {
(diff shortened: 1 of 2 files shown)


src/map/map.ts
```diff
@@ -1018,6 +1018,13 @@ export class PropertiesMap {
         layout.coloraxis.colorbar.len = this._colorbarLen();
         layout.coloraxis.showscale = this._options.hasColors();
 
+        // Set ranges for the axes
+        const setAxisRange = (min: number, max: number) => {
+            return min !== 0 && max !== 0 ? [min, max] : undefined;
+        };
+        layout.xaxis.range = setAxisRange(this._options.x.min.value, this._options.x.max.value);
+        layout.yaxis.range = setAxisRange(this._options.y.min.value, this._options.y.max.value);
```
Luthaf, line 1026: Do the other axis-looking properties have the right range? These should be the z axis and color axis.

sofiia-chorna: Yes, indeed, it could be fixed for the z-axis & color as well

src/map/map.ts
```diff
@@ -659,9 +665,20 @@ export class PropertiesMap {
         // ======= color axis settings
         if (this._options.hasColors()) {
             // setup initial color range (must do before setting up events)
-            const values = this._colors(0)[0] as number[];
-            const { min, max } = arrayMaxMin(values);
+            const determineColorRange = (optionMin: number, optionMax: number) => {
```
Luthaf, line 668: I could see a case where min/max is undefined, if the user only gives us one of the two. We should handle it

src/map/map.ts
```diff
@@ -659,9 +665,20 @@ export class PropertiesMap {
...
+                // range is provided in settings by user
+                if (optionMin !== 0 && optionMax !== 0) {
```
Luthaf, line 670: I don't see why 0 is not allowed here? We could have a range of `0, 4` or `-2, 0`.

sofiia-chorna: In fact, if user didn't define them, the values are set to zeros in the initialization of `MapOptions`:
```
this.color = {
    mode: new HTMLOption('string', 'linear'),
    property: new HTMLOption('string', ''),
    min: new HTMLOption('number', 0),
    max: new HTMLOption('number', 0),
};
```
Hence I check if this is not the case
```
...
if (optionMin !== 0 && optionMax !== 0) {
...
```

src/map/map.ts
sofiia-chorna: If user didn't set it, the value is zero (due to the initialization of MapOptions):
```
this.color = {
    ...
    min: new HTMLOption('number', 0),
    max: new HTMLOption('number', 0),
};
```

Luthaf: Right, I missed that. It's a bit unfortunate that we can not distinguish between "the user gave `min=-3` and not max" vs "the user gave `min=-3` and `max=0`". In the first case, we should compute the max from the array.

Should we change the default to `min: new HTMLOption('number', NaN)`, and check with `isnan` here? Using `NaN` like `undefined`.

sofiia-chorna: Indeed, good idea! Let me change it!

ceriottm:
while you are at it, did you also check what is done for the _structure_ color?

sofiia-chorna:
> while you are at it, did you also check what is done for the _structure_ color?

@ceriottm as far as I checked, the color range for the structure settings is applied

src/map/map.ts
```diff
@@ -1086,6 +1132,27 @@ export class PropertiesMap {
...
+        const maxProvided = !isNaN(max);
+
+        // At least one range value is specified. By default, zeros are set
+        if (minProvided && maxProvided) {
+            if (min <= max) {
+                return [min, max];
+            }
+            sendWarning(
+                `The inserted min and max values in ${axisName} are such that min > max! The default values will be used.`
+            );
```
Luthaf, line 1151: Is this actually an issue? Or does plotly switch the order and goes high to low?

src/map/map.ts
```diff
@@ -1086,6 +1132,27 @@ export class PropertiesMap {
         this._plot.addEventListener('wheel', () => {});
     }
 
+    /** Validate provided min max values */
```
Luthaf, line 1135:
```suggestion
    /** Validate min/max options provided by the user. We use `NaN` internally to mark missing values, which are then transformed into undefined by this function */
```

src/map/map.ts
sofiia-chorna: Yes, if min remains bigger than max and user tries to change that range in the popup, it raises `Maximum call stack overflow` error due to this function 

https://github.com/lab-cosmo/chemiscope/blob/47770ab70bee39995da479bcd68df7aca7b3103c/src/map/map.ts#L723-L736

(I suppose it does reset() to the previous value which is still min > max => loop)

Luthaf: 👍

Luthaf (approved):

PR #352, opened 2024-07-25
Disable "more options" button in structure options by default

sofiia-chorna:
Bug reproduction steps:
1. Open `Chemical Shieldings` in the examples in chemiscope.org
2. Open settings in one of the grid cells
3. Click on "more options" button near color
4. Click on "reset min/max" button
-> exception

image

The proposed fix is to disable "more options" initially as color property is `element` by default => no color bar is available

### DIFF
--- a/src/structure/viewer.ts
+++ b/src/structure/viewer.ts
@@ -269,6 +269,9 @@ export class MoleculeViewer {
         this._colorMoreOptions =
             this._options.getModalElement<HTMLButtonElement>('atom-color-more-options');
 
+        // Disable as by default color property is element, so no color bar available
+        this._colorMoreOptions.disabled = true;
+
         this._connectOptions();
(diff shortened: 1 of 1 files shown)


Luthaf:
This looks good! One small issue I found as well is that if the user selects another coloring, then open the extra settings, and switch again to `element`, the options are disabled, but still visible/expanded. Maybe we should automatically collapse it when switching back to `element`?

sofiia-chorna:
> This looks good! One small issue I found as well is that if the user selects another coloring, then open the extra settings, and switch again to `element`, the options are disabled, but still visible/expanded. Maybe we should automatically collapse it when switching back to `element`?

@Luthaf 
Thanks a lot for checking! I added a collapsing.

Luthaf (approved):

PR #357, opened 2024-08-13
chemiscope.metatensor_featurizer

sofiia-chorna:
The function that takes a metatensor model and returns a function to be used in `chemiscope.explore` as a featurizer.


pyproject.toml
```diff
@@ -75,3 +75,7 @@ explore = [
     "dscribe",
     "scikit-learn",
 ]
+
+torchscript_featurizer = [
```
Luthaf, line 79: I would name the feature & function `metatensor`, since this is not taking arbitrary TorchScript code. So `pip install chemiscope[metatensor]` and `chemiscope.metatensor_featurizer()`.

pyproject.toml
```diff
@@ -75,3 +75,7 @@ explore = [
...
+    "metatrain @ git+https://github.com/lab-cosmo/metatrain",
```
Luthaf, line 80: why metatrain? You should not need the training code to run an exported model

python/chemiscope/explore.py
```diff
@@ -30,6 +32,12 @@ def explore(frames, featurize=None, properties=None, mode="default"):
         with the atomic structures. Properties can be extracted from frames with
         :py:func:`extract_properties` or manually defined by the user.
 
+    :param environments: optional. List of environments (described as
+        ``(structure id, center id, cutoff)``) to include when extracting the
+        atomic properties. Can be extracted from frames with
+        :py:func:`all_atomic_environments` (or :py:func:`librascal_atomic_environments`)
+        or manually defined.
+
```
Luthaf, line 40: if the user gives a subset of environments, we need to make sure the per-atom features are only defined for these specific atom-centered environments. I would do this in a separate PR, since it is orthogonal to metatensor models.

python/chemiscope/explore.py
```diff
@@ -163,3 +173,194 @@ def soap_pca_featurize(frames):
     # Compute pca
     pca = PCA(n_components=2)
     return pca.fit_transform(feats)
+
+
+def torchscript_featurizer(model, extensions_directory="extensions/", device=None):
+    """
+    Create a featurizer function using a `metatrain`_  model.
```
Luthaf, line 180: So instead of just metatrain models, I think this should be able to take arbitrary metatensor models (metatrain is one way to create such model, but there are other ways).

python/chemiscope/explore.py
```diff
@@ -163,3 +173,194 @@ def soap_pca_featurize(frames):
...
+        """
+        Featurize the given frames using the loaded model.
+
+        :param list frames: A list of atomic structures to obtain systems
+
+        :returns numpy.ndarray: Last layer features for each atom.
+        """
+        opts = {"mtt::aux::last_layer_features": ModelOutput(per_atom=True)}
+        if isinstance(frames, list):
+            # Load the model
```
Luthaf, line 233: we should be able to load the model and create the calculator outside of this function, this way the `featurize` function only handle calculations, and the user gets an error earlier if the model does not exists or has some issue

python/chemiscope/explore.py
```diff
@@ -163,3 +173,194 @@ def soap_pca_featurize(frames):
...
+
+    def featurize(frames):
+        """
+        Featurize the given frames using the loaded model.
+
+        :param list frames: A list of atomic structures to obtain systems
+
+        :returns numpy.ndarray: Last layer features for each atom.
+        """
+        opts = {"mtt::aux::last_layer_features": ModelOutput(per_atom=True)}
```
Luthaf, line 231: I don't like that we are using an output internal to metatrain here. I would prefer to properly define a standard output for "features" and then use that. I'll be in the office tomorrow, we can have a chat then.

I think we could unify this and the output used in PLUMED for machine learning collective variables: https://www.plumed.org/doc-master/user-doc/html/_m_e_t_a_t_e_n_s_o_r.html, look at the "Collective variables and metatensor models" section of the code.

python/chemiscope/explore.py
```diff
@@ -163,3 +173,194 @@ def soap_pca_featurize(frames):
...
+        if isinstance(frames, list):
```
Luthaf, line 232: why not call the existing ASE calculator multiple times in a loop?

python/chemiscope/explore.py
sofiia-chorna: Since I primarily use the [MetatensorCalculator](https://docs.metatensor.org/latest/atomistic/reference/ase.html#metatensor.torch.atomistic.ase_calculator.MetatensorCalculator), I believe it takes care of this. I will change the docstring

python/chemiscope/explore.py
sofiia-chorna: This can be done with the `_normalize_environments` as it was implemented here ?

https://github.com/lab-cosmo/chemiscope/blob/49b1182a9f97a22fed7e4918148e0620bda5c0b7/python/chemiscope/input.py#L310-L315

sofiia-chorna:
The idea was to rename the branch but remain the PR (unsuccessfully ^^), finally I will keep this one for convenience of review

python/examples/data/extensions/rascaline/lib/librascaline.so
Luthaf, line 1: this should not be committed to the repo, it is compiled code that only works on linux (and for specific versions of torch as well)

python/examples/data/model.pt
Luthaf, line 1: same, ideally we would not commit a pre-trained model, but rather a script to create such model. 

Maybe something that computes k features like `f_i^k = \sum_j r_ij^k` would be a good simple example. I can help with writing this!

python/chemiscope/explore.py
```diff
@@ -197,6 +198,121 @@ def soap_pca_featurize(frames, environments=None):
     return pca.fit_transform(feats)
 
 
+def metatensor_featurizer(model, extensions_directory="extensions/", device=None):
```
Luthaf, line 201: I would add `check_consistency` here as well, to follow the API of all other metatensor integrations

python/chemiscope/explore.py
```diff
@@ -197,6 +198,121 @@ def soap_pca_featurize(frames, environments=None):
...
+    """
+    Create a featurizer function using a `metatensor`_ model to obtain the features
+    from structures with a pre-trained ``metatensor`` model.
+
+    Note: a model output is defined per-atom by default. The additional check of
+    consistency is run to verify the model.
```
Luthaf, line 207: model should run per frame if the user wants frame features & per atom if there are environements

python/chemiscope/explore.py
Luthaf, line 201: Default should be None, like for all other integrations


```suggestion
def metatensor_featurizer(model, extensions_directory=None, device=None):
```

python/chemiscope/explore.py
```diff
@@ -197,6 +198,121 @@ def soap_pca_featurize(frames, environments=None):
...
+        """
+        # Run the model
+        opts = {"features": ModelOutput(per_atom=True)}
+        output = model.run_model(frame, opts)
+
+        # Extract and return features
+        llf = output["features"]
+        return llf.block().values.detach().cpu().numpy()
+
+    def featurize(frames):
```
Luthaf, line 299: this should accept an `environements` parameter, which can then be translated to `selected_atoms` and given to the metatensor model

python/examples/6-explore.ipynb
Luthaf, line 1: please no jupyter notebook in the repo!

sofiia-chorna: Oh didn't noticed pushing it, thanks for catching

Luthaf:
I rebased and updated the example a bit to show the simple case of using a pre-trained model first. This should be good to go for me, @ceriottm do you want to give this a quick look?

Luthaf (approved):

PR #360, opened 2024-08-30
Use GZIP magic bytes to detect whether a file is compressed

### LINKED ISSUE #359: Obtaining "incorrect header check" in Sphinx
Dear all,

Thank you for the great project!

I am interested in embedding access to a database in a sphinx page through read the docs, and I followed the instructions in the documentation for this process.

I am testing it with the default Arginine-Dipeptide `.json`.

After building with read the docs, I get a yellow box saying `incorrect header check`, which disappears quite quickly after.

Because this file, and my other test file, both work when uploaded to chemiscope, I was hoping you could point me in the right direction to understanding this issue?

Luthaf:
This should fix #359.

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -125,7 +125,7 @@ export class ChemiscopeApp {
             }
         }
 
-        const dataset = readJSON(this.dataset, buffer);
+        const dataset = readJSON(buffer);
         await this.load(config as Configuration, dataset);
     }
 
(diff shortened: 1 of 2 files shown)


ceriottm (approved):
LGTM, and I tested it works - at least with tox and nox builds. Don't see why it wouldn't work on readthedocs.

PR #362, opened 2024-09-03
Use of environments in explore

sofiia-chorna:
Added a `environments` parameter in the `chemiscope.explore` and added an example of its usage

### DIFF
--- a/python/chemiscope/explore.py
+++ b/python/chemiscope/explore.py
@@ -111,6 +111,10 @@ def explore(frames, featurize=None, properties=None, environments=None, mode="de
 
     # Pick frames and properties related to the environments if provided
     if environments is not None:
+        # Sort environments by structure id and atom id
+        environments = sorted(environments, key=lambda x: (x[0], x[1]))
+
+        # Check structure indexes
         unique_structures = list({env[0] for env in environments})
         if any(index >= len(frames) for index in unique_structures):
             raise IndexError(
@@ -118,9 +122,6 @@ def explore(frames, featurize=None, properties=None, environments=None, mode="de
             )
 
         if len(unique_structures) != len(frames):
-            # Sort environments by structure id and atom id
-            environments = sorted(environments, key=lambda x: (x[0], x[1]))
-
             # Pick frames corresponding to the environments
             frames = [frames[index] for index in unique_structures]
 
--- a/python/chemiscope/explore.py
+++ b/python/chemiscope/explore.py
@@ -18,6 +18,9 @@ def explore(frames, featurize=None, properties=None, environments=None, mode="de
     cores for parallelization. PCA reduces the dimensionality to two components by
     default.
 
+    If `environments` is provided, frames and properties are filtered to include only
+    those related to the environments.
+
     :param list frames: list of frames
 
     :param callable featurize: optional. Function to compute features and perform
--- a/python/chemiscope/explore.py
+++ b/python/chemiscope/explore.py
@@ -204,7 +204,7 @@ def soap_pca_featurize(frames, environments=None):
 
 def _extract_environment_indices(envs):
     """
-    Extract environment indices per structure
+    Convert from chemiscope's environements to DScribe's centers selection 
 
     :param: list envs: each element is a list of [env_index, atom_index, cutoff]
     """
--- a/python/chemiscope/explore.py
+++ b/python/chemiscope/explore.py
@@ -69,7 +69,7 @@ def explore(frames, featurize=None, properties=None, environments=None, mode="de
 
 
         # Define a function for dimensionality reduction
-        def soap_kpca_featurize(frames):
+        def soap_kpca_featurize(frames, _environments):
             # Compute descriptors
             soap = dscribe.descriptors.SOAP(
                 species=["C"],
@@ -175,6 +175,12 @@ def soap_pca_featurize(frames, centers=None):
 
 
 def _pick_env_frames(envs, frames):
+    """
+    Get environment indices par structures and pick corresponding frames
+
+    :param: list envs: each element is a list of [env_index, atom_index, cutoff]
+    :param: list frames: list of frames
+    """
     grouped_envs = {}
     picked_frames = []
     for [env_index, atom_index, _cutoff] in envs:
--- a/python/chemiscope/explore.py
+++ b/python/chemiscope/explore.py
@@ -110,10 +110,9 @@ def explore(frames, featurize=None, properties=None, environments=None, mode="de
 
     # Use default featurizer
     else:
-        centers = None
         if environments is not None:
-            centers, frames = _pick_env_frames(environments, frames)
-        X_reduced = soap_pca_featurize(frames, centers)
+            frames = [frames[env_index] for env_index, _, _ in environments]
+        X_reduced = soap_pca_featurize(frames, environments)
 
     # Add dimensionality reduction results to properties
     properties["features"] = X_reduced
@@ -122,7 +121,7 @@ def explore(frames, featurize=None, properties=None, environments=None, mode="de
     return show(frames=frames, properties=properties, mode=mode)
 
 
-def soap_pca_featurize(frames, centers=None):
+def soap_pca_featurize(frames, environments=None):
     """
     Computes SOAP features for a given set of atomic structures and performs
     dimensionality reduction using PCA. Custom featurize functions should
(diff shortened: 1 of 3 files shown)


python/chemiscope/explore.py
```diff
@@ -30,6 +30,12 @@ def explore(frames, featurize=None, properties=None, mode="default"):
         with the atomic structures. Properties can be extracted from frames with
         :py:func:`extract_properties` or manually defined by the user.
 
+    :param environments: optional. List of environments (described as
+        ``(structure id, center id, cutoff)``) to include when extracting the
+        atomic properties. Can be extracted from frames with
+        :py:func:`all_atomic_environments` (or :py:func:`librascal_atomic_environments`)
```
Luthaf, line 36:
```suggestion
        :py:func:`all_atomic_environments`.
```

I think we can stop promoting `librascal_atomic_environments` now. I'll see about removing it altogether.

python/chemiscope/explore.py
```diff
@@ -63,7 +69,7 @@ def explore(frames, featurize=None, properties=None, mode="default"):
 
 
         # Define a function for dimensionality reduction
-        def soap_kpca_featurize(frames):
+        def soap_kpca_featurize(frames, _environments):
             # Compute descriptors
```
Luthaf, line 73:
```suggestion
        def soap_kpca_featurize(frames, environments):
            if environments is not None:
                raise ValueError("'environments' are not supported")
            # Compute descriptors
```

python/examples/6-explore.py
```diff
@@ -85,6 +84,18 @@ def fetch_dataset(filename, base_url="https://zenodo.org/records/12748925/files/
...
 
+# %%
+#
+# Besides this, it is possible to run the dimentionality reduction algorithm and display
+# specific atom-centered environments. They can be manually defined by specifying a list
+# of tuples in the format ``[index of structure, index of atom, cutoff]``, as shown in
+# this example.
+#
+# Alternatively, the environments can be extracted from the frames using the function
+# :py:func:`all_atomic_environments` or :py:func:`librascal_atomic_environments`.
```
Luthaf, line 95:
```suggestion
# :py:func:`all_atomic_environments`.
```

python/examples/6-explore.py
```diff
@@ -103,13 +114,14 @@ def fetch_dataset(filename, base_url="https://zenodo.org/records/12748925/files/
...
-# passed to the ``featurize`` function.
+# Define the function ``soap_kpca_featurize`` which takes two arguments
+# (``frames``, which contains the structures provided to
+# :py:func:`chemiscope.explore` and internally passed to the ``featurize`` function;
+# ``environments``,  optional aurgument with the atom-centered environments,
+# if they were provided to the :py:func:`chemiscope.explore`.
 
 
-def soap_kpca_featurize(frames):
+def soap_kpca_featurize(frames, _environments):
```
Luthaf, line 124:
```suggestion
def soap_kpca_featurize(frames, environments):
    if environments is not None:
        raise ValueError("'environments' are not supported")
```

python/examples/7-explore-advanced.py
```diff
@@ -62,7 +62,7 @@ def fetch_dataset(filename, base_url="https://zenodo.org/records/12748925/files/
 # return the reduced data.
 
 
-def mace_off_tsne(frames):
+def mace_off_tsne(frames, _environments):
```
Luthaf, line 65:
```suggestion
def mace_off_tsne(frames, environments):
    if environments is not None:
        raise ValueError("'environments' are not supported")
```

python/examples/7-explore-advanced.py
```diff
@@ -141,7 +141,7 @@ def mace_off_tsne(frames):
 # different mace calculator.
 
 
-def mace_mp0_tsne(frames):
+def mace_mp0_tsne(frames, _environments):
```
Luthaf, line 144:
```suggestion
def mace_mp0_tsne(frames, environments):
    if environments is not None:
        raise ValueError("'environments' are not supported")
```

Luthaf (commented):
> Plus, maybe it makes sense to add an example of possible usage of environments inside a custom featurizer.

yes, this would be nice!

Also, I'd like to see a test of what happen if the environments are not sorted (i.e. something like `[(0, 1, 3.5), (0, 3, 3.5), (0, 2, 3.5), (1, 0, 3.5), (0, 5, 3.5)]`

python/chemiscope/explore.py
```diff
@@ -98,13 +109,33 @@ def soap_kpca_featurize(frames):
...
+    # Pick frames and properties related to the environments if provided
+    if environments is not None:
+        # Sort environments by structure id and atom id
+        environments = sorted(environments, key=lambda x: (x[0], x[1]))
+
+        # Check structure indexes
+        unique_structures = list({env[0] for env in environments})
+        if any(index >= len(frames) for index in unique_structures):
+            raise IndexError(
+                "Some or more indices are greater than the length of the frames"
```
Luthaf, line 121:
```suggestion
                "Some structure indices in 'environments' are larger than the number of frames"
```

python/chemiscope/explore.py
```diff
@@ -158,8 +195,50 @@ def soap_pca_featurize(frames):
...
+    feats = soap.create(frames, centers=centers, n_jobs=n_jobs)
 
     # Compute pca
     pca = PCA(n_components=2)
     return pca.fit_transform(feats)
+
+
+def _extract_environment_indices(envs):
+    """
+    Extract environment indices per structure
```
Luthaf, line 207:
```suggestion
    Convert from chemiscope's environements to DScribe's centers selection 
```

python/chemiscope/explore.py
```diff
@@ -98,13 +109,33 @@ def soap_kpca_featurize(frames):
...
+        # Check structure indexes
+        unique_structures = list({env[0] for env in environments})
+        if any(index >= len(frames) for index in unique_structures):
+            raise IndexError(
+                "Some or more indices are greater than the length of the frames"
+            )
+
+        if len(unique_structures) != len(frames):
+            # Pick frames corresponding to the environments
+            frames = [frames[index] for index in unique_structures]
```
Luthaf, line 126: I'm not sure this is correct: after running it, the indexes in environments will no longer match the positions of items in the `frames` list.

python/chemiscope/explore.py
```diff
@@ -98,13 +109,33 @@ def soap_kpca_featurize(frames):
...
+
+            # Pick properties corresponding to the environments
+            properties = _extract_properties_by_environments(
+                properties, unique_structures
+            )
```
Luthaf, line 131: I don't understand why this is needed. chemiscope should be fine with extra values in the structure properties, and when staying in 'structure' mode we want to be able to use all of them.

python/chemiscope/explore.py
Luthaf: If this is an issue with DScribe, it should be handled in the DScribe adapter.

python/chemiscope/explore.py
sofiia-chorna: Indeed, we don't need it if we provide the environemnts to `chemiscope.show`

python/chemiscope/explore.py
sofiia-chorna: Done

python/chemiscope/explore.py
```diff
@@ -133,6 +145,29 @@ def soap_pca_featurize(frames):
...
+        # Check structure indexes
+        unique_structures = list({env[0] for env in environments})
+        if any(index >= len(frames) for index in unique_structures):
+            raise IndexError(
+                "Some structure indices in 'environments' are larger than the number of"
+                "frames"
+            )
+
+        if len(unique_structures) != len(frames):
+            # Pick frames corresponding to the environments
```
Luthaf, line 168:
```suggestion
            # only include frames that are present in the user-provided environments
```

python/examples/7-explore-advanced.py
```diff
@@ -62,7 +63,10 @@ def fetch_dataset(filename, base_url="https://zenodo.org/records/12748925/files/
 # return the reduced data.
 
 
-def mace_off_tsne(frames):
+def mace_off_tsne(frames, environments):
+    if environments is not None:
+        raise ValueError("'environments' are not supported")
```
Luthaf, line 68:
```suggestion
        raise ValueError("'environments' are not supported by this featurizer")
```

python/examples/7-explore-advanced.py
```diff
@@ -178,3 +185,75 @@ def mace_mp0_tsne(frames):
...
+
+    # Get atom-centered indices from environments
+    for [env_index, atom_index, _cutoff] in environments:
+        if env_index not in grouped_envs:
+            grouped_envs[env_index] = []
+        grouped_envs[env_index].append(atom_index)
+        unique_structures.add(env_index)
+    centers = list(grouped_envs.values())
+
+    # Pick the frames corresponding to the structure indices
```
Luthaf, line 216:
```suggestion
    # only include frames that are present in the environments
```

Luthaf (approved):
Looks good, I only have a couple of suggestions on doc/comment/error messages and then this should be good to merge!

PR #365, opened 2024-09-12
Better error messages and target selection logic

ceriottm:
This PR is to solve two bugs:
1. the error message when one requires atom target without specifying environments
2. the error logic when one requires a target explicitly and there aren't >=2 properties - I think we shouldn't do the check here because one could be asking for a structure-only visualization (this is how I found the problem). 

(2) can be fixed by NOT checking number of properties (which is what I propose, that is checked later on when displaying the map, if the map is requested) or by also passing the viz mode to getTarget (which would be a pain and IMO unnecessary).

### DIFF
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -302,7 +302,7 @@ export function getTarget(dataset: Dataset): DisplayTarget {
     // Target is specified in settings
     if (target !== undefined) {
         if (target === 'atom' && dataset.environments === undefined) {
-            throw new Error('To use "atom" target, "settings.properties" should be provided');
+            throw new Error('To use "atom" target, a list of environments should be provided');
         }
         if (getTargetProps(target).length < 2) {
             throw new Error(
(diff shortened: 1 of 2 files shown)


Luthaf:
> that is checked later on when displaying the map, if the map is requested

I could only find this assert that would be hit when passing too few properties:

https://github.com/lab-cosmo/chemiscope/blob/7fed36287a09e59243318c7391b2b5579800819c/src/map/options.ts#L102C8-L102C44

I would prefer if we gave a proper error message instead of an assertion failure in this case. Another solution would be to have a separate function for `getTargetStructureViewer`, since this is the one with different requirements.

ceriottm:
Agree, seems the easiest thing is to throw an error here.

sofiia-chorna (approved):
Looks and works good for me, unless @Luthaf has more comments, it is fine for me to merge! Thanks for the fix

PR #367, opened 2024-09-26
Remove functions that are no longer relevant

Luthaf:
`chemiscope.composition_properties` should now use `chemiscope.metatensor_featurizer` or something similar. If people really want one of these, we can add a custom featurizer for it, but I feel like the utility of composition features is very low.

`chemiscope.librascal_atomic_environements`: librascal is dead, long live it's successors!

This should be merged after #357 and #366 (to limit rebasing pain)

### DIFF
--- a/docs/src/python/reference.rst
+++ b/docs/src/python/reference.rst
@@ -13,8 +13,6 @@
 
 .. autofunction:: chemiscope.all_atomic_environments
 
-.. autofunction:: chemiscope.librascal_atomic_environments
-
 .. autofunction:: chemiscope.ellipsoid_from_tensor
 
 .. autofunction:: chemiscope.arrow_from_vector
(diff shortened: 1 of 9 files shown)


ceriottm (approved):
Looks good and sensible. Do we want to somehow deprecate these? The cross section should be small enough that very few ppl should be hurt but since this is a minor "point" release maybe we should be over-cautious? I'm approving this and feel free to ignore but I wanted to raise the issue.

Luthaf:
Since this is changing the API, I was more thinking of releasing the code as a new incompatible version (i.e. 0.8.0, not 0.7.4). I don't think these are used by anyone, so dropping without deprecation should be fine.

As an aside, we should try to see what we still want to change for a 1.0 release and start moving toward that!

PR #372, opened 2024-10-08
Release 0.8.0

Luthaf:
The main changes in this release are 

- Added support for `stk` structures in the Python package
- Added support for environments in `chemiscope.explore`
- Added featurizers based in metatensor models in `chemiscope.explore`
- Removed old functions from the Python API


ceriottm:
Can we wait until #371 is merged? It also changes the API by removing an option for size.
I actually have another quick PR I'd like to set up. Let's leave this open for a bit if that's OK.

Luthaf:
I need this for the plumed tutorial, but we can try to get #371 merged, as well as something else if you need it!

Although I don't think #371 is a breaking change, the only thing that would happen with the removed setting would be a warning about unknown setting.

Issue #369, opened 2024-10-07
chemiscope local web app size 0


bernstei:
I have an input file I'm creating with `write_input`, and I want to use the species of the atoms as a property to control some aspect of visualization, probably size. I add it to the `Atoms` object with `atoms.new_array("Z", atoms.numbers)`, and then interactive set size as coming from `"Z"`.  However, for Cu Al, Zs 29 and 13, I don't get circles that are off by a factor of ~2, but rather tiny specks, and reasonable size circles.

Is setting `size` from a property reasonably tested, such that this is surprising? Or is it just something that's not well supported?  Either way I can try to do a more complete bug report.

ceriottm:
Hello! It should work pretty well, this is something we do often. See e.g. this to give you an idea of how property value and size should compare
![Image](https://github.com/user-attachments/assets/1d4bff77-a957-420c-8b4c-815e92bf9a08)

One can never exclude a bug though. If you upload the chemiscope.json file, we can have a look.

ceriottm:
Actually, I think I know what your problem is. The size is "rescaled" so that the smallest value is "tiny" and the largest a given "reasonable" size. There isn't really a way to set it to be actually proportional to the value in an "absolute" way. One would basically need to add "range" properties as there are for other indicators such as color and axis range. 
It's not a bad "first issue" if you feel like having some typescript fun ^^

bernstei:
Maybe it's just that the tiny is too small in my system, perhaps because I have 2-3 widely separated clumps?

bernstei:
Even with size fixed, something weird is going on.  I did symbol by a categorical property (config_type), and the squares (circled in red pen) are much smaller than the circles.  Is this also normal?

![Image](https://github.com/user-attachments/assets/f413819b-c460-482e-bc87-cf68a21b1094)

ceriottm:
So, I tweaked a few things that should now work better, pls have a look. Your second issue looks strange. Here point size, color and z coordinates are all expressions of the same quantity, and all sizes look proportionate independent of the symbol
![Image](https://github.com/user-attachments/assets/8dda3935-45a2-403e-9eaf-0c977190adb9)
See if it also breaks with the branch (if you want to run locally you need the whole node.js stack, and launch `npm install; npm start` from the repo root, then you should be able to open localhost:8080 and see the web interface.

bernstei:
That's more like what I expected.  Let me check the `size-cleanup` branch and see how it behaves.

[...]

PR #373, opened 2024-10-09
Added trace line and marker outline options

ceriottm:
When plotting trajectories it 's nice to have a line connecting the dots. There is now an option for it. 
Also, there are cases in which the outlines of the points make it actually harder to see the colors, so I also added an option to remove the outlines.

### DIFF
--- a/python/chemiscope/input.py
+++ b/python/chemiscope/input.py
@@ -643,7 +643,7 @@ def quick_settings(
 
     computed_map_settings.update(
         {
-            "traceLine": trajectory,
+            "joinPoints": trajectory,
         }
     )
 
@@ -659,8 +659,8 @@ def quick_settings(
         computed_structure_settings.update(structure_settings)
 
     return {
-        "map": map_settings,
-        "structure": [structure_settings],
+        "map": computed_map_settings,
+        "structure": [computed_structure_settings],
     }
 
 
--- a/python/chemiscope/input.py
+++ b/python/chemiscope/input.py
@@ -627,14 +627,11 @@ def quick_settings(
             structure viewer (following the chemiscope settings schema).
     """
 
-    if map_settings is None:
-        map_settings = {}
-
-    if structure_settings is None:
-        structure_settings = {}
-
-    map_settings.update(
-        {
+    if (x + y + z + color + size + symbol) == "":
+        # if at least one of the properties is requested
+        computed_map_settings = {}
+    else:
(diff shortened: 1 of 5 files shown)


python/chemiscope/input.py
```diff
@@ -627,30 +627,36 @@ def quick_settings(
...
             "symbol": symbol,
         }
-    )
 
-    structure_settings.update(
+    computed_map_settings.update(
         {
-            "keepOrientation": trajectory,
-            "playbackDelay": 10 if trajectory else 700,
+            "traceLine": trajectory,
```
Luthaf, line 645: Should we check if the rendering mode is per-structure before setting this? I might be weird for per-atom properties

src/map/map.ts
```diff
@@ -535,8 +535,8 @@ export class PropertiesMap {
                 color: this._colors(0)[0],
                 coloraxis: 'coloraxis',
                 line: {
-                    color: this._lineColors(0)[0],
-                    width: 1,
+                    color: 'black',
+                    width: this._options.markerOutline.value ? 0.5 : 0,
```
Luthaf, line 539: I'm not sure this really solve the issue in https://github.com/lab-cosmo/chemiscope/issues/369#issuecomment-2400722768: if there are a lot of black points together, they will still smudge together.

I'm fine with reducing the width to 0.5, but I would rather use black outline for light colors and a grey outline (since the background is white) for dark colors. I'm sure there is a package to compute color luminance in JS.

python/chemiscope/input.py
ceriottm: good point. IDK if it's actually feasible in this function ATM but I'll look into it.

src/map/map.ts
Luthaf: I agree with you that it should not be a fully customizable plotting code, but you are the one adding a customization option here 😃 Anyway, seems that most people want the option, so let's do it.

src/map/options.ts
```diff
@@ -80,6 +80,8 @@ export class MapOptions extends OptionsGroup {
         mode: HTMLOption<'string'>;
         property: HTMLOption<'string'>;
     };
+    public markerOutline: HTMLOption<'boolean'>;
+    public traceLine: HTMLOption<'boolean'>;
```
Luthaf, line 84: this name feels a bit too plotly specific. How about `joinPoints` instead? I'll update the code if the name works for you!

PR #376, opened 2024-10-09
Post-release patches

ceriottm:
Solution courtesy of chatty

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "construct-style-sheets-polyfill": "^3.1",
         "css-loader": "^7",
         "dts-bundle-generator": "^9",
-        "eslint": "^9.12",
+        "eslint": "^9.12.0",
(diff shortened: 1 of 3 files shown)


ceriottm:
@Luthaf IDK if this needs also to do a new release, probably not given it only affects building from source.

Luthaf:
The code was building fine for me without the changes (and is building fine on main), does it still fail for you if you clone a fresh main & run `npm ci`?

ceriottm:
Ok it actually compiles from a clean clone. I see why it does that for the packages.json, not what triggers the typescript error. I'm closing this, we can return to this at a later stage, or at least include some info in Contributing.md about packages that should be pinned, and stuff that shouldn't be touched in packages.json

Luthaf:
One of the updated depencency was the typescript compiler itself, so the old version did not understand the code anymore.

PR #383, opened 2024-10-17
Fix symbol and map glitches

### LINKED ISSUE #380: "x" symbol size too big
The "fixed" size "x" symbols are much bigger than squares or circles (see screenshot)

Image

ceriottm:
Fixes #380 - or at least, patches up the most glaring problem and adds a bunch of symbols by monkey-patching plotly. The symbols are glyphs from fonts, so size depends on which font is picked that seems to be browser dependent.  Also tackles a couple of small glitches when switching back and forth from 3D/2D

### DIFF
--- a/src/map/options.ts
+++ b/src/map/options.ts
@@ -23,14 +23,7 @@ import HTML_OPTIONS from './options.html.in';
 // in 3D mode, only strings are supported for 'marker.symbol', and only very few
 // of them. See https://github.com/plotly/plotly.js/issues/4205 as the plotly
 // issue tracking more symbols in 3D mode.
-const POSSIBLE_SYMBOLS_IN_3D = [
-    'circle',
-    'square',
-    'diamond',
-    'cross',
-    'circle-open',
-    'square-open',
-];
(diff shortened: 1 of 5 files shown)


src/map/options.ts
```diff
@@ -20,11 +20,11 @@ import { COLOR_MAPS } from './colorscales';
...
 import HTML_OPTIONS from './options.html.in';
 
-// in 3D mode, only strings are supported for 'marker.symbol', and only very few
-// of them. See https://github.com/plotly/plotly.js/issues/4205 as the plotly
-// issue tracking more symbols in 3D mode.
-const POSSIBLE_SYMBOLS_IN_3D = ['circle', 'square', 'diamond', 'cross', 'x'];
-
+// in 3D mode, only strings are supported for 'marker.symbol'.
+/* eslint-disable */
+const markers3d = require('./plotly/markers3d');
```
Luthaf, line 25: Why not `import` here?

ceriottm: can't get it to work as an import from both javascript and typescript in any other way. if you know how to do it, change it, I couldn't figure anything else that worked

bernstei:
Sorry I didn't get a chance to check this before you merged, but so far it looks like it fixes the problem for me as well.

PR #384, opened 2024-10-18
Add qualitative palettes for coloring by categorical property

### LINKED ISSUE #381: better support for coloring by a categorical value
As far as I can tell, the only truly categorical visual property that's available is symbol, but I find that it can be hard to find the one circle in a sea of squares, etc.  I would find it very helpful if color could be used for a categorical property.  That works now, but with very little control.  

At a minimum, that could be done by allowing more standard matplotlib palettes, e.g. `tab10`, which currently lead to an only partially helpful error message (in particular the "are you sure" part):
```
invalid property 'tab10' for palette, are you sure the settings correspond to the current dataset?
    at HTMLOption.validate (http://localhost:8080/chemiscope-app.min.js:212328:15)
    at HTMLOption._update (http://localhost:8080/chemiscope-app.min.js:212481:14)
    at set value (http://localhost:8080/chemiscope-app.min.js:212362:14)
    at http://localhost:8080/chemiscope-app.min.js:212585:30
    at foreachOptionImpl (http://localhost:8080/chemiscope-app.min.js:212648:13)
    at MapOptions.foreachOption (http://localhost:8080/chemiscope-app.min.js:212622:9)
    at MapOptions.applySettings (http://localhost:8080/chemiscope-app.min.js:212564:14)
    at MapOptions.applySettings (http://localhost:8080/chemiscope-app.min.js:211512:15)
    at new MapOptions (http://localhost:8080/chemiscope-app.min.js:211484:14)
    at new PropertiesMap (http://localhost:8080/chemiscope-app.min.js:210020:25)
```
Additional control, e.g. mapping specific categorical values to numerical values in the scale, would be even better, but I think just letting me set the min and max of the mapping would be enough.

bernstei:
Add tab10, tab20, tab20b, and tab20c palettes from matplotlib (https://matplotlib.org/stable/users/explain/colors/colormaps.html#qualitative)

closes #381


Luthaf (approved):
Thanks @bernstei !

PR #385, opened 2024-10-28
Revamped chemiscope icon

ceriottm:
VERY important PR of course. A "nicer" icon for chemiscope so we can also make stickers ^^

Old:
chemiscope-icon

New:
chemiscope-icon


Luthaf (commented):
I like the change overall, here are a couple of suggestions:

- I find the telescope a bit too big IMO, meaning it is pushing points in the axes, making the 32px version a bit noisy and not well defined
- Should we have a version with the project name inside as well? We are currently using one in https://docs.metatensor.org/latest/atomistic/engines/index.html

src/static/chemiscope-icon.svg
Luthaf, line 1: Just to be sure, is this using a custom font for the text? If so, did you convert the test to curves instead of using a font, to make sure the rendering is consistent on all systems?

I'm asking because the text looks a bit weird on github, and off-center.

ceriottm: good catch. I was using Helvetica Neue, which is also the font used in the website. it looked fine on my system as I have the same font substitution used in inkscape. converted to path so at least it is what it is.

Luthaf (approved):
Good to go for me! Should we try to include the new icon in the docs/readme/... ?

ceriottm:
> Good to go for me! Should we try to include the new icon in the docs/readme/... ?

I'd say that's for another PR working on restyling the docs.

PR #386, opened 2024-11-03
General clean-up, especially around docs, in preparation for v1.0

ceriottm:
Mildly customized furo theme looks much better IMO


ceriottm:
Fixed a couple of papercut bugs, and revamped the docs appearance. @Luthaf - IDK what you had in mind in terms of docs restructuring, overall it doesn't look bad to me as it is now.

ceriottm:
OK I tried also to look into the symbol bugs that @bernstei  pointed at, but it looks like it's not a trivial fix so probably deserving of a separate PR. Time to merge this before it becomes an everything-pr!

sofiia-chorna:
Thanks a lot for the clean up and update of the doc style! :)

I pushed a [commit](https://github.com/lab-cosmo/chemiscope/pull/386/commits/c3926dc5ef55aad11d14de6c2f13da93fcbbabf3) to fix a warning `ignored unknown settings '{"size":{"reverse":true}}'` that started to appear in some examples, like Dielectric Response.

ceriottm:
> Thanks a lot for the clean up and update of the doc style! :)
> 
> I pushed a [commit](https://github.com/lab-cosmo/chemiscope/pull/386/commits/c3926dc5ef55aad11d14de6c2f13da93fcbbabf3) to fix a warning `ignored unknown settings '{"size":{"reverse":true}}'` that started to appear in some examples, like Dielectric Response.

Thanks - we actually removed this setting in a previous PR as it was pretty useless and complicating a lot the logic around size scaling. This error should actually be also already fixed as I modified the example .json - if you still get the error locally you probably need to clear your examples cache so they get downloaded again.

src/map/options.ts
```diff
@@ -79,6 +79,7 @@ export class MapOptions extends OptionsGroup {
         factor: HTMLOption<'number'>;
         mode: HTMLOption<'string'>;
         property: HTMLOption<'string'>;
+        reverse: HTMLOption<'boolean'>;
```
ceriottm, line 82: Please revert his commit, I think this should not be necessary.

docs/_static/css/chemiscope.css
```diff
@@ -14,3 +14,11 @@ pre {
 .admonition.note a:hover {
     color: #18bc9c;
 }
+
+.sidebar-logo {
+    display: block;
+    margin: 0 auto;
+    max-width: 100%;
+    clip-path: inset(9px 9px 25px 9px);
```
Luthaf, line 22: It this to remove the dark blue border around the logo? I find the result a bit jarring:

Screenshot

Maybe using a lighter color of blue for the background would blend it better? I can try to play with the colors

src/options.ts
```diff
@@ -374,10 +373,14 @@ export abstract class OptionsGroup {
             const lastKey = keys[keys.length - 1];
 
             if (lastKey in root) {
-                const value = root[lastKey];
+                var value = root[lastKey];
```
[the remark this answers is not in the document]
Luthaf: It was disabled for this part of the code, I re-enabled it

app/index.html
```diff
@@ -377,8 +398,12 @@ <h5 style="margin-top: 1em">Visualization state</h5>
...
                     await app.fetchAndLoad(externalDataUrl(external), config);
                 } else {
-                    // default to Arginine-Dipeptide
-                    await app.loadExample('Arginine-Dipeptide');
+                    // loads a random example
+                    var exampleKeys = Object.keys(examples);
+                    var randomIndex = Math.floor(Math.random() * exampleKeys.length);
+                    var randomExampleKey = exampleKeys[randomIndex];
+
+                    await app.loadExample(randomExampleKey);
```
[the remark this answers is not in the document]
ceriottm: True, but `chemiscope.min.js` is 3MB already so even in the worst case it will make loading 2x slower.... 
Maybe we can test a bit and see if it's a problem, and in case I think the "right thing" would be to reduce a bit the size of the demo datasets.

docs/_static/css/chemiscope.css
ceriottm: so the idea is to have a single svg for the logo-cum-name and the square with just the figure. feel free to play with colors of the website, the blue border in the full logo uses the same hue in the website, so I'd keep it.

docs/requirements.txt
```diff
@@ -1,7 +1,8 @@
 sphinx>=4
-sphinx-bootstrap-theme
+sphinx_rtd_theme   # sphinx theme
```
ceriottm, line 2:
```suggestion
```
remove

docs/_static/css/chemiscope.css
Luthaf: Ok, I'll keep the same hue but move around in saturation/lightness

Luthaf: I could not get the right amount of white padding around the "graph axes", so I added a separate file for it. The hack was nice, but two SVG is not what will kill this repo ^_^

Luthaf (approved):

Issue #127, opened 2021-04-15
Add ribbon representation in structure viewer


Luthaf:
This is a popular representation for bio-molecules, in particular proteins, supported by both JSmol and 3Dmol.js. Two separate people asked for us to support this!

This kind of representation requires some information about the secondary structure, which we would need to add to the input file. I'm not sure which exact information is required to construct the ribbon, I would guess the list and order of alpha carbons in amino-acids? If you have more information on the subject, please let us know!

ceriottm:
would make sense to only do this after we converge on jsmol vs 3dmol.js
though

Luthaf:
Agreed! I'm opening the issue to keep track of it.

With a bit more research, it looks like we need to have secondary structure assignment for atoms (i.e. is this atom part of an helix, a beta strand, etc.). There is some code to guess the secondary structure in 3Dmol: https://github.com/3dmol/3Dmol.js/blob/17aa414914b0898e4089436524c89cf875505307/3Dmol/parsers.js#L310. Alternatively, we can put this data inside the input JSON, and guess/compute it when generating the JSON file.

ceriottm:
Fixed in #442

PR #395, opened 2024-11-18
Support `MDAnalysis.Universe` as the `frame` parameter

### LINKED ISSUE #358: Allow passing an MDAnalysis AtomGroup as `frames` paramater
A direct passing of an MDAnalysis `AtomGroup` would be very helpful for people from the Bio molecular dynamics field.

As @Luthaf already mentioned one should implement a file similar to

https://github.com/lab-cosmo/chemiscope/blob/main/python/chemiscope/structures/_ase.py

A very useful additional parameter to be added, is an option to `slice` the frames to be shown. Since in MDAnalysis, the whole trajectory information is contained in a single object, it will require this additional parameter to allow users to select which frames they want to show. I think, this might be integrated globally for all supported frames object to maintain consistency

After a possible integration we can also promote chemiscope on the MDAnalysis [website](https://www.mdanalysis.org/pages/mdakits)  :-)

GardevoirX:
Closes #358

This PR allows passing a `Universe` object to the `frame` parameter of `chemiscope.show`. It can be checked by running
```python
import chemiscope
import MDAnalysis as mda

u = mda.Universe("./python/examples/data/trajectory.xyz")
chemiscope.show(frames=u, mode="structure")
```

As mda is usually used in analyzing biosystem-related trajectories, I also tried it on a pdb file containing two peptides, but I found the current representation schemes cannot display the protein well (see below), maybe we can later consider supporting the ribbon diagram representation?
image

### DIFF
--- a/python/chemiscope/jupyter.py
+++ b/python/chemiscope/jupyter.py
@@ -246,7 +246,11 @@ def show(
             # also adds an index property to have something to show in the info panel
             properties["index"] = {
                 "target": "structure",
-                "values": list(range(len(frames))),
+                "values": (
+                    list(range(len(frames)))
+                    if hasattr(frames, "__len__")
+                    else list(range(len(frames.trajectory)))
+                ),
             }
 
         widget_class = StructureWidget
--- a/python/chemiscope/structures/__init__.py
+++ b/python/chemiscope/structures/__init__.py
@@ -18,6 +18,12 @@ from ._ase import (  # noqa: F401
     ase_tensors_to_ellipsoids,
     ase_vectors_to_arrows,
 )
+from ._mda import (
+    _mda_to_json,
+    _mda_valid_structures,
+    _mda_list_atom_properties,
+    _mda_list_structure_properties,
+)
 from ._stk import (  # noqa: F401
     _stk_valid_structures,
     _stk_to_json,
@@ -41,6 +47,10 @@ def _guess_adapter(frames):
     stk_frames, use_stk = _stk_valid_structures(frames)
     if use_stk:
         return stk_frames, "stk"
+    
+    mda_frames, use_mda = _mda_valid_structures(frames)
+    if use_mda:
+        return mda_frames, "mda"
 
     raise Exception(f"unknown frame type: '{frames[0].__class__.__name__}'")
 
@@ -60,6 +70,8 @@ def frames_to_json(frames):
         return [_ase_to_json(frame) for frame in frames]
     elif adapter == "stk":
         return [_stk_to_json(frame) for frame in frames]
+    elif adapter == "mda":
+        return [_mda_to_json(frames.atoms) for frame in frames.trajectory]
     else:
         raise Exception("reached unreachable code")
 
@@ -76,7 +88,8 @@ def _list_atom_properties(frames):
         return _ase_list_atom_properties(frames)
     elif adapter == "stk":
         return _stk_list_atom_properties(frames)
-
+    elif adapter == "mda":
+        return _mda_list_atom_properties(frames)
     else:
         raise Exception("reached unreachable code")
 
@@ -93,6 +106,8 @@ def _list_structure_properties(frames):
         return _ase_list_structure_properties(frames)
     elif adapter == "stk":
         return _stk_list_structure_properties(frames)
+    elif adapter == "mda":
+        return _mda_list_atom_properties(frames)
     else:
         raise Exception("reached unreachable code")
 
@@ -117,6 +132,11 @@ def extract_properties(frames, only=None, environments=None):
     elif adapter == "stk":
         raise RuntimeError(
             "stk molecules do not contain properties, you must manually provide them"
+        )
+    
+    elif adapter == "mda":
+        raise RuntimeError(
+            "MDAnalysis molecules do not contain properties, you must manually provide them"
         )
 
     else:
--- a/python/chemiscope/structures/_mda.py
+++ b/python/chemiscope/structures/_mda.py
@@ -1,0 +1,43 @@
+import numpy as np
+
+try:
+    import MDAnalysis as mda
+
+    HAVE_MDA = True
+except ImportError:
+    HAVE_MDA = False
+
+
+def _mda_valid_structures(frames: mda.Universe):
+    if HAVE_MDA and isinstance(frames, mda.Universe):
+        return frames, True
+    else:
+        return [], False
+
+
+def _mda_to_json(ag):
+    data = {}
+    data["size"] = len(ag)
+    data["names"] = [atom.name for atom in ag]
+    data["x"] = [float(value) for value in ag.positions[:, 0]]
+    data["y"] = [float(value) for value in ag.positions[:, 1]]
+    data["z"] = [float(value) for value in ag.positions[:, 2]]
+    if ag.dimensions is not None:
+        data["cell"] = list(
+            np.concatenate(
+                mda.lib.mdamath.triclinic_vectors(ag.dimensions), dtype=np.float64
+                # should be np.float64 otherwise not serializable
+            )
+        )
+
+    return data
+
+
+def _mda_list_atom_properties(frames) -> list:
+    # mda cannot have atom properties or structure properties, so skipping.
+    return []
+
+
+def _mda_list_structure_properties(frames) -> list:
+    # mda cannot have atom properties or structure properties, so skipping.
+    return []
--- a/python/chemiscope/structures/_stk.py
+++ b/python/chemiscope/structures/_stk.py
@@ -18,7 +18,7 @@ def _stk_valid_structures(
     if HAVE_STK and isinstance(frames, Molecule):
         # deal with the user passing a single frame
         return [frames], True
-    elif HAVE_STK and isinstance(frames[0], Molecule):
+    elif HAVE_STK and hasattr(frames, "__iter__") and isinstance(frames[0], Molecule):
         for frame in frames:
             assert isinstance(frame, Molecule)
         return frames, True
--- a/python/tests/mda_structures.py
+++ b/python/tests/mda_structures.py
@@ -1,0 +1,61 @@
+import unittest
+
+import MDAnalysis as mda
+import numpy as np
+
+import chemiscope
+
+BASE_FRAME = mda.Universe.empty(n_atoms=3, trajectory=True)
+BASE_FRAME.add_TopologyAttr("name", ["C", "O", "O"])
+BASE_FRAME.atoms.positions = np.array([[0, 0, 0], [1, 1, 1], [2, 2, 5]])
+
+
+
+class TestStructures(unittest.TestCase):
+    """Conversion of structure data to chemiscope JSON"""
+
+    def test_structures(self):
+        data = chemiscope.create_input(BASE_FRAME)
+        self.assertEqual(len(data["structures"]), 1)
+        self.assertEqual(data["structures"][0]["size"], 3)
+        self.assertEqual(data["structures"][0]["names"], ["C", "O", "O"])
+        self.assertEqual(data["structures"][0]["x"], [0, 1, 2])
+        self.assertEqual(data["structures"][0]["y"], [0, 1, 2])
+        self.assertEqual(data["structures"][0]["z"], [0, 1, 5])
+        self.assertEqual(data["structures"][0].get("cell"), None)
+
+        frame = BASE_FRAME.copy()
+        frame.dimensions = [23, 22, 11, 90, 90, 90]
+        data = chemiscope.create_input(frame)
+        self.assertEqual(len(data["structures"]), 1)
+        self.assertEqual(data["structures"][0]["cell"], [23, 0, 0, 0, 22, 0, 0, 0, 11])
+
+        frame = BASE_FRAME.copy()
+        frame.dimensions = [23, 22, 11, 120, 90, 70]
+        data = chemiscope.create_input(frame)
+        self.assertEqual(len(data["structures"]), 1)
+
+        cell = [
+            23.0,
+            0.0,
+            0.0,
+            7.5244431531647145,
+            20.673237657289985,
+            0.0,
+            0.0,
+            -5.852977748617515,
+            9.313573507209156,
+        ]
+        self.assertTrue(np.allclose(data["structures"][0]["cell"], cell))
+
+
+class TestExtractProperties(unittest.TestCase):
+    """Properties extraction"""
+
+    def test_exception(self):
+        with self.assertRaises(RuntimeError):
+            chemiscope.extract_properties(BASE_FRAME)
+
+
+if __name__ == "__main__":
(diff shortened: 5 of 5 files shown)


Luthaf:
Thanks a lot, this is very nice!


From the screenshot, it looks like the structure is using the wrong unit. Chemiscope/3Dmol.js assumes Angstroms, but I think MDA is using nm. I think there should be at least some automatically guessed bonds in this structure.

As a related point, does MDA add back the implicit H atoms that are often missing in bioinformatics file formats?


> maybe we can later consider supporting the ribbon diagram representation?

This is #127, PR welcome! The biggest issue there will be that we will need to add another field to chemiscope's JSON containing information about alpha carbons / some level of pre-computed secondary structure information.

python/chemiscope/structures/__init__.py
```diff
@@ -60,6 +70,8 @@ def frames_to_json(frames):
         return [_ase_to_json(frame) for frame in frames]
     elif adapter == "stk":
         return [_stk_to_json(frame) for frame in frames]
+    elif adapter == "mda":
+        return [_mda_to_json(frames.atoms) for frame in frames.trajectory]
```
Luthaf, line 74:
```suggestion
        return [_mda_to_json(step.atoms) for step in frames.trajectory]
```

python/chemiscope/jupyter.py
```diff
@@ -246,7 +246,11 @@ def show(
             # also adds an index property to have something to show in the info panel
             properties["index"] = {
                 "target": "structure",
-                "values": list(range(len(frames))),
+                "values": (
+                    list(range(len(frames)))
+                    if hasattr(frames, "__len__")
+                    else list(range(len(frames.trajectory)))
```
Luthaf, line 252: I don't love adding special cases everywhere in the code =/ Could it make sense to change the API such that the user would call `chemiscope.show(u.trajectory)` instead of `chemiscope.show(u)`? It does not seems that the code needs the other data from a `Universe`.

python/chemiscope/structures/_mda.py
```diff
@@ -0,0 +1,44 @@
...
+    if HAVE_MDA and isinstance(frames, mda.Universe):
+        return frames, True
+    else:
+        return [], False
+
+
+def _mda_to_json(ag):
+    data = {}
+    data["size"] = len(ag)
+    data["names"] = [atom.name for atom in ag]
```
Luthaf, line 21: Is this the atom name or element? i.e. `CA` or `C` for alpha carbons? From chemiscope perspective, this should be the element name rather than the atom type.

python/chemiscope/structures/_stk.py
```diff
@@ -18,7 +18,7 @@ def _stk_valid_structures(
     if HAVE_STK and isinstance(frames, Molecule):
         # deal with the user passing a single frame
         return [frames], True
-    elif HAVE_STK and isinstance(frames[0], Molecule):
+    elif HAVE_STK and hasattr(frames, "__iter__") and isinstance(frames[0], Molecule):
```
Luthaf, line 21: Why this check? The following `isinstance(frames[0], Molecule)` should already return early when using non stk data.

python/chemiscope/jupyter.py
GardevoirX: No, I tried to pass the `u.trajectory`, but it does not contain the atom name information, which seems to be only stored in the `Universe`.

python/chemiscope/structures/_stk.py
GardevoirX: Mainly because `Universe` is not an iterable. Without this check, a `TypeError` will be thrown here if we passes an `Universe`.

python/chemiscope/structures/_mda.py
GardevoirX: Here this is the atom name, `CA` for alpha carbons. Yes I think the element name makes more sense.

python/chemiscope/jupyter.py
GardevoirX: Just had some more look into it. The information is also stored in `u.atoms`, and it has all the information that we need, but it follows a lazy loading way. Even though with the expression `[u.atoms.copy() for _ in u.trajectory]`, every `AtomGroup` object in the list still points to the first frame of the trajectory. If we can solve this problem, we can pass `list[AtomGroup]` instead of passing a `Universe`.

python/chemiscope/structures/_stk.py
Luthaf: Is it also not indexable? i.e. `universe[0]` fails? I would rather check for this by doing 
```suggestion
    elif HAVE_STK and isinstance(frames, list) and isinstance(frames[0], Molecule):
```

GardevoirX: Yes not indexable

GardevoirX:
Regarding the unit issue, I think mda read the coordinate in the structure file as-is. The real difference comes from the units that are used by different formats (like Å for xyz and nm for pdb, generally speaking). And it might be hard to tell which unit is being using from chemiscope side.

Luthaf:
https://userguide.mdanalysis.org/stable/units.html Looks like everything should be in A. I'm not sure what's happening in the initial screenshot though. Which file did you use for it?

GardevoirX:
Ahh yes, thanks! I used [this file](https://github.com/wengjingwei35/p53-MDM2/blob/main/sampled_structures/RMSD0.37Q%20score0.54_sample/complex-md-Frame006795.pdb). And in the initial screenshot, I toggled off the bond. Perhaps you can take the latter one for your reference.

PicoCentauri:
What is the status here? Would be still nice to have this feature if possible!

PicoCentauri:
Once this is merged we can add chemiscope here:

https://www.mdanalysis.org/pages/mdakits/#visualization-tools

PicoCentauri (commented):
Nice. Instead of passing a `universe` I would pass an `Atomgroup`, which can be accessed by `universe.atoms`. In your code you can get the trajectory by `atoms.universe.trajectory`.

If we allow passing an atomgroup users can easily subselect atoms.

PicoCentauri (commented):
Nice. The tests are not happy though.

GardevoirX:
> Nice. The tests are not happy though.

Now good👍

python/tests/mda_structures.py
```diff
@@ -0,0 +1,62 @@
...
+BASE_FRAME = mda.Universe.empty(n_atoms=3, trajectory=True)
+BASE_FRAME.add_TopologyAttr("type", ["C", "O", "O"])
+BASE_FRAME.atoms.positions = np.array([[0, 0, 0], [1, 1, 1], [2, 2, 5]])
+
+
+class TestStructures(unittest.TestCase):
+    """Conversion of structure data to chemiscope JSON"""
+
+    def test_structures(self):
+        print(BASE_FRAME.atoms)
```
PicoCentauri:
```suggestion
```

PicoCentauri (commented):
Looks good to me. But maybe also we should add some info somewhere in the docs. Probably @Luthaf can show the correct spot.

GardevoirX:
> Looks good to me. But maybe also we should add some info somewhere in the docs. Probably @Luthaf can show the correct spot.

Maybe we can add something like this [example](chemiscope.org/docs/examples/9-showing-custom-bonds.html)? but if so, I'd like to merge this PR first and add an example in the following PR about introducing cartoon

python/chemiscope/structures/__init__.py
```diff
@@ -116,6 +133,12 @@ def extract_properties(frames, only=None, environments=None):
             "stk molecules do not contain properties, you must manually provide them"
         )
 
+    elif adapter == "mda":
+        raise RuntimeError(
+            "MDAnalysis molecules do not contain properties, you"
```
Luthaf:
```suggestion
            "MDAnalysis structures do not contain properties, you"
```

python/chemiscope/structures/_mda.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def _mda_get_atom_properties(frames) -> list:
+    # mda cannot have atom properties or structure properties, so skipping.
+    return []
+
+
+def _mda_list_structure_properties(frames) -> list:
+    # mda cannot have atom properties or structure properties, so skipping.
+    return []
```
Luthaf: Is this true? Can we not get the forces/energies out of an MDA atom group?

python/chemiscope/jupyter.py
```diff
@@ -271,7 +271,13 @@ def show(
...
             properties["index"] = {
                 "target": "structure",
-                "values": list(range(len(frames))),
+                "values": (
+                    list(range(len(frames)))
+                    if hasattr(frames, "__len__")
+                    # MDAnalysis `frames` will not be a list, and does not have
+                    # a `__len__`
+                    else list(range(len(frames.trajectory)))
+                ),
```
Luthaf: This part of the code should not have to care about the exact structure provider. We should fine another way to express this, maybe creating the input first and only then adding this property?

Luthaf:
Adding an example later for docs sounds good, although we could update the docstring for `create_input`, `write_input`, `show` (and anything else taking `frames`) to mention both stk and mda (they currently say "For now, only ase.Atoms objects are supported").

python/chemiscope/structures/_mda.py
GardevoirX: Seems like it's only possible when we have an `.edr` file from `gmx`
https://www.mdanalysis.org/2022/11/20/bjarne_gsoc/

python/chemiscope/jupyter.py
GardevoirX: I think we can put this here? The code below basically extracts the same thing as this part
https://github.com/lab-cosmo/chemiscope/blob/991bc4bcf89f8696b60d354c4974d3167035eb49/python/chemiscope/input.py#L279-L282

GardevoirX: but we also have to let `create_input` know which mode we are in, if we choose this way

GardevoirX:
> Adding an example later for docs sounds good, although we could update the docstring for `create_input`, `write_input`, `show` (and anything else taking `frames`) to mention both stk and mda (they currently say "For now, only ase.Atoms objects are supported").

I think that's good for finishing this PR. Adding an example can be postponed, but yes it's nice to show others what we can do through chemiscope

python/chemiscope/structures/_mda.py
PicoCentauri: There is no standard field for energies but for forces there are a [forces](https://docs.mdanalysis.org/stable/documentation_pages/core/groups.html#MDAnalysis.core.groups.AtomGroup.forces), and many more like, velocities, dipole_moments etc.

GardevoirX: Is there an example? I tried to load a `extxyz` dataset which has a forces field, but mda complained
```python
>>> u.atoms.forces
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/qxu/Programs/miniconda3/envs/po4/lib/python3.12/site-packages/MDAnalysis/core/groups.py", line 2780, in __getattr__
    raise NoDataError("This Timestep has no " + attr)
MDAnalysis.exceptions.NoDataError: This Timestep has no forces
```

PicoCentauri: Might be that the xyz reader in mda can't read it. Xyz is the opposite of a well defined file format. 😱

 Maybe via the chemfiles interface? I can try tomrrow.

GardevoirX: Thanks!

python/chemiscope/jupyter.py
Luthaf: Why would we have to do anything in `create_input`?

GardevoirX: Because that's a part of the input, I think moving it into `create_input` seems to be nicer

GardevoirX: Now it's updated without touching `create_input`

python/chemiscope/input.py
```diff
@@ -61,16 +61,10 @@ def create_input(
...
-            "description": "...",
-            # list of str, dataset authors, OPTIONAL
-            "authors": [
+            # str, dataset name "name": "...", # str, dataset description "description":
+            "...", # list of str, dataset authors, OPTIONAL "authors": [
                 "...",
-            ],
-            # list of str, references for this dataset, OPTIONAL
-            "references": [
+            ], # list of str, references for this dataset, OPTIONAL "references": [
```
Luthaf: Looks like these changes are unrelated, could you undo them?

python/chemiscope/jupyter.py
```diff
@@ -303,6 +291,12 @@ def show(
         settings=settings,
     )
 
+    if (mode == "structure") and ("index" not in dict_input["properties"]):
```
Luthaf:
```suggestion
    if mode == "structure" and "index" not in dict_input["properties"]:
```

python/chemiscope/structures/__init__.py
```diff
@@ -111,9 +129,12 @@ def extract_properties(frames, only=None, environments=None):
     if adapter == "ASE":
         return _ase_extract_properties(frames, only, environments)
 
+    elif adapter == "mda":
+        return _mda_extract_properties(frames, only, environments)
+
     elif adapter == "stk":
         raise RuntimeError(
-            "stk molecules do not contain properties, you must manually provide them"
+            "stk structures do not contain properties, youmust manually provide them"
```
Luthaf:
```suggestion
            "stk structures do not contain properties, you must manually provide them"
```

python/chemiscope/structures/_mda.py
```diff
@@ -0,0 +1,134 @@
...
+        for _ in ag.universe.trajectory[1:]:
+            for property in all_properties:
+                try:
+                    all_properties[property].append([getattr(ag, property)])
+                except mda.exceptions.NoDataError:
+                    extra.add(property)
+
+    if len(extra) != 0:
+        warnings.warn(
+            "the following structure properties are only defined for a subset "
```
Luthaf:
```suggestion
            "the following per-atom properties are only defined for a subset "
```

python/chemiscope/structures/_mda.py
```diff
@@ -0,0 +1,134 @@
...
+                dtype=np.float64,
+                # should be np.float64 otherwise not serializable
+            )
+        )
+
+    return data
+
+
+def _mda_get_atom_properties(ag) -> dict:
+    POSSIBLE_PROPERTIES = ["velocities", "forces"]
```
Luthaf, line 45: @PicoCentauri is the set of possible properties defined somewhere in MDA doc?

Luthaf (approved):

python/chemiscope/structures/_mda.py
Luthaf: https://docs.mdanalysis.org/dev/documentation_pages/coordinates/timestep.html

PR #405, opened 2025-04-06
Re-haul of warning system, making handlers private to specific viewers instances

ceriottm:
Also adds configurable timeouts

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -3,7 +3,7 @@
 import 'bootstrap';
 import 'bootstrap/dist/css/bootstrap.min.css';
 
-import { getByID, addWarningHandler, Warnings } from '../src/utils';
+import { getByID, Warnings } from '../src/utils';
 import { Dataset, Structure } from '../src/dataset';
 import { version, DefaultVisualizer, Settings } from '../src/index';
 
@@ -25,8 +25,7 @@ export class ChemiscopeApp {
     /// CSS style sheet to hide the setting in loading panel about on-demand
     /// loading
     private hideOnDemandStructures: HTMLStyleElement;
-    private warnings: Warnings;
-    public warningTimeout: number = 4000;
+    public warnings: Warnings = new Warnings();
 
     /**
      * Create a new instance of the chemiscope application.
@@ -35,11 +34,6 @@ export class ChemiscopeApp {
         // show the version of chemiscope currently running
         const versionDisplay = getByID('chemiscope-version');
         versionDisplay.innerText = `version ${version()}`;
-
-        // handle warnings
-        this.warnings = new Warnings();
-        this.warnings.add((message) => displayWarning(message, this.warningTimeout));
-        addWarningHandler((message) => displayWarning(message, this.warningTimeout));
 
         // when the window is resized, change the size available to the info
         // widget
@@ -157,7 +151,12 @@ export class ChemiscopeApp {
             this.visualizer.remove();
         }
 
-        console.log("loading visualizer with warnings", this.warnings);
+        // adds warning handler
+        this.warnings.timeout = 4000; // 4s visibility
+        this.warnings.addHandler((message) => {
+            displayWarning(message, this.warnings.timeout);
+        });
+
         this.visualizer = await DefaultVisualizer.load(config, dataset, this.warnings);
 
         this.visualizer.structure.positionSettingsModal = (rect) => {
@@ -285,8 +284,7 @@ export class ChemiscopeApp {
 }
 
 function displayWarning(message: string, timeout: number = 4000) {
-    console.log('displaying warining', message, 'timeout', timeout);
-    if (timeout > 0) {        
+    if (timeout > 0) {
         const display = getByID('warning-display');
         display.getElementsByTagName('p')[0].innerText = message;
         display.style.display = 'block';
--- a/app/app.ts
+++ b/app/app.ts
@@ -152,9 +152,9 @@ export class ChemiscopeApp {
         }
 
         // adds warning handler
-        this.warnings.timeout = 4000; // 4s visibility
-        this.warnings.addHandler((message) => {
-            displayWarning(message, this.warnings.timeout);
+        this.warnings.defaultTimeout = 4000; // 4s visibility
+        this.warnings.addHandler((message, timeout?) => {
+            displayWarning(message, timeout);
         });
 
         this.visualizer = await DefaultVisualizer.load(config, dataset, this.warnings);
--- a/app/app.ts
+++ b/app/app.ts
@@ -284,12 +284,13 @@ export class ChemiscopeApp {
 }
 
 function displayWarning(message: string, timeout: number = 4000) {
-    if (timeout > 0) {
+    // draws a message box (and closes after the specified timeout)
+    if (timeout > 0) { // use a negative timeout not to print messages
         const display = getByID('warning-display');
         display.getElementsByTagName('p')[0].innerText = message;
         display.style.display = 'block';
 
-        // automatically remove the warning after 4s
+        // automatically remove the warning after the specified timeout
         setTimeout(() => {
             display.style.display = 'none';
         }, timeout);
--- a/app/app.ts
+++ b/app/app.ts
@@ -3,7 +3,7 @@
 import 'bootstrap';
 import 'bootstrap/dist/css/bootstrap.min.css';
 
-import { getByID, addWarningHandler } from '../src/utils';
+import { getByID, addWarningHandler, Warnings } from '../src/utils';
 import { Dataset, Structure } from '../src/dataset';
 import { version, DefaultVisualizer, Settings } from '../src/index';
 
@@ -25,6 +25,8 @@ export class ChemiscopeApp {
     /// CSS style sheet to hide the setting in loading panel about on-demand
     /// loading
     private hideOnDemandStructures: HTMLStyleElement;
+    private warnings: Warnings;
+    public warningTimeout: number = 4000;
 
     /**
      * Create a new instance of the chemiscope application.
@@ -35,7 +37,9 @@ export class ChemiscopeApp {
         versionDisplay.innerText = `version ${version()}`;
 
         // handle warnings
-        addWarningHandler((message) => displayWarning(message));
+        this.warnings = new Warnings();
+        this.warnings.add((message) => displayWarning(message, this.warningTimeout));
+        addWarningHandler((message) => displayWarning(message, this.warningTimeout));
 
         // when the window is resized, change the size available to the info
         // widget
@@ -153,7 +157,8 @@ export class ChemiscopeApp {
             this.visualizer.remove();
         }
 
-        this.visualizer = await DefaultVisualizer.load(config, dataset);
+        console.log("loading visualizer with warnings", this.warnings);
+        this.visualizer = await DefaultVisualizer.load(config, dataset, this.warnings);
 
         this.visualizer.structure.positionSettingsModal = (rect) => {
             const structureRect = getByID('chemiscope-structure').getBoundingClientRect();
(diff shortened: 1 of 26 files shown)


app/app.ts
```diff
@@ -25,6 +25,8 @@ export class ChemiscopeApp {
     /// CSS style sheet to hide the setting in loading panel about on-demand
     /// loading
     private hideOnDemandStructures: HTMLStyleElement;
+    private warnings: Warnings;
+    public warningTimeout: number = 4000;
```
Luthaf, line 29: this should have some documentation

app/app.ts
```diff
@@ -25,6 +25,8 @@ export class ChemiscopeApp {
     /// CSS style sheet to hide the setting in loading panel about on-demand
     /// loading
     private hideOnDemandStructures: HTMLStyleElement;
+    private warnings: Warnings;
```
Luthaf, line 28: Naming-wise, I'd rather call this 
```suggestion
    private warningHandler: WarningHandlers;
```

since this not about the warning themselves, but about what to do with them

app/app.ts
```diff
@@ -153,7 +157,8 @@ export class ChemiscopeApp {
             this.visualizer.remove();
         }
 
-        this.visualizer = await DefaultVisualizer.load(config, dataset);
+        console.log("loading visualizer with warnings", this.warnings);
```
Luthaf, line 160:
```suggestion
```

app/app.ts
```diff
@@ -279,15 +284,18 @@ export class ChemiscopeApp {
...
-    const display = getByID('warning-display');
-    display.getElementsByTagName('p')[0].innerText = message;
-    display.style.display = 'block';
-
-    // automatically remove the warning after 4s
-    setTimeout(() => {
-        display.style.display = 'none';
-    }, 4000);
+function displayWarning(message: string, timeout: number = 4000) {
+    console.log('displaying warining', message, 'timeout', timeout);
```
Luthaf, line 288:
```suggestion
```

app/app.ts
```diff
@@ -279,15 +284,18 @@ export class ChemiscopeApp {
...
+    if (timeout > 0) {
```
Luthaf, line 289: this behavior needs to be documented as well =)

docs/src/python/sphinx.rst
```diff
@@ -48,7 +48,8 @@ The ``structure`` mode will show the structures only
 and the ``map`` mode only the property map
 
 .. chemiscope:: ../datasets/showcase.json.gz
-    :mode: map 
+    :mode: map
+    :warning_timeout: 0.0
```
Luthaf, line 52: should we make this the default in sphinx?

python/chemiscope/jupyter.py
```diff
@@ -28,14 +28,18 @@ class ChemiscopeWidgetBase(ipywidgets.DOMWidget, ipywidgets.ValueWidget):
...
-    def __init__(self, data, has_metadata):
+    def __init__(self, data, has_metadata, warning_timeout=10000):
         super().__init__()
         self.value = json.dumps(data)
         self.has_metadata = has_metadata
         if "settings" in data:
             self.settings = data["settings"]
         self._settings_sync = True
+        self.warning_timeout = warning_timeout
+        print("set timeout to ", self.warning_timeout)
```
Luthaf, line 42:
```suggestion
```

python/chemiscope/jupyter.py
```diff
@@ -115,6 +119,7 @@ def show_input(path, mode="default"):
     :param str path: load the chemiscope widget from path.
 
     :param str mode: widget mode, either ``default``, ``structure`` or ``map``.
+    :param float warning_timeout: timeout (in ms) for warnings
```
Luthaf, line 122: The same doc that exists in `show` should be duplicated everywhere this parameter exists, otherwise people might not understand what this parameter is for

python/chemiscope/sphinx/nodes.py
```diff
@@ -41,14 +45,19 @@ def depart_chemiscope_html(self, node):
 
 
 def generate_html_content(
-    filepath, mode="default", include_headers=True, static_path="_static"
+    filepath,
+    mode="default",
+    warning_timeout=4000,
```
Luthaf, line 50: I think we should pick one default value for the timeout and use it everywhere. It looks like `4000` and `10000` are used in different places

python/jupyter/src/widget.ts
```diff
@@ -92,15 +111,14 @@ export class ChemiscopeView extends ChemiscopeBaseView {
         // and then inserting this.el inside the HTML document.
         const element = this.el;
 
+        this._updateWarningTimeout();
```
Luthaf, line 114: Why do we need to call this everytime? I though the point of traitlets in jupyter was to auto-sync such data.

src/index.ts
```diff
@@ -165,9 +166,9 @@ class DefaultVisualizer {
      * @param  dataset visualizer input, containing a dataset and optional visualization settings
      * @return         Promise that resolves to a {@link DefaultVisualizer}
      */
-    public static load(config: DefaultConfig, dataset: Dataset): Promise<DefaultVisualizer> {
+    public static load(config: DefaultConfig, dataset: Dataset, warnings?: Warnings): Promise<DefaultVisualizer> {
```
Luthaf, line 169: Why does this need to take a `warnings` instance? It could create one, since the instance will then be publicly accessible anyway

src/map/map.ts
```diff
@@ -256,6 +258,7 @@ export class PropertiesMap {
         this.onselect = () => {};
         this.activeChanged = () => {};
         this._selected = new Map<GUID, MarkerData>();
+        this.warnings = new Warnings;
```
Luthaf, line 261: shouldn't this class take a warnings parameter to use the same one as the rest of the visualizer?

src/structure/grid.ts
```diff
@@ -866,7 +872,8 @@ export class ViewersGrid {
             // add a new cells if necessary
             if (!this._cellsData.has(cellGUID)) {
                 const propertiesName = this._properties ? Object.keys(this._properties) : [];
-                const viewer = new MoleculeViewer(this._getById(`gi-${cellGUID}`), propertiesName);
+                console.log("adding viewer ", this.warnings);
```
Luthaf, line 875:
```suggestion
```

src/utils/warnings.ts
```diff
@@ -5,23 +5,60 @@
...
 
 /** Register the `handler` function to be called when a warning is emitted */
-export function addWarningHandler(handler: WarningHandler): void {
-    WARNINGS_HANDLERS.push(handler);
+export function addWarningHandler(handler: WarningHandler, source: WarningSource=null): void {
+    WARNINGS_HANDLERS.push([handler, source]);
 }
+
+/** Remove warning `handler`s matching a given source */
+export function removeWarningHandlers(source: WarningSource): void {
```
Luthaf, line 55: I guess this is leftover from a previous experiment?

src/utils/warnings.ts
```diff
@@ -5,23 +5,60 @@
...
+    }
+
+    /** Sends a warning message though all handlers */
+    public send(message: string) : void{
+        for (const cb of this.handlersList) {
+            console.log(this.handlersList.length, ': using handler ', cb)
+            cb(message);
+        }
+    }
+}
```
Luthaf, line 29: We could keep the name `Warning` for this class if `send` is a member function; but then I'd rename `add` to `addHandler`, and maybe add a `clearHandlers` function?

Luthaf (commented):
I agree with making warning a per-visualizer object, and this looks like a good way to implement it!

src/index.ts
ceriottm: I was thinking of creating this one step "up", in the app or the widget but I agree this is probably the right point to start the chain

docs/src/python/sphinx.rst
ceriottm: After thinking a bit I made the default 2s, because if one plays around with the viewer then it may trigger warnings we want to show

python/chemiscope/sphinx/nodes.py
ceriottm: so the logic here is that in jupyter you might be debugging and so you want messages to stay up longer... maybe overkill but imo is better to tweak persistence to the use case

python/jupyter/src/widget.ts
ceriottm: we need to call it also at initialization to fetch the value from the traitlet

src/map/map.ts
ceriottm: yes was a quick hack

src/utils/warnings.ts
ceriottm: yes now it's all cleaned up

src/utils/warnings.ts
ceriottm: for the moment I just renamed add. Now we've one per visualizer I don't see a use case to remove them

ceriottm:
Ready for another (and final?) round of review.

python/chemiscope/jupyter.py
Luthaf: @ceriottm still relevant

python/chemiscope/sphinx/static/chemiscope-sphinx.js
```diff
@@ -12,12 +12,19 @@ const VISUALISER_MODE = {
...
+    filePath,
+    visualizerMode = VISUALISER_MODE.DEFAULT,
+    timeout = 4000
+) {
     // Display loading
     toggleLoadingVisible(divId, true);
 
+    const warnings = new Chemiscope.Warnings;
+    warnings.timeout = timeout; // defaults to 4s visibility
+    warnings.addHandler((message) => { displayWarning(divId, message, warnings.timeout)});
```
Luthaf, line 26: I don't understand why `warnings.timeout` is needed here, since the timeout is more a property of the warning handler


```suggestion
    warnings.addHandler((message) => { displayWarning(divId, message, timeout)});
```

python/chemiscope/sphinx/static/chemiscope-sphinx.js
```diff
@@ -36,7 +43,7 @@ async function loadChemiscopeSphinx(divId, filePath, visualizerMode = VISUALISER
 
         // Load widget
         const visualiser = getVisualizer(visualizerMode);
-        await visualiser.load(config, dataset);
+        let viz = await visualiser.load(config, dataset, warnings);
```
Luthaf, line 46: why the `let`?

python/jupyter/src/widget.ts
```diff
@@ -18,6 +18,18 @@ const PlausibleTracker = Plausible({
 class ChemiscopeBaseView extends DOMWidgetView {
     protected visualizer?: DefaultVisualizer | StructureVisualizer | MapVisualizer;
     protected guid!: string;
+    protected warnings: Warnings = new Warnings();
+    protected static className: string = 'base-view';
+
+    public render(): void {
+        PlausibleTracker.trackPageview({
+            url: (location.pathname.split('/')[1] || '') + '/' + (this.className as string),
```
Luthaf, line 26: Are you sure this works with overridden **static** data members? I would have use functions instead, but maybe this works as well? It feels a bit strange scoping-wise that parent classes have access to child classes static data

src/structure/grid.ts
```diff
@@ -176,7 +178,8 @@ export class ViewersGrid {
         target: DisplayTarget,
         properties?: { [name: string]: Property },
         environments?: Environment[],
-        maxViewers: number = 9
+        maxViewers: number = 9,
+        warnings?: Warnings
```
Luthaf, line 182: Since `Warnings` is now a parameter to a lot of classes/functions, I would only make it optional for the couple of top-level classes that we expect users to create without an existing warnings instance; and have TypeScript check that we are passing a valid instance everywhere else.

Otherwise I'm afraid we'll forget one callsite some day, and warnings will go in limbo.

src/utils/warnings.ts
```diff
@@ -6,22 +6,30 @@
 /** A {@link WarningHandler} is called when a new warning is emitted */
 export type WarningHandler = (message: string) => void;
```
Luthaf, line 7: I think this should take the timeout as well, making the Warnings class the only one dealing with timeout (and then handlers can decide to ignore it)

```suggestion
export type WarningHandler = (message: string, timeout: number) => void;
```

python/chemiscope/sphinx/static/chemiscope-sphinx.js
ceriottm: I renamed this as `defaultTimeout` so it's clearer - one can always override from showMessage, but it's otherwise set there.

python/chemiscope/sphinx/static/chemiscope-sphinx.js
ceriottm: well-spotted, leftover from console.logs I think

src/structure/grid.ts
ceriottm: I made it mandatory in all the places that didn't have a bunch of optional parameters already, as I think it should come at the end.

src/utils/warnings.ts
ceriottm: done

python/jupyter/src/widget.ts
ceriottm: well spotted.

Issue #407, opened 2025-04-28
Visualization as jupyter widget with dict input


awvwgk:
Currently for using the Jupyter widget via `chemiscope.show` requires an ASE Atoms object, in case those are not available the only alternative seems `chemiscope.show_input` which does only take a filepath. This makes it somewhat inconvenient to create the Jupyter widget when ASE is not used, because the chemiscope input needs to be serialized and written to disk before.

*Example*: https://tblite--257.org.readthedocs.build/en/257/tutorial/python/geometry.html#visualization

It would be great to either:
* pass a file descriptor or input stream to `chemiscope.show_input`
* pass a dict in chemiscope input schema to `chemiscope.show`

ceriottm:
Fixed in #410

PR #410, opened 2025-04-28
Allow file-like objects as input to chemiscope.show_input

awvwgk:
This is useful in combination with libraries like [fsspec](https://filesystem-spec.readthedocs.io/) which allow to fetch data files from remote resources.

See https://github.com/lab-cosmo/chemiscope/issues/407

### DIFF
--- a/python/chemiscope/jupyter.py
+++ b/python/chemiscope/jupyter.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 import gzip
-import io
 import json
 import warnings
+from pathlib import Path
 
 import ipywidgets
 from traitlets import Bool, Dict, Int, Unicode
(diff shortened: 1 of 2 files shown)


python/chemiscope/jupyter.py
```diff
@@ -146,7 +153,9 @@ def show_input(path, mode="default", warning_timeout=10000):
     elif mode == "map":
         widget_class = MapWidget
 
-    if path.endswith(".gz"):
+    if isinstance(path, io.IOBase):
```
Luthaf, line 156: I'm not sure this is the best way to check, aren't there file-like objects that don't inherit from IOBase? I would rather do the check the other way around:

```py
if isinstance(file, (str, Path)):
    # current code
else:
    # assume file-like object
```

Luthaf:
Thanks a lot for the contribution!

python/chemiscope/jupyter.py
awvwgk: Changed this to check for `str` or `Path` first.

Luthaf (approved):

PR #411, opened 2025-04-28
Disable OffscreenCanvas rendering in 3DMol.js

### LINKED ISSUE #409: Omitting missing properties
How can one handle missing property values with chemiscope? For example I have a data set with only 95% values for one property and I would still like to display it, mixing float values with None is not allowed due to a type error and using NaNs seemingly crashed the chemiscope viewer.

Luthaf:
Fixes #409 for me.

@awvwgk, if you have the opportunity to check this PR please let me know if it fixes the issue for you as well. If you don't know how to use code from a PR let me know and I'll send you a pre-built Python wheel.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,6 +77,6 @@ explore = [
 ]
 
 metatensor = [
-    "metatensor-operations >=0.3,<0.4",
-    "metatensor-torch >=0.6,<0.7"
+    "metatensor-operations >=0.3.3,<0.4",
+    "metatensor-torch >=0.7.6,<0.8"
 ]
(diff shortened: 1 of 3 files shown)


awvwgk:
I used the wheel produced by the CI, but the issue with displaying data with NaNs is still persistent for me.

Luthaf:
Ok, thanks for checking! What's the error message you get in this case? I assume you are using `show_input` in jupyter?

ceriottm:
Also, @awvwgk , if you're running this in jupyter make sure you restart the server and refresh the browser window, it has a tendency to stick to the cached chemiscope.min.js after an update.

ceriottm:
OK I found the problem preventing loading a JSON with NaNs from chemiscope.show_input. Should be fixed now.

ceriottm (approved):
LGTM and fixes both bugs

ceriottm:
@awvwgk  let us know if this fixes it all for you too, so we can merge.

PR #413, opened 2025-04-30
Allow standalone HTML viewer to load dataset with NaN

### LINKED ISSUE #412: Omitting missing property in standalone mode
Hello,

Similarly to #409 , I have a dataset with NaN values for certain properties and for certain entries and this caused an issue with the offline visualizer. Based on #411 , I fixed it by adding the function ```parseJsonWithNaN``` to ```./utils/index.html``` (see the commit [here](/VicTrqt/chemiscope/commit/b9cd63bf3963102dcddeb108c31d2dd18fbed6a6) ).

Just wanted to flag this in case you want to fix it in a cleaner way than my quick fix.

Thanks for this nice tool!

Luthaf:
Fix #412

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -502,6 +502,16 @@
         // `onclick` above
         let CHEMISCOPE_APP;
 
+        /**
+         * Allow NaN in the JSON file. They are not part of the spec, but Python's json
+         * module output them, and they can be useful.
+         */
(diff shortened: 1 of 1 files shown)


ceriottm (approved):
Maybe one day we could provide this from the chemiscope package, given that we use it in three different apps, but for the moment LGTM.

PR #414, opened 2025-05-12
Implement a `metatomic_featurizer` to replace the metatensor one

Luthaf:
For now, both stay in the code to no disturb users until everything is ready on the metatomic side.


sofiia-chorna (approved):
Looks great ! Thank you for also fixing rst strings

PR #415, opened 2025-05-12
Use ruff for linting and formatting Python

Luthaf:


python/examples/9-showing_custom_bonds.py
```diff
@@ -65,7 +66,7 @@
                     position_matrix=[[0.0, 0.0, 0.0]],
                 ),
                 stk.BuildingBlock(
-                    smiles=("C1=NC=CC(C2=CC=CC(C3=C" "C=NC=C3)=C2)=C1"),
+                    smiles=("C1=NC=CC(C2=CC=CC(C3=CC=NC=C3)=C2)=C1"),
```
Luthaf, line 69: @andrewtarzia does this looks right? The code was already interpreted this way, but I'm wondering if this should have a comma instead, i.e. `smiles=("C1=NC=CC(C2=CC=CC(C3=C", "C=NC=C3)=C2)=C1")`

andrewtarzia: Hey! Yeah I think that's correct. Adding a comma would change the meaning.

ceriottm (approved):
Looks good and it's crazily fast.

Issue #361, opened 2024-09-01
Allow control over automatic bond assignment


andrewtarzia:
Hey all, This is working great so far. Except that my models are coarse-grained, so the bonds that come out from automatic assignment are not correct.

I have the topology information, so I can provide bonds (ideal solution), or I was think I could generate cylinder shapes for each bond. An alternative is that I could scale, or modify, the radii constrain for determining bonding information.

Any suggestions on which approach is best?

ceriottm:
If you want to implement the machinery to provide bond information we'd be open to include it (though I'd suggest we first have a bit of discussion on how to do it so that it doesn't make other use cases more complicated or slow). We most definitely do not have the bandwidth to actively help though.

Otherwise, if you want to do this with existing infrastructure, manually defining the bonds as shapes is probably the easiest way, look at the implementation of `ase_merge_pi_frames`.

andrewtarzia:
I would be very happy to contribute this feature in the future. I am not familiar with typescript, but it does seem like that would be quite a major overhaul of `dataset.ts` and `viewer.ts` to do this without breaking anything.

So, I will work with the `ase_merge_pi_frames` code for now, which seems to do exactly what I want with the shapes.

Thank you!

ceriottm:
If I may make a suggestion - if you make a `topology_to_shapes` utility function that is fairly general and based on standard topo formats, this would be a very much welcome contribution. I'm leaving this open in case you go this way.

Luthaf:
This would be quite nice to have. If someone wants to take a shot at it, here is what would need to happen for a full, clean integration:

- add bond information in the [`Structure`](https://github.com/lab-cosmo/chemiscope/blob/d162f876babae4a7f03ab907dded496d92a63ca2/src/dataset.ts#L88-L130) interface, likely as `Array[[number, number]]`, where each number is the index of one atom in the bonds
- use this bond information in the `StructureViewer`, using the current automatic bond assignment only if the data is not provided in the input.
- update the Python code to emit bond information when converting from user-provided structures. I don't think ASE stores the topology in the `Atoms` objects, so this would be a feature of other structure converters / manually added after conversion.

None of this should break other code, it should be fairly self-contained!

andrewtarzia:
Thank you, both! 

I will start with the `topology_to_shape` approach because it seems simpler. 

Once I have that working, I can share that and we can see if @Luthaf suggestion would work nicely based on it.

Thank you again!

[...]

PR #416, opened 2025-05-14
Custom bonds through a direct argument derived from structures

andrewtarzia:
Regarding to #361, I have implemented somewhat the path toward @Luthaf solution. I submit a draft issue for discussion (Update: it is now a complete PR).

Currently the test just takes a structure that has bonds (`stk` objects), and parses that through `data` to `3Dmol.model`. If bonds are not present in the provided structure, it assigns them as before. Hence:

- [x] A bond reading method is needed for all possible structure types: ase, mdanalysis?

My focus here was on learning basic type script to make the parsing of bond data through possible - I have not finished this, I think, because I could not get it to compile on my machine (`npm run build` worked, but how do I then visualise that output?? (silly question - but if you could teach me how to explore the newly built website, that would be helpful!)).

Once this works, and is handled appropriately, I will introduce the argument to either `structures` or `show`/`write_input` to take a custom bonds argument that the user can modify.

- [x] Modify the example to use the bond argument, if implemented.

* Note: My IDE (Zed) has implemented some formatting changes automatically, that I can/will manually revert at the end.


andrewtarzia:
I just saw the tutorial in Contributing - testing that, sorry!

andrewtarzia:
> I just saw the tutorial in Contributing - testing that, sorry!

This works - silly me for not reading. Continuing development!

Luthaf:
>  A bond reading method is needed for all possible structure types: ase, mdanalysis?

ASE `Atoms` does not support topology/bonding information, and mdanalysis is not implemented yet. So for now I'd say stk support is enough!

> Once this works, and is handled appropriately, I will introduce the argument to either structures or show/write_input to take a custom bonds argument that the user can modify.

I would rather not add an extra `bonds` argument to `write_input` & co, but instead put this it in a more explicit structure conversion function. Since this is mostly orthogonal to the changes in this PR, I would do it in a separate PR.

 I'm thinking we could add something like this:

```py
# name and API to be decided
structures = chemiscope.structures_data(ase_frame, bonds=[...])
structures = chemiscope.structures_data(stk_molecules)

chemiscope.write_input("input.json", structures, ...)
```

src/dataset.ts
```diff
@@ -119,6 +119,11 @@ export interface Structure {
      * expressed in Angströms.
      */
     cell?: number[];
+    /**
+     * bonds in structures given as `[id1 id2 order]`, where
+     * **id1**, **id2**, and **order** are integers.
+     */
+    bonds?: Array<Array<number>>;
```
Luthaf, line 126: I think this should work as well and convey the intent more clearly

```suggestion
    bonds?: Array<[number, number, number]>;
```

src/dataset.ts
```diff
@@ -119,6 +119,11 @@ export interface Structure {
      * expressed in Angströms.
      */
     cell?: number[];
+    /**
+     * bonds in structures given as `[id1 id2 order]`, where
+     * **id1**, **id2**, and **order** are integers.
```
Luthaf, line 124:
```suggestion
     * bonds in structures given as `[i, j, order]`, where
     * `i`, `j`, and `order` are integers. `i` and `j` are the indices of 
     * the atoms bonded together, and `order` is the bond order, which can be 
     * 1 (for single bonds) or 2 (for double bonds)
```

are other bond order supported?

andrewtarzia: I wonder if we should be more general, because bond orders could also be 1.5 for aromatic, and integers up to 9 for rdkit molecules. I will use definitions in 3dmol

andrewtarzia: Note, that this should be handled by the user/structure definition functions to convert appropriately.

ceriottm:
I agree with the need to update example 9 and with your plan.

src/dataset.ts
Luthaf: yes, but we need to have a fairly clear definition of what things mean =). For example I don't think that all kinds of bond order in rdkit are worth supporting, since they will not have a separate representation in 3Dmol (or any other molecule viewer we'll use at some point)

andrewtarzia: Agreed, so I have used your suggestion (adding 3 for triple bonds), and added a converter in `_stk.py` specific to what I know stk does.

andrewtarzia:
I agree and have updated the example to hopefully be clearer about the logic.

I am happy to help update examples, or add more information about stk features if appropriate based on your expansion to other methods/codes.

Luthaf (commented):
We also need some update to the input file reference: https://chemiscope.org/docs/manual/input.html#input-file-reference, and then this should be good to go!

Luthaf (approved):

Luthaf:
Thanks a lot @andrewtarzia!

PR #419, opened 2025-06-06
Validate NaNs in properties

sofiia-chorna:
The issue is when a property contains NaN no widget is displayed but no error message either

### DIFF
--- a/python/chemiscope/input.py
+++ b/python/chemiscope/input.py
@@ -7,6 +7,7 @@ import gzip
 import json
 import os
 import warnings
+from numbers import Number
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


Luthaf:
I'm not sure what the issue is here. Some NaN in properties are allowed, and where used historically to hide some points from the map. 

Can you share an input file that fails?

sofiia-chorna:
> I'm not sure what the issue is here. Some NaN in properties are allowed, and where used historically to hide some points from the map.
> 
> Can you share an input file that fails?

For [this PR](https://github.com/lab-cosmo/chemiscope/pull/420) I used `mts.std_over_samples` to compute the features of `chemiscope.explore` and since I wasn't using the latest version for `mts`, it returned NaNs for some features, and no widget was displayed. After updating to the latest version, it is all good. I would suggest closing this PR

ceriottm:
Actually, maybe it'd be good to print a warning. @sofiia-chorna 's case is one in which the NaNs are accidental (due to rounding errors and a sqrt) so having structures hidden would be bad. I think that throwing a warning (not an error) on the python functions when one tries to build a chemiscope with nans might be a good design decision.

PR #420, opened 2025-06-06
Sketchmap featurizer

sofiia-chorna:


Luthaf:
Could you explain why we need this inside chemiscope? From a quick look it seems the sketchmap could be handled by a custom metatomic model and then go through `metatomic_featurizer`.

sofiia-chorna:
Conclusion from our short discussion:
- do as a metatomic model
-  tbd how to distribute it (e.g. in petmad repo)
- perhaps have a `get_pet_mad_sketchmap_featurizer` for more practical usage
- tbd where to put sketchmap

python/chemiscope/explore/_mad_explorer.py
```diff
@@ -0,0 +1,224 @@
...
+        if isinstance(mtt_model, (str, Path)):
+            self.petmad = mta.load_atomistic_model(
+                mtt_model, extensions_directory=extensions_directory
+            )
+        else:
+            self.petmad = mtt_model
+
+        capabilities = self.petmad.capabilities()
+
+        self.output_name = "mtt::aux::energy_last_layer_features"
```
Luthaf, line 73: this should be using the standard `"features"` output instead of a metatrain-specific one

python/chemiscope/explore/_mad_explorer.py
```diff
@@ -0,0 +1,224 @@
...
+        else:
+            features_tensor = torch.FloatTensor(features)
+
+        with torch.no_grad():
+            X_reduced = self.projector(features_tensor)
+
+        projections = X_reduced.detach().numpy()
+
+        if low_scaler:
+            projections = low_scaler.inverse_transform(X_reduced)
```
Luthaf, line 142: is this using sklearn? This will not be compatible with TorchScript

python/chemiscope/explore/_mad_explorer.py
sofiia-chorna: I see it would be better to use the generic "features" output name but I am afraid that "features" gives different type of representations... 

https://metatensor.github.io/metatrain/latest/advanced-concepts/auxiliary-outputs.html#auxiliary-outputs

Luthaf: It depends if you are trying to create a PET-MAD only sketchmap implementation, or a generic one. For a generic implementation, using `"features"` would be better (and users can always switch which "features" they want to use).

sofiia-chorna:
Moving the PR to pet-mad repo:

https://github.com/lab-cosmo/pet-mad/pull/19

PR #422, opened 2025-07-04
Enhance docs of featurize argument of `chemiscope.explore`

PicoCentauri:
I think the current docs of the `featurize` argument in chemiscope.explore do not reflect the actual requirements on the callable. I updated the text and also did a minor proofread of the whole docstring.

### DIFF
--- a/python/chemiscope/explore/__init__.py
+++ b/python/chemiscope/explore/__init__.py
@@ -22,7 +22,7 @@ def explore(
     """
     Automatically explore a dataset containing all structures in ``frames``.
 
-    This function computes some low-dimensionality representation of the frames, and
+    This function computes a low-dimensionality representation of the frames, and
     uses chemiscope to visualize both the resulting embedding and the structures
     simultaneously. ``featurize`` can be used to specify a custom representation and/or
     dimensionality reduction method.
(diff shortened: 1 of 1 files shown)


python/chemiscope/explore/__init__.py
```diff
@@ -35,10 +35,12 @@ def explore(
...
     :param list frames: list of frames
 
-    :param callable featurize: optional. Function to compute features and perform
-        dimensionality reduction on the ``frames``. The function should take ``frames``
-        as input and return a 2D array of reduced features. If ``None``, a default SOAP
-        and PCA based featurizer is used.
+    :param callable featurize: Optional callable to compute features and perform
+        dimensionality reduction on the ``frames``. The callable should take ``frames``
+        as the first argument and ``environments`` as the second argument. The return
+        value must be a features array of dimension number of frames times number of
```
Luthaf, line 41: not quite: the shape should be n_frames x n_features if `environments is None`, and n_environments x n_features otherwise

sofiia-chorna: Indeed, with environnements provided it should be n_environments x n_features, I will add it. Thanks

PR #426, opened 2025-07-13
Clean-up and simplify the usage of the explore module

ceriottm:
The idea is to reduce the complexity around using the explore module:
* provide a single chemiscope[explore] that install both metatomic and pet_mad
* remove dependency on dscribe
* one example of pet_mad explore, one example of making a simple (density?) featurizer, and storing it as a metatomic model


sofiia-chorna:
Regarding the third point, we already have an example using a metatomic model here https://chemiscope.org/docs/examples/8-explore-with-metatomic.html. The idea is to add another simpler example? Also, since the pet_mad featurizer is used as the default in explore, the basic explore example will already demonstrates it, is that sufficient?

ceriottm:
> Regarding the third point, we already have an example using a metatomic model here https://chemiscope.org/docs/examples/8-explore-with-metatomic.html. The idea is to add another simpler example? Also, since the pet_mad featurizer is used as the default in explore, the basic explore example will already demonstrates it, is that sufficient?

I was thinking of having these:
1. a demo of the default chemiscope.explore that uses pet-mad behind the scenes
2. a very simple demo of how to make your own explorer, e.g. computing the fractional composition. this could also then contain an example of how to wrap that into a metatomic model maybe (as it shows how to export, which is not trivial)
3. the current "advanced" model that uses MACE and t-sne (which I presume doesn't really need dscribe right?)

Luthaf:
> a very simple demo of how to make your own explorer, e.g. computing the fractional composition. this could also then contain an example of how to wrap that into a metatomic model maybe (as it shows how to export, which is not trivial)

We don't need to go through metatomic here. A simple featurizer can just be a function in chemiscope, and I think this is a more relevant example here.  Computing the fractional composition sounds like a good idea!

sofiia-chorna:
Currently the build website job fails => another version should be used for the petmad checkpoint, without UQ wrapper => https://github.com/lab-cosmo/pet-mad/pull/30 should be merged for the fix of this issue

python/examples/6-explore.py
```diff
@@ -28,154 +30,102 @@
...
 # Basic example
 # +++++++++++++
 #
-# This example shows the basic usage of the :py:func:`chemiscope.explore`. At first,
-# read or load the structures from the dataset. Here we use an `ASE package
-# <https://wiki.fysik.dtu.dk/ase>`_ to read the structures from the file and have the
-# frames as the `ase.Atoms <https://wiki.fysik.dtu.dk/ase/ase/atoms.html>`_ objects.
+# This example shows the basic usage of :py:func:`chemiscope.explore`. First, load a
+# dataset of structures as `ase.Atoms <https://wiki.fysik.dtu.dk/ase/ase/atoms.html>`_
+# objects. Here, we use the samples from the M3CD dataset:
```
Luthaf: worth a link to the MC3D dataset

PR #428, opened 2025-07-18
Improvements of `featurizer`

sofiia-chorna:
Update `chemiscope.explore` with the following changes:

- accept string for `featurizer` for the default featurizer to be used
- automatically extract all properties from the structures
- allow setting `None` for the `featurizer` => extracted / user-defined properties will be displayed
- rename `featurize` to `featurizer`

The branch starts from `explore-cleanup`, to be merged after https://github.com/lab-cosmo/chemiscope/pull/426


python/chemiscope/explore/__init__.py
```diff
@@ -1,22 +1,73 @@
...
-__all__ = [
-    "explore",
-    "metatomic_featurizer",
-]
+__all__ = ["explore", "metatomic_featurizer", "get_featurizer"]
+
+FEATURIZER_VERSION_MAP = {"pet-mad-1.0": "1.0.0"}
+
+
+def get_featurizer(name, device=None, batch_size=1):
```
Luthaf:
```suggestion
def get_featurizer(name):
```

This function is more of a convenience function that could know about a bunch of featurizers, for some of which these would not be valid parameters. I think at this point we can make a choice for the user and give them a good default option.

python/chemiscope/explore/__init__.py
```diff
@@ -1,22 +1,73 @@
...
+
+            pip install chemiscope[explore]
+    """
+    try:
+        from pet_mad.explore import PETMADFeaturizer
+    except ImportError as e:
+        raise ImportError(
+            f"Required package not found: {e}. Please install the "
+            "dependencies with `pip install chemiscope[explore]`."
+        )
```
Luthaf:
```suggestion
            f"Required package not found. Please install the "
            "dependencies with `pip install chemiscope[explore]`."
        ) from e
```

this keeps more of the context around about why we failed to load the package

python/chemiscope/explore/__init__.py
```diff
@@ -1,22 +1,73 @@
...
+
+    if not isinstance(name, str):
+        raise TypeError(f"Featurizer name must be a string, got {type(name)}")
```
Luthaf:
```suggestion
        raise TypeError(f"featurizer name must be a string, not {type(name)}")
```

python/chemiscope/explore/__init__.py
```diff
@@ -1,22 +1,73 @@
...
+        raise ValueError(
+            f"The featurizer version {name} does not exist. Available options are: "
+            f"{', '.join(versions) if len(versions > 1) else versions[0]}"
+        )
+
+    version = FEATURIZER_VERSION_MAP.get(name)
+
+    featurizer = PETMADFeaturizer(version=version, device=device, batch_size=batch_size)
+
+    return featurizer
```
Luthaf: I would rather do something like this:

```suggestion
    if name == "pet-mad-1.0":
        device = ...
        model = PETMADFeaturizer(version="1.0.0", device=device, batch_size=1)
        return metatomic_featurizer(model)
    else:
        raise ValueError("unknown featurizer: …")
```

python/chemiscope/explore/__init__.py
```diff
@@ -30,14 +81,27 @@ def explore(
...
-        ``environments`` is ``None``, or ``(n_environments, n_features)`` otherwise. If
-        ``None``, a default ``PETMADFeaturizer`` is used.
+    :param featurizer: either string specifying a featurizer version (currently only
+        'pet-mad-1.0'), a custom callable function, or None. Used to compute features
+        and perform dimensionality reduction on the ``frames``. For automatic default
+        option, use ``pet-mad-1.0``. The callable should take ``frames`` as the first
+        argument and ``environments`` as the second argument. The return value must be a
+        features array of shape ``(n_frames, n_features)`` if ``environments`` is
+        ``None``, or ``(n_environments, n_features)`` otherwise. By providing a string
+        version, the related ``PETMADFeaturizer`` is used.
```
Luthaf:
```suggestion
        ``None``, or ``(n_environments, n_features)`` otherwise.
```

python/chemiscope/explore/__init__.py
```diff
@@ -63,6 +127,17 @@ def explore(
     :param int batch_size: optional. Number of structures processed in each batch with
         the default ``PETMADFeaturizer``.
 
+    :param string write_input: optional. A path to save the chemiscope visualization.
```
Luthaf:
```suggestion
    :param string write_input: optional. A path to save the chemiscope input file created by this function.
```

python/chemiscope/explore/__init__.py
```diff
@@ -89,7 +164,11 @@ def explore(
         frames = ase.io.read("trajectory.xyz", ":")
 
         # 1) Basic usage with the default featurizer (PET-MAD featurization + SMAP)
```
Luthaf:
```suggestion
        # 1) Basic usage with the default featurizer (PET-MAD featurization + Sketch-Map)
```

python/chemiscope/explore/__init__.py
```diff
@@ -124,39 +203,108 @@ def soap_kpca_featurize(frames, environments):
...
 
-    # Apply dimensionality reduction from the provided featurizer
-    if featurize is not None:
-        X_reduced = featurize(frames, environments)
-    else:
-        # Run default featurizer
-        featurizer = PETMADFeaturizer(
-            version="latest", device=device, batch_size=batch_size
+    extracted_properties = extract_properties(frames, environments)
+    merged_properties = {**extracted_properties, **properties}
```
Luthaf: does this override things from the frame with whatever the user gave if they have the same name?

python/chemiscope/explore/__init__.py
```diff
@@ -1,22 +1,73 @@
...
 from ._metatomic import metatomic_featurizer
 
 
-__all__ = [
-    "explore",
-    "metatomic_featurizer",
-]
+__all__ = ["explore", "metatomic_featurizer", "get_featurizer"]
+
+FEATURIZER_VERSION_MAP = {"pet-mad-1.0": "1.0.0"}
```
Luthaf: How about 
```suggestion
KNOWN_FEATURIZERS = ["pet-mad-1.0"]
```

python/chemiscope/explore/__init__.py
```diff
@@ -124,39 +203,108 @@ def soap_kpca_featurize(frames, environments):
...
+        if hasattr(value, "shape") and len(value.shape) > 1:
+            # for multi-dimentional props, each will be displayed as a separate property
+            # in the widget
+            count += value.shape[1]
+        else:
+            count += 1
+    return count
+
+
+def _configure_map_settings(settings, merged_properties):
```
Luthaf: why is this needed? It seems to be duplicating logic from the TypeScript code

python/chemiscope/explore/__init__.py
sofiia-chorna: yes, I could display a warning in this case?

python/chemiscope/explore/__init__.py
sofiia-chorna: it ensures that the default axes are set to the "features" and color is another property if available. by default, from my tests, it seems it does not work like this (or maybe not supposed to work this way)

without this, another property might be automatically selected as the axis => user needs to manually adjust it /  define in the settings in advance

python/chemiscope/explore/__init__.py
Luthaf: I'm not sure. We should 100% pick the values from the user over the values from the frame, but maybe a warning is not needed? Especially since the user can't control it directly

python/chemiscope/explore/__init__.py
Luthaf: Ah, I would just create the properties dict by adding the features there first if we are computing them. They should end up first in the JSON input and be picked by the visualizer.

docs/src/python/jupyter.rst
```diff
@@ -16,3 +16,9 @@ Creating a chemiscope widget
 
 .. autofunction:: chemiscope.show
 .. autofunction:: chemiscope.show_input
+
+Dataset exploration
+-------------------
+
+.. autofunction:: chemiscope.explore
+.. autofunction:: chemiscope.metatomic_featurizer
```
Luthaf:
```suggestion
.. autofunction:: chemiscope.explore
.. autofunction:: chemiscope.get_featurizer
.. autofunction:: chemiscope.metatomic_featurizer
```

python/chemiscope/explore/__init__.py
```diff
@@ -1,22 +1,73 @@
...
+    <https://arxiv.org/abs/2506.19674>`_.
+
+    :param str name: name of the featurizer. Must match one of the known versions.
+        Currently available is "pet-mad-1.0"
+
+    :param str device: optional. device to run the featurizer on (e.g., "cpu" or "cuda")
+
+    :param int batch_size: optional. number of structures to process per batch during
+        featurization, defaults to 1
+
```
Luthaf:
```suggestion
```

python/chemiscope/explore/__init__.py
```diff
@@ -1,22 +1,73 @@
...
+    .. warning::
+
+        Requires additional dependencies. Install it using:
```
Luthaf:
```suggestion
        This function requires additional dependencies. Install them using:
```

python/chemiscope/explore/__init__.py
```diff
@@ -1,22 +1,73 @@
...
+        return model
+    else:
+        options = (
+            ", ".join(KNOWN_FEATURIZERS)
+            if len(KNOWN_FEATURIZERS) > 1
+            else KNOWN_FEATURIZERS[0]
+        )
+        raise ValueError(
+            f"unknown featurizer: {name}. Available options are: {options}"
+        )
```
Luthaf: `join` does nothing with a 1 element list, so you can simplify here:

```suggestion
        raise ValueError(
            f"unknown featurizer: {name}. Available options are: {', '.join(KNOWN_FEATURIZERS)}"
        )
```

Luthaf (commented):
just a handful of tweaks!

Luthaf (approved):

PR #429, opened 2025-07-18
Check for lengths consistency in properties, ignore inconsistent

sofiia-chorna:


python/chemiscope/structures/_ase.py
```diff
@@ -222,19 +222,36 @@ def _remove_invalid_properties(properties, origin):
...
 
+        if name in to_remove or not hasattr(values[0], "__len__"):
+            continue
+
+        # check length consistency
+        lengths = [len(v) for v in property["values"]]
+        if len(set(lengths)) != 1:
+            to_remove.add(name)
+            warnings.warn(
+                f"values of the property '{name}' have inconsistent length, it will be "
```
Luthaf:
```suggestion
                f"values of the property '{name}' have inconsistent length across different structures, it will be "
```

PR #431, opened 2025-07-24
Add rst examples to the sphinx gallery documetation

sofiia-chorna:


docs/src/python/sphinx.rst
```diff
@@ -44,15 +44,27 @@ Once compiled, this will show as this widget
 
 .. chemiscope:: ../datasets/showcase.json.gz
 
-The ``structure`` mode will show the structures only,
-with disabled warnings
+The ``structure`` mode will show the structures only, with disabled warnings. It is also
+possible to add an optional ``warning_timeout`` parameter:
```
Luthaf, line 48: maybe also explain what this parameter does, and what `-1/0` means?

sofiia-chorna: It is explained before in the this documenation, here:
https://github.com/lab-cosmo/chemiscope/blob/1347a5998fc2c40173eb542bbd0c80349bd1b40a/docs/src/python/sphinx.rst?plain=1#L31-L34

Do you think it would be better/clearer to specify it here one more time?

Luthaf: ah, I missed it. Then everything is fine!

Luthaf (approved):

PR #439, opened 2025-09-30
Allow using np.array of string as categorical values

### LINKED ISSUE #427: Categorical properties bug
I was following the documentation example for Atom property coloring and it seems, that the atom properties are given with the shape (N,1) instead of (N,). There is an error when trying to use a categorical property with shape (N,1), as it is not recognizing the string in 
https://github.com/lab-cosmo/chemiscope/blob/05e2e1f236864315a9a259bd33ee39c19378933d/python/chemiscope/input.py#L925C5-L925C33. 

I think simply flattening the input would solve the issue!

Luthaf:
Fix #427

### DIFF
--- a/python/chemiscope/input.py
+++ b/python/chemiscope/input.py
@@ -940,6 +940,16 @@ def _validate_property(name, property):
             warnings.warn(f"ignoring unexpected property key: {key}", stacklevel=2)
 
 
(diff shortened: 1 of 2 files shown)


sofiia-chorna (approved):
Looks good to me. Thank you for the clean refactoring

PR #442, opened 2025-10-23
Implement cartoon representation

GardevoirX:
[protein_all.zip](https://github.com/user-attachments/files/23108373/protein_all.zip)
To test:
```python
import chemiscope
import MDAnalysis as mda

universe = mda.Universe("protein_all.pdb" , "protein_all.xtc")
chemiscope.show(
    frames=universe.atoms,
    mode="structure",
)
```

Preview:
image

### DIFF
--- a/python/chemiscope/structures/_mda.py
+++ b/python/chemiscope/structures/_mda.py
@@ -45,7 +45,7 @@ def _mda_to_json(ag):
     if hasattr(ag, "chainIDs") and ag.chainIDs is not None:
         data["chains"] = [atom.chainID for atom in ag]
     if hasattr(ag, "resnames") and ag.resnames is not None:
-        data["residues"] = [
+        data["resnames"] = [
             atom.resname if atom.resname is not None and atom.resname != "" else "UNK"
             for atom in ag
         ]
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -138,13 +138,24 @@ export interface Structure {
     };
     /** Element names of all atoms in the structure, if available */
     elements?: string[];
-    /** Residue names of all atoms in the structure, if available. Under the case where some of the residue names are not available, `UNK` would be automatically assigned */
-    residues?: string[];
-    /** Residue IDs of all atoms in the structure, if available */
+    /** Residue name for all atoms in the structure, if available.
+     * Atoms without a residue (e.g. water molecules, ions, etc.) should have
+     * their resname set to `"UNK"`.
+     *
+     * This is used for cartoon representation. */
+    resnames?: string[];
+    /** Residue index for all atoms in the structure, if available.
+     * This is only used for secondary structure assignment. */
     resids?: number[];
-    /** Chain names of all atoms in the structure, if available */
+    /** Chain names of all atoms in the structure, if available.
+     * This is only used for secondary structure assignment. */
     chains?: string[];
-    /** If atom is a heteroatom, i.e. not part of a biomolecule (protein, dna, …) */
+    /**
+     * Wether an atom is a heteroatom, i.e. not part of a biomolecule (protein, dna, …).
+     * If this is not provided, it is assumed to be true for all atoms.
+     *
+     * This is only used for secondary structure assignment and cartoon representation.
+     */
     hetatom?: boolean[];
 }
 
@@ -556,7 +567,7 @@ export function checkStructure(s: JsObject): string {
     }
 
     let biomolInfoCount = 0;
-    for (const key of ['hetatom', 'chains', 'residues', 'resids']) {
+    for (const key of ['hetatom', 'chains', 'resnames', 'resids']) {
         if (key in s) {
             biomolInfoCount++;
             const array = s[key];
@@ -569,7 +580,7 @@ export function checkStructure(s: JsObject): string {
         }
     }
     if (biomolInfoCount > 0 && biomolInfoCount !== 4) {
-        return 'found at least one of "hetatom", "chains", "residues" and "resids", but not all of them';
+        return 'found at least one of "hetatom", "chains", "resnames" and "resids", but not all of them';
     }
 
     return '';
--- a/src/structure/viewer.ts
+++ b/src/structure/viewer.ts
@@ -60,8 +60,8 @@ function setup3DmolStructure(model: $3Dmol.GLModel, structure: Structure): void 
             ss: 'c', // initialize to coil by default, the same as in 3Dmol
         } as unknown as $3Dmol.AtomSpec;
 
-        if (structure.residues !== undefined) {
-            atom.resn = structure.residues[i];
+        if (structure.resnames !== undefined) {
+            atom.resn = structure.resnames[i];
         }
 
         if (structure.chains !== undefined) {
@@ -513,7 +513,7 @@ export class MoleculeViewer {
         }
         this._showSupercellInfo();
 
-        if (structure.resids === undefined) {
+        if (structure.resnames === undefined) {
             this._styles.noResidue.replaceSync('.chsp-hide-if-not-biomol { display: none; }');
         } else {
             this._styles.noResidue.replaceSync('');
--- a/docs/src/manual/input.rst
+++ b/docs/src/manual/input.rst
@@ -116,9 +116,10 @@ contains the following fields and values:
             {
                 // number of atoms in the structure
                 "size": 42,
-                // elements of the atoms in the structure
-                "elements": ["H", "O", "C", "C", ...],
-                // names of the atoms in the structure, can be "CA", "CG1"... if your are using `MDAnalysis` and passing in biomolecule stuff
+                // names of the atoms in the structure. If the atom names do not
+                // correspond to element names (for example in biomolecules,
+                // using CA for some Carbon), the `elements` field should also
+                // be provided.
                 "names": ["H", "O", "C", "C", ...],
                 // x cartesian coordinate of all the atoms, in Angstroms
                 "x": [0, 1.5, 5.2, ...],
@@ -132,6 +133,8 @@ contains the following fields and values:
                 // a, b, and c are the unit cell vectors. All values are
                 // expressed in Angstroms.
                 "cell": [10, 0, 0, 0, 10, 0, 0, 0, 10],
+                // OPTIONAL: name of the elements of the atoms in the structure
+                "elements": ["H", "O", "C", "C", ...],
                 // OPTIONAL: bonds in the system, if any.
                 //
                 // This should be given as [i, j, order], where i, j, and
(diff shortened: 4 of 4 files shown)


python/chemiscope/structures/__init__.py
```diff
@@ -65,7 +66,10 @@ def frames_to_json(frames):
     elif adapter == "mda":
         # Be careful of the lazy loading of `frames.atoms`, which is updated during the
         # iteration of the trajectory
-        return [_mda_to_json(frames) for _ in frames.universe.trajectory]
+        data_collection = [_mda_to_json(frames) for _ in frames.universe.trajectory]
+        if all("residues" in data for data in data_collection):
+            data_collection = _mda_get_secondary_structure(data_collection, frames)
```
Luthaf: this should ideally be done by `_mda_to_json`, the functions in this file should not have to care

src/dataset.ts
```diff
@@ -136,6 +136,20 @@ export interface Structure {
          */
         [name: string]: ShapeParameters;
     };
+    /** Element names of all atoms in the structure, if available */
+    elements?: string[];
+    /** Residue names of all atoms in the structure, if available */
+    residues?: string[];
```
Luthaf: So how should we represent atoms without a residue name here? Do we assume that everything has a resname, even small molecules?

src/dataset.ts
```diff
@@ -136,6 +136,20 @@ export interface Structure {
...
+    /** Residue IDs of all atoms in the structure, if available */
+    resids?: number[];
+    /** Chain names of all atoms in the structure, if available */
+    chains?: string[];
```
Luthaf, line 152: is this required for cartoon?

src/structure/viewer.ts
```diff
@@ -26,6 +27,15 @@ function defaultOpacity(): number {
     return 0.85;
 }
 
+interface Atom {
```
Luthaf: I think this should use the interface from 3dmol instead of redefining it

src/structure/viewer.ts
```diff
@@ -6,6 +6,7 @@
 import assert from 'assert';
 
 import * as $3Dmol from '3dmol';
+// import { computeSecondaryStructure } from "3dmol/src/parsers/utils/computeSecondaryStructure";
```
Luthaf: so did you try using this instead of the MDA code?

src/structure/viewer.ts
```diff
@@ -156,6 +189,9 @@ export class MoleculeViewer {
     private _resetSupercell!: HTMLButtonElement;
     /// Show some information on the currently displayed cell to the user
     private _cellInfo: HTMLElement;
+    /// Show some information on the currently displayed protein (or other things) to
+    // the user
```
Luthaf:
```suggestion
    /// Show settings specific to biomolecules
```

src/dataset.ts
GardevoirX: I am sure that `resids` is required. For `chains`, I guess it's needed to determine where the cartoon representation starts and ends

src/structure/viewer.ts
GardevoirX: I tried but encountered with numerous importing errors, so I gave up

python/chemiscope/structures/__init__.py
GardevoirX: This is a workaround, I haven't found a way to calculate the secondary frame by frame. The current calculating function would reset the pointer of the frame to the first frame, leading to infinite loop if we do it in `_mda_to_json`

src/dataset.ts
GardevoirX: This is more for the compatibility of `ase` and `stk`, not sure if they have resname. but even `mda` can give atoms that doesn't have resname, e.g. structure loaded from xyz

GardevoirX:
## 1AV5 (protein-ligand complex)
image

## 9N2N (protein-nucleic complex)
image

src/dataset.ts
Luthaf: And then the resname is an empty string?

src/dataset.ts
Luthaf: Did you check whether it is required or not?

src/structure/viewer.ts
Luthaf: ok, I'll check if I can get it to work

src/dataset.ts
GardevoirX: No under this case there won't be the `resname` property

src/dataset.ts
GardevoirX: I checked, yes the chain id is unnecessary for the cartoon representation, but the secondary structure calculation in 3dmol requires it
https://github.com/3dmol/3Dmol.js/blob/5bc64414929f6d269c5e5a674e4707c1fd541738/src/parsers/utils/computeSecondaryStructure.ts#L16-L24

Luthaf:
I changed the code to use the secondary structure calculation from 3Dmol instead of the one from mda, meaning I could also remove some of the fields from TypeScript. 

This still needs more testing, I'll have a go tomorrow. Two things I saw already:

- the code is missing validation for the new fields in the `Structure` interface (here: https://github.com/lab-cosmo/chemiscope/blob/c8beb47608a1a7c95c7bebdeedcb18634bda02ec/src/dataset.ts#L519)
- Disabling "cartoon" should switch back to ball & stick for everything, but it currently just removes the protein altogether

GardevoirX:
> I changed the code to use the secondary structure calculation from 3Dmol instead of the one from mda, meaning I could also remove some of the fields from TypeScript.
> 
> This still needs more testing, I'll have a go tomorrow. Two things I saw already:
> 
> * the code is missing validation for the new fields in the `Structure` interface (here: https://github.com/lab-cosmo/chemiscope/blob/c8beb47608a1a7c95c7bebdeedcb18634bda02ec/src/dataset.ts#L519
>   )
> * Disabling "cartoon" should switch back to ball & stick for everything, but it currently just removes the protein altogether

It's good now!

src/dataset.ts
Luthaf: I mean inside the `Structure` interface, what should be stored in `residues` if there is no resname?

src/dataset.ts
```diff
@@ -525,7 +535,7 @@ export function checkStructure(s: JsObject): string {
         return 'missing "size"';
     }
 
-    for (const key of ['names', 'x', 'y', 'z']) {
+    for (const key of ['names', 'x', 'y', 'z', 'hetatom']) {
```
Luthaf: `hetatom` is still optional, it should not be here!

src/dataset.ts
```diff
@@ -545,6 +555,15 @@ export function checkStructure(s: JsObject): string {
         }
     }
 
+    for (const key of ['chainIDs', 'resnames', 'resids']) {
```
Luthaf: there is no `chainIDs` or `resnames`, please review and test LLM-generated code before committing =)

src/structure/viewer.ts
```diff
@@ -156,6 +178,8 @@ export class MoleculeViewer {
     private _resetSupercell!: HTMLButtonElement;
     /// Show some information on the currently displayed cell to the user
     private _cellInfo: HTMLElement;
+    /// Show settings specific to biomolecules
+    private _residueInfo: HTMLElement;
```
Luthaf:
```suggestion
    private _biomolOptions: HTMLElement;
```

Luthaf: it is not really about residues

python/chemiscope/structures/_mda.py
```diff
@@ -11,6 +11,14 @@
...
 
 
+SS_MAPPING = {
+    # Markers of secondary, from mda's to 3dmol's
+    "H": "h",
+    "E": "s",
+    "-": "c",
+}
+
+
```
Luthaf:
```suggestion
```

no longer required

src/dataset.ts
```diff
@@ -545,6 +555,15 @@ export function checkStructure(s: JsObject): string {
...
+        if (key in s) {
```
Luthaf, line 571: We should also validate that all of the biomolecule options are defined together if anyone of them is defined

src/dataset.ts
GardevoirX: sry it's actually the llm inside my brain went wrong :cry:

ceriottm: QianjunGPT-5.0-Instant

src/dataset.ts
GardevoirX: idk but for pdb format the columns 18-20 are interpreted as the resname. See [this](https://www.rbvi.ucsf.edu/chimera/docs/UsersGuide/tutorials/pdbintro.html).
In case of strange inputs, I just require that the size of `residues` and other bioinfos must equal to the number of atoms

Luthaf: We are the one specifying what should go in there, it is our data format. I'm happy if we decide to use an empty string if there is no residue information for a given atom (so `residues` could look like `["ALA", "ALA", "", "", "", "ALA"]`. 

If this works for you as well, could you make sure it is clear in the input file reference in the documentation?

src/dataset.ts
```diff
@@ -555,14 +555,36 @@ export function checkStructure(s: JsObject): string {
...
             if (!Array.isArray(array)) {
                 return `"${key}" must be an array`;
             }
+            if (s.size > 0 && array.length !== s.size) {
+                return `wrong size for "${key}", expected ${s.size}, got ${array.length}`;
+            }
         }
     }
+    if (biomolInfoCount > 0 && biomolInfoCount !== 3) {
+        return 'found at least one of "chains", "residues" and "resids", but not all of them';
```
Luthaf: should we also include `hetatom` here?

src/structure/viewer.ts
```diff
@@ -266,9 +266,9 @@ export class MoleculeViewer {
         );
         this._root.appendChild(this._cellInfo);
 
-        this._residueInfo = document.createElement('span');
-        this._residueInfo.classList.add('chsp-hide-if-no-residue');
-        this._root.appendChild(this._residueInfo);
+        this._biomolOptions = document.createElement('span');
+        this._biomolOptions.classList.add('chsp-hide-if-no-residue');
```
Luthaf: we should also rename the class, maybe `chsp-hide-if-not-biomol`?

src/dataset.ts
GardevoirX: yes sure

src/dataset.ts
GardevoirX: Yes this works for me, but my concern is if there is a valid example where some atoms have the residue name while others don't have. If we can have one example, I'd be happy to add it. For the `.pdb` and `.gro` format, it seems to be impossible to craft such an example

python/chemiscope/structures/_mda.py
```diff
@@ -45,7 +45,10 @@ def _mda_to_json(ag):
     if hasattr(ag, "chainIDs") and ag.chainIDs is not None:
         data["chains"] = [atom.chainID for atom in ag]
     if hasattr(ag, "resnames") and ag.resnames is not None:
-        data["residues"] = [atom.resname for atom in ag]
+        data["residues"] = [
+            atom.resname if atom.resname is not None and atom.resname != "" else "UNK"
```
Luthaf, line 49: This works for me, could you please document that this is the expected value in `dataset.ts` and the input file reference documentation?

Luthaf:
Thanks @GardevoirX, this is very nice!

PR #448, opened 2025-11-15
Add `write_external_structures`, use it in Jupyter and Sphinx integrations

ceriottm:
Given we are starting to work with larger structures, it's becoming tricky to load everything in memory, so this PR makes it possible to use the mechanism to load structures from disk also from jupyter & c.

### DIFF
--- a/docs/src/python/sphinx.rst
+++ b/docs/src/python/sphinx.rst
@@ -3,9 +3,9 @@
 Chemiscope in ``sphinx``
 ========================
 
-The chemiscope python module also provides a ``sphinx`` extension that can process
-a custom RST directive into an interacting viewer, that can be embedded in the
-generated HTML documentation.
+The chemiscope python module also provides a ``sphinx`` extension that can
+process a custom RST directive into an interacting viewer, that can be embedded
+in the generated HTML documentation.
 
 Configuration
 ^^^^^^^^^^^^^
@@ -44,8 +44,8 @@ Once compiled, this will show as this widget
 
 .. chemiscope:: ../datasets/showcase.json.gz
 
-The ``structure`` mode will show the structures only, with disabled warnings. It is also
-possible to add an optional ``warning_timeout`` parameter:
+The ``structure`` mode will show the structures only, with disabled warnings. It
+is also possible to add an optional ``warning_timeout`` parameter:
 
 .. code-block:: rst
 
@@ -76,13 +76,15 @@ External structure files
 
 It is possible to use the ``chemiscope`` directive to load datasets that
 reference external structure files. This is useful when the structures are
-large, or when the same structures are to be used in multiple datasets.
-The main `dataset.json` file should contain structure entries that reference
-external `.json` files - that can be created for instance using
-:py:func:`~chemiscope.write_external_structures`.
-The paths should be defined relative 
-to the position of the main chemiscope file, and the external files will 
-be copied to the documentation build folder automatically. 
+large, or when the same structures are to be used in multiple datasets. The main
+`dataset.json` file should contain structure entries that follows the
+`UserStructure`_ interface, with the ``data`` field set to the path of the
+structure to be loaded. The structure files can be created manually or using
+:py:func:`~chemiscope.write_external_structures`. The paths should be defined
+relative to the position of the main chemiscope file, and the external files
+will be copied to the documentation build folder automatically.
+
+.. _UserStructure: https://chemiscope.org/docs/api/interfaces/UserStructure.html
 
 .. code-block:: rst
 
@@ -97,10 +99,10 @@ be copied to the documentation build folder automatically.
 Viewing a documentation locally
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Because of the need to dynamically load the chemiscope module and the
-desired file, it is not possible to load the widget by viewing the
-the raw HTML pages. Instead, one needs to run a local HTTP server.
-For instance, using the port 8765,
+Because of the need to dynamically load the chemiscope module and the desired
+file, it is not possible to load the widget by viewing the the raw HTML pages.
+Instead, one needs to run a local HTTP server. For instance, using the port
+8765,
 
 .. code-block:: bash
 
--- a/python/jupyter/src/widget.ts
+++ b/python/jupyter/src/widget.ts
@@ -1,4 +1,5 @@
 import { DOMWidgetView } from '@jupyter-widgets/base';
+import { JSONValue } from '@lumino/coreutils';
 import Plausible from 'plausible-tracker';
 
 import { Warnings, generateGUID, getByID } from '../../../src/utils';
@@ -172,7 +173,7 @@ class ChemiscopeBaseView extends DOMWidgetView {
                 type: 'load-structure',
                 requestId,
                 index,
-                data: structure.data as string,
+                data: structure.data as JSONValue,
             });
         });
     }
--- a/src/index.ts
+++ b/src/index.ts
@@ -502,11 +502,15 @@ class DefaultVisualizer {
             copy.structures = [] as Structure[];
             for (let i = 0; i < this._dataset.structures.length; i++) {
                 const structure = this.structure.loadStructure(i, this._dataset.structures[i]);
-                // this code path should only be reached with the default loader, that is
-                // synchronous
-                assert(!(structure instanceof Promise));
 
-                copy.structures.push(structure);
+                Promise.resolve(structure).then(
+                    (resolved) => {
+                        copy.structures[i] = resolved;
+                    },
+                    (err: unknown) => {
+                        throw Error(`could not load structure at index ${i}: ${err}`);
+                    }
+                );
             }
         }
         return copy;


src/index.ts
```diff
@@ -501,7 +501,14 @@ class DefaultVisualizer {
         if (getStructures) {
             copy.structures = [] as Structure[];
             for (let i = 0; i < this._dataset.structures.length; i++) {
-                copy.structures.push(this.structure.loadStructure(i, this._dataset.structures[i]));
+                const structure = this.structure.loadStructure(i, this._dataset.structures[i]);
+                if (structure instanceof Promise) {
+                    throw new Error(
+                        'loadStructure returned a Promise, but this code path must be synchronous'
+                    );
```
sofiia-chorna: It'd be helpful if the message explained why it could happen...

docs/src/python/sphinx.rst
```diff
@@ -70,6 +70,29 @@ The ``map`` mode shows only the property map, with persistent warnings
     :mode: map
     :warning_timeout: 0
 
+
+External structure files
+^^^^^^^^^^^^^^^^^^^^^^^^
```
sofiia-chorna, line 75: this feature is very cool however I find this example not very clear. it looks identical to previous sphinx examples even though the difference is _inside_ the dataset file.
maybe we can show a minimal json or python snippet with `write_external_structures`, unless there are some better ideas

python/chemiscope/sphinx/utils.py
```diff
@@ -13,3 +15,63 @@ def copy_file(src_file, dst_file):
...
+            )
+
+        rel_path = os.path.normpath(rel_path)
+
+        # Source file: relative to *original* JSON location
+        src_struct = os.path.join(src_base, rel_path)
+
+        if not os.path.exists(src_struct):
+            # Optional: log a warning if you have access to a logger
+            # logger.warning("Missing external structure file %s", src_struct)
```
sofiia-chorna: we can delete it I guess

python/chemiscope/sphinx/scraper.py
```diff
@@ -48,16 +50,25 @@ def __call__(self, _block, block_vars, _gallery_conf):
...
             dataset_file_path = os.path.join(rst_dataset_dir, dataset_filename)
             widget.save(dataset_file_path)
 
+            # also saves in-place to facilitate copying external structures
+            src_file = block_vars.get("src_file")
+            src_dataset = os.path.join(os.path.dirname(src_file), dataset_filename)
+            widget.save(src_dataset)
+
+            print(" SCRAPER COPYING FILES ")
+            print(src_file, src_dataset, dataset_filename, rst_dataset_dir)
```
sofiia-chorna: I'd delete those logs

python/chemiscope/input.py
```diff
@@ -377,6 +381,44 @@ def create_input(
...
+        frames = ase.io.read("trajectory.xyz", ":")
+        user_frames = chemiscope.write_external_structures(
+            frames, prefix="my_structure"
+        )
+        write_input("chemiscope.json", frames=user_frames)
+
+
+    :param list frames: list of atomic structures in a format that can
+            be understood by `chemiscope`.
+    :param str prefix: prefix to use for the generated JSON files.
```
sofiia-chorna:
```suggestion
    :param str prefix: prefix to use for each generated JSON filename. Files will be named like "{prefix}-0.json", "{prefix}-1.json" etc.
```

sofiia-chorna (commented):
Very cool feature! Thanks a lot for implementation and clean up.

However, I wasn't able to get it working in a notebook, although the examples in the docs generated with `tox -e docs` work fine in the browser. In notebook I do:

```
import ase.io
import chemiscope

frames = ase.io.read("python/examples/data/showcase.xyz", ":")

external_frames = chemiscope.write_external_structures(frames, "showcase-external")

chemiscope.show(
    frames=external_frames,
    mode="structure",
)
```
but no structures appeared nor error
Screenshot 2025-11-16 at 00 48 15

ceriottm:
> Very cool feature! Thanks a lot for implementation and clean up.
> 
> However, I wasn't able to get it working in a notebook, although the examples in the docs generated with `tox -e docs` work fine in the browser. 

are you sure that it has updated the chemiscope package you are using in your environment? check the JS console in the browser, if it says there is no loadStructure callback, then it's likely it's using the `main` CS widget

src/index.ts
ceriottm: I will turn this into an assert, as this is more to check that the underlying assumption we are making that this does never get called from an async path is not broken.

python/chemiscope/input.py
```diff
@@ -28,8 +28,12 @@ def create_input(
     Create a dictionary that can be saved to JSON using the format used by the default
     chemiscope visualizer.
 
-    :param list frames: list of atomic structures. For now, only `ase.Atoms`_,
-        `stk.BuildingBlocks`_, and `MDAnalysis.AtomGroup`_ objects are supported
+    :param list frames: list of atomic structures. These can be either `chemiscope`
+        compatible dictionaries, or `ase.Atoms`_, `stk.BuildingBlocks`_, and
+        `MDAnalysis.AtomGroup`_ objects. It is also possible to provide a list of
+        paths to `.json` files containing chemiscope-compatible structures. These
```
GardevoirX: "These" refers only to the list of paths? If so, to me it looks like a workaround, but I have no better ideas.

python/chemiscope/input.py
```diff
@@ -377,6 +381,44 @@ def create_input(
     return data
 
 
+def write_external_structures(frames, prefix="structure"):
```
GardevoirX: Might be nice if we could add an argument for `chemiscope.show`, say `disk_mode`. If it's set as `True`, we call this function and write the `.json`s into a `tmp` dir. Seems to be easier to use, but Idk if there's good way of handling (deleting) the `tmp` dir after the chemiscope widget is off.

python/chemiscope/input.py
```diff
@@ -377,6 +381,44 @@ def create_input(
...
+
+    json_frames = frames_to_json(frames)
+
+    if "data" in json_frames[0]:
+        raise ValueError(
+            "frames should be valid structures, but got external links instead"
+        )
+
+    user_frames = []
+    for i, frame in enumerate(json_frames):
```
GardevoirX, line 423: Can we parallelize this?

python/chemiscope/sphinx/utils.py
```diff
@@ -13,3 +15,63 @@ def copy_file(src_file, dst_file):
...
+
+    src_base = os.path.dirname(src_json)
+
+    for struct in structures:
+        if not isinstance(struct, dict):
+            continue
+
+        data = struct["data"]
+
+        # Case 1: data is a simple string: "frames/frame_0001.json"
```
GardevoirX: We have case 2?

python/chemiscope/structures/__init__.py
```diff
@@ -53,24 +92,30 @@ def frames_to_json(frames):
...
+    chemiscope-compatible dicts, `ase.Atoms`, `stk.BuildingBlocks`_, and
+    `MDAnalysis.AtomGroup`_ objects.
 
     :param frames: iterable over structures (typically a list of frames)
     """
     frames, adapter = _guess_adapter(frames)
 
-    if adapter == "ASE":
-        return [_ase_to_json(frame) for frame in frames]
+    json_frames = []
```
GardevoirX: Why this? To make the linter happy?

src/structure/grid.ts
```diff
@@ -652,37 +654,38 @@ export class ViewersGrid {
...
         structureIndex: number,
         atomIndex?: number
     ): Promise<void> {
-        return new Promise((resolve) => {
-            // Initialize load options with trajectory enabled
-            const options: Partial<LoadOptions> = {
-                trajectory: true,
-            };
+        //return new Promise((resolve, _error) => {
+        //    void _error; // explicitly mark as intentionally unused
```
GardevoirX: These two lines could be removed?

python/chemiscope/input.py
```diff
@@ -377,6 +381,44 @@ def create_input(
...
+
+
+    :param list frames: list of atomic structures in a format that can
+            be understood by `chemiscope`.
+    :param str prefix: prefix to use for each generated JSON filename. Files will be named like "{prefix}-0.json", "{prefix}-1.json" etc.
+
+    :return: list of paths to JSON files.
+    """
+
+    json_frames = frames_to_json(frames)
```
GardevoirX, line 415: sometimes I think it might be good to add a progress bar in this function, given that trajectories could be thousand-frame long and the processing time here could reach minutes

docs/src/python/sphinx.rst
ceriottm: I added some stats showing the file outputs and that the dataset is smaller

python/chemiscope/input.py
```diff
@@ -377,6 +381,44 @@ def create_input(
...
+
+    if "data" in json_frames[0]:
+        raise ValueError(
+            "frames should be valid structures, but got external links instead"
+        )
+
+    user_frames = []
+    for i, frame in enumerate(json_frames):
+        path = f"{prefix}-{i}.json"
+        json.dump(frame, open(path, "w"), indent=2)
```
GardevoirX: I tested with my data and one `.json` file takes 6.4 MB in the disk, shall we add an argument here to let it write `.json.gz`?

python/chemiscope/input.py
GardevoirX: Writing 1544 frames took it 7 min

GardevoirX (commented):
Nice! but issues like disk usage still exist

python/chemiscope/input.py
ceriottm: If structures are large, they are large. Main issue is with the memory load, disk space is cheap. Maybe one could allow (and generate) gzipped structures.

python/chemiscope/input.py
ceriottm: dee below why I think it's best to have this as a separate call

python/chemiscope/input.py
ceriottm: maybe could be made faster, but I don't think it can be parallelized. this is a good reason to have a preparatory job that can be done just once.

python/chemiscope/sphinx/utils.py
ceriottm: chatty is a bit verbose. I cleaned this up

python/chemiscope/structures/__init__.py
ceriottm: just to have a single exit point, the number of returns was getting silly

python/chemiscope/input.py
ceriottm: yes, working on it already

ceriottm: done

python/chemiscope/input.py
ceriottm: maybe frame_to_json can be made faster.... we don't have progress bars  anywhere I think so this might break something somewhere else. TBH I don't think chemiscope is the right tool to handle files that are >1GB in size, regardless of how they are stored and handled

python/chemiscope/structures/__init__.py
ceriottm: was just to avoid having like 8 return paths. it builds the json frames and returns them at the end of the function.

python/chemiscope/input.py
ceriottm: I have made the compression level adjustable, should make a smaller output, and if you set it to 1 or 2, should be a bit faster than before

python/chemiscope/sphinx/static/chemiscope-sphinx.js
```diff
@@ -32,13 +32,33 @@ async function loadChemiscopeSphinx(
...
+                if (!response.ok) {
+                    throw new Error(
+                        `Failed to fetch structure file "${url}": ${response.status} ${response.statusText}`
+                    );
+                }
+
+                return await response.json();*/
+            };
+        }
+
```
GardevoirX: Need some clean?

python/chemiscope/structures/_ase.py
```diff
@@ -207,9 +207,12 @@ def _ase_to_json(frame):
     data = {}
     data["size"] = len(frame)
     data["names"] = list(frame.symbols)
-    data["x"] = [float(value) for value in frame.positions[:, 0]]
-    data["y"] = [float(value) for value in frame.positions[:, 1]]
-    data["z"] = [float(value) for value in frame.positions[:, 2]]
+    # data["x"] = [float(value) for value in frame.positions[:, 0]]
+    # data["y"] = [float(value) for value in frame.positions[:, 1]]
+    # data["z"] = [float(value) for value in frame.positions[:, 2]]
```
GardevoirX: Clean?

docs/src/datasets/structures/frame-28.json
Luthaf, line 1: not sure why this is committed here?

python/chemiscope/input.py
```diff
@@ -377,6 +382,54 @@ def create_input(
     return data
 
 
+def write_external_structures(frames, prefix="structure", compresslevel=5):
```
Luthaf, line 385: Why are we writing to the disk? I was more thinking to implement this as a fully in-memory thing.

Either the data fits into Python memory (but not JS memory), and then we can load everything in Python, create "UserStructure" and send that to the JS side, while keeping the Python data for the frames around. Then when JS requests a frame, we can transform it to JSON and return just this one.

If the data does not even fit in Python memory, we could have a callback that fires when JS requests more data, reading the one frame we need from disk and converting it (although I'd say this is out of scope for this PR)

python/chemiscope/structures/_ase.py
```diff
@@ -207,9 +207,9 @@ def _ase_to_json(frame):
     data = {}
     data["size"] = len(frame)
     data["names"] = list(frame.symbols)
-    data["x"] = [float(value) for value in frame.positions[:, 0]]
-    data["y"] = [float(value) for value in frame.positions[:, 1]]
-    data["z"] = [float(value) for value in frame.positions[:, 2]]
+    data["x"] = frame.positions[:, 0].tolist()
+    data["y"] = frame.positions[:, 1].tolist()
+    data["z"] = frame.positions[:, 2].tolist()
```
Luthaf, line 212: The `float` conversion was intentional here, because we are not checking the dtype elsewhere

src/structure/viewer.ts
```diff
@@ -1943,7 +1947,7 @@ export class MoleculeViewer {
      * Reset the view by re-centering it and zooming to fit the model as much as
      * possible inside the views.
      */
-    private _resetView(): void {
+    public _resetView(): void {
```
Luthaf: why is this public?

docs/src/datasets/structures/frame-28.json
ceriottm: this is to demonstrate the sphinx directive

python/chemiscope/input.py
ceriottm: the whole idea was to make available in jupyter and sphinx the same mechanism we have for the chemiscope website, to load the structures dynamically

src/structure/viewer.ts
ceriottm: was experimenting w/something and forgot to change it back

python/chemiscope/structures/_ase.py
ceriottm: can they be anything else?  where is the problem if they are say single precision?

docs/src/datasets/structures/frame-28.json
Luthaf: but we don't need to commit files to the repository for that, if they are dynamically generated?

python/chemiscope/structures/_ase.py
Luthaf: I think `np.float64` this could cause some issues with JSON serialization. It might have been fixed, but I don't know when and if this is in a version we still support or not.

docs/src/datasets/structures/frame-28.json
ceriottm: they are dynamically generated for shinx-_gallery_ but not for the sphinx directive

python/chemiscope/input.py
Luthaf: For jupyter this makes sense, but I'm not so sure about sphinx. The advantage of jupyter is that we have a server that can do things and return the correct structure, while for sphinx the only solution is the one you picked here of writing files to the disk, which is not ideal for jupyter.

ceriottm:
> I'm not sure writing a bunch of files to the disk is the best course of action here, I would rather keep everything in memory as much as possible.

if stuff fits in memory, it fits in memory. if it doesn't, this provides functionality to load dynamically when needed. in practice this is pretty fast, there's no noticeable lag except for running a trajectory.

Luthaf:
My problem is more about requiring a separate step for dynamically loading instead of transparently making it work.

> if stuff fits in memory, it fits in memory.

System memory and JS memory are separate, depending on the browser it is hard to use more than ~4GiB of memory from JavaScript

python/chemiscope/structures/_ase.py
Luthaf: And I'm not sure if they can be something else, I'll try locally

python/chemiscope/input.py
ceriottm: I would first get something that can read from disk. Then if we want to add some caching python-side, it can be done easily in a separate PR. on the sphinx side, it serves the same purpose as the loader for the website, faster loading of a page.

ceriottm:
> My problem is more about requiring a separate step for dynamically loading instead of transparently making it work.
> 
> > if stuff fits in memory, it fits in memory.
> 
> System memory and JS memory are separate, depending on the browser it is hard to use more than ~4GiB of memory from JavaScript

OK I understand what you want to do. It is different from what this PR does, although I'd say it sets up much of the infrastructure. Let's merge this and do the python-side-caching-as-default in a separate PR, as it also needs handling carefully the saving side.

Luthaf:
> Let's merge this and do the python-side-caching-as-default in a separate PR, as it also needs handling carefully the saving side.

I'm not sure we are talking about the same thing. My suggestion would **not** do any caching or saving anything, but rather dynamically generate the requested structures.

If you think `chemiscope.write_external_structures` will be worth it long term, we can finish cleaning this up and merging (if we release it for 1.0, we have to keep maintaining the functionality for a while).

python/chemiscope/jupyter.py
```diff
@@ -63,6 +66,63 @@ def save(self, path):
...
+
+            except Exception as exc:
+                self.send(
+                    {
+                        "type": "load-structure-error",
+                        "requestId": request_id,
+                        "error": f"failed to load structure from '{filename}': {exc}",
+                    }
+                )
+            else:
```
Luthaf: why is this an `else` block?

python/chemiscope/jupyter.py
```diff
@@ -63,6 +66,63 @@ def save(self, path):
...
+                # `structure` must be a plain JSON-serialisable object that matches
+                # chemiscope's `Structure` interface on the JS side.
+                self.send(
+                    {
+                        "type": "load-structure-result",
+                        "requestId": request_id,
+                        "filename": filename,
```
Luthaf: I'm not sure why we have to send back the filename?

python/chemiscope/sphinx/utils.py
```diff
@@ -13,3 +15,61 @@ def copy_file(src_file, dst_file):
     dst_dir = os.path.dirname(dst_file)
     os.makedirs(dst_dir, exist_ok=True)
     shutil.copyfile(src_file, dst_file)
+
+
+def copy_external_structures(src_json: str, destination: str) -> None:
+    """Copy external structure JSON files referenced by `src_json`
+    into the same `_datasets` tree as `dst_json`, preserving relative paths.
+    """
```
Luthaf:
```suggestion
    """
    Copy external structure JSON files referenced by ``src_json``
    into the same ``_datasets`` tree as ``destination``, preserving relative paths.
    """
```

python/chemiscope/structures/__init__.py
Luthaf:
```suggestion
```

python/chemiscope/structures/_ase.py
Luthaf: Seem that numpy is now doing the right thing.

python/examples/1-base.py
```diff
@@ -64,6 +69,40 @@
...
+#
+# When working with large datasets, it is also possible to save separately the structure
+# data, and have chemiscope load them on demand. This can be done with a couple of
+# utility functions. Note that you will need to share the structure files alongside the
+# main dataset file, and that it will not be possible to use the standalone viewer
+# at `chemiscope.org`, as it requires all data to be included in the JSON file.
+
+# This will write the external structures as separate files `structure-*.json`
+external_frames = chemiscope.write_external_structures(
+    frames, "structure", compresslevel=True
```
Luthaf:
```suggestion
    frames, prefix="structure"
```

python/jupyter/src/widget.ts
```diff
@@ -92,6 +112,100 @@ class ChemiscopeBaseView extends DOMWidgetView {
...
+    protected _requestStructureFromPython(index: number, filename: string): Promise<Structure> {
+        const requestId = this._nextRequestId++;
+        return new Promise<Structure>((resolve, reject) => {
+            this._pendingStructureRequests.set(requestId, { resolve, reject });
+
+            this.model.send({
+                type: 'load-structure',
+                requestId,
+                index,
+                filename,
```
Luthaf: I think this should just send the `data` from UserStructure as-is (instead of sending "filename"), so we can change it later more easily

python/jupyter/src/widget.ts
```diff
@@ -92,6 +112,100 @@ class ChemiscopeBaseView extends DOMWidgetView {
...
+        config: DefaultConfig | StructureConfig,
+        data: Dataset
+    ): void {
+        const structures = data.structures;
+        if (!Array.isArray(structures) || structures.length === 0) {
+            return;
+        }
+
+        const sample = structures[0];
+        const hasUserData = sample && typeof sample === 'object' && 'data' in sample;
```
Luthaf:
```suggestion
        const structure = structures[0];
        const hasUserData = structure && typeof structure === 'object' && 'data' in structure;
```

src/index.ts
```diff
@@ -501,7 +501,12 @@ class DefaultVisualizer {
         if (getStructures) {
             copy.structures = [] as Structure[];
             for (let i = 0; i < this._dataset.structures.length; i++) {
-                copy.structures.push(this.structure.loadStructure(i, this._dataset.structures[i]));
+                const structure = this.structure.loadStructure(i, this._dataset.structures[i]);
+                // this code path should only be reached with the default loader, that is
+                // synchronous
```
Luthaf: I don't think this is the case, this function is public so anyone could call it with `getStructures=true`, even with completely custom loading callbacks.

ceriottm:
> If you think `chemiscope.write_external_structures` will be worth it long term, we can finish cleaning this up and merging (if we release it for 1.0, we have to keep maintaining the functionality for a while).

I think it's useful, even just for sphinx gallery and sphinx directive to reduce the loading time of the pages when datasets are large-ish.

python/chemiscope/jupyter.py
ceriottm: we can maybe make it optional? or call it structure_id? I think it's nice to have a bit of redundancy, but I can also take it out if you think it's useless.

python/chemiscope/jupyter.py
ceriottm: boh, why not. if loading does not fail then it sends the file. changed it, but I don't see the problem

src/index.ts
ceriottm: well, then we should make all the loadFunction calls be asynchronous. if you think this belongs in this PR, feel free to contribute it - it goes to change parts of chemiscope I do not understand and I have no time to understand.

python/jupyter/src/widget.ts
ceriottm: I renamed it in a more generic way, but we can't send any type, so I think this is it. Feel free to change it if you want something else, I don't understand what you want here

PR #450, opened 2025-11-19
Python side structures

ceriottm:
Caches structure data on the  python side to speed up loading widgets and to reduce JS memory footprint

### DIFF
--- a/python/chemiscope/jupyter.py
+++ b/python/chemiscope/jupyter.py
@@ -46,22 +46,30 @@ class ChemiscopeWidgetBase(ipywidgets.DOMWidget, ipywidgets.ValueWidget):
 
         # hold structures on the python side to save js memory
         self._structures_cache = {}
-        if cache_structures and "data" not in data["structures"][0]:
-            structure_refs = []
-            for i, structure in enumerate(data["structures"]):
-                if "data" in structure:
-                    raise ValueError("Mixed explicit and external structures in frames")
-                self._structures_cache[f"structure-{i}"] = structure
-                structure_refs.append(
-                    {"size": structure["size"], "data": f"structure-{i}"}
-                )
-            data["structures"] = structure_refs
+        if cache_structures:
+            data["structures"] = self._cache_structures(data["structures"])
 
         # pass data to javascript through a traitlet
         self.value = json.dumps(data)
 
         # listen for custom messages from the JS side
         self.on_msg(self._handle_js_msg)
+
+    def _cache_structures(self, structures):
+        """Cache full structures and return structure references"""
+        if not structures or "data" in structures[0]:
+            return structures
+
+        structure_refs = []
+        for i, structure in enumerate(structures):
+            if "data" in structure:
+                raise ValueError("Mixed explicit and external structures in frames")
+
+            cache_key = f"structure-{i}"
+            self._structures_cache[cache_key] = structure
+            structure_refs.append({"size": structure["size"], "data": cache_key})
+
+        return structure_refs
 
     def save(self, path):
         """


ceriottm:
@Luthaf  I think this is what you had in mind. works well, it is slower in loading frames so it has a tendency to hang if you use a small (<20ms) interval for the frames. I'm thinking we should anyway put a floor to 50ms, as that's basically a 20fps framerate. Maybe one should not allow, as in the current implementation, to queue multiple requests, but block if there is another request pending. I'll try that.

ceriottm:
OK, ready for review. I implemented a mechanism to skip frames that works very well. Structure caching defaults to true but can be disabled.

GardevoirX:
That's good. For my 3577 atoms trajectory, I saw a speed up in `chemiscope.show` from 13s to 8s (100 frames) and from 1m3s to 46s (500 frames)

ceriottm:
> That's good. For my 3577 atoms trajectory, I saw a speed up in `chemiscope.show` from 13s to 8s (100 frames) and from 1m3s to 46s (500 frames)

Nice. Could you try to use frames_to_json before calling chemiscope.show? I suspect that most of the remaining delay is from converting the structures from ase/mda to json. Maybe that can be accelerated a bit too, especially on the mda side.

python/jupyter/src/widget.ts
```diff
@@ -167,14 +168,23 @@ class ChemiscopeBaseView extends DOMWidgetView {
...
         return new Promise<Structure>((resolve, reject) => {
-            this._pendingStructureRequests.set(requestId, { resolve, reject });
-
-            this.model.send({
-                type: 'load-structure',
-                requestId,
-                index,
-                data: structure.data as JSONValue,
-            });
+            if (this._pendingStructureRequests.size > 0) {
```
sofiia-chorna, line 171: I am afraid this skips ALL structure requests whenever ANY request is pending, plus, I suppose promises never resolve when requests are skipped...

python/jupyter/src/widget.ts
```diff
@@ -167,14 +168,23 @@ class ChemiscopeBaseView extends DOMWidgetView {
...
+                // avoid piling up too many requests
+                // eslint-disable-next-line no-console
+                console.warn(
+                    `Skipping frame ${index} - ${structure.data}. Increase playback delay.)`
```
sofiia-chorna:
```suggestion
                    `Skipping frame ${index} - ${structure.data}. Increase playback delay.`
```

python/chemiscope/jupyter.py
```diff
@@ -33,16 +33,33 @@ class ChemiscopeWidgetBase(ipywidgets.DOMWidget, ipywidgets.ValueWidget):
...
         # timeout for warning messages (ms). 0 to make persistent, -1 to disable
         self.warning_timeout = warning_timeout
 
+        # hold structures on the python side to save js memory
+        self._structures_cache = {}
+        if cache_structures and "data" not in data["structures"][0]:
+            structure_refs = []
+            for i, structure in enumerate(data["structures"]):
+                if "data" in structure:
+                    raise ValueError("Mixed explicit and external structures in frames")
```
sofiia-chorna: if the 5th structure has "data" in it, structures 0-4 will already be cached before the error is raised. properly, we should validate the entire array first before caching anything... but idk if it might introduce some overhead

python/chemiscope/jupyter.py
```diff
@@ -182,6 +217,8 @@ def show_input(path, mode="default", warning_timeout=10000):
     :param str mode: widget mode, either ``default``, ``structure`` or ``map``.
     :param float warning_timeout: timeout (in ms) for warnings. Set to a negative value
         to disable warnings, and to zero to make them persistent.
+    :param bool cache_structures: whether to cache structure data on the Python side
```
sofiia-chorna, line 226: we can add when to set it to False (e.g. when structures are in external files?) + add that cached structures remain in python memory for the widget's lifetime. Same comment for `show` method

python/chemiscope/jupyter.py
```diff
@@ -62,6 +79,11 @@ def save(self, path):
         # update the settings in the data to the latest value
         data = json.loads(self.value)
         data["settings"] = self.settings
+        if len(self._structures_cache) > 0:
+            # there are cached structures, add them to the dump
+            data["structures"] = [
+                structure for structure in self._structures_cache.values()
+            ]
```
sofiia-chorna: maybe just
```suggestion
            data["structures"] = list(self._structures_cache.values())
```

python/chemiscope/jupyter.py
ceriottm: Is that a problem? It will raise the error and not create the object, that would invalid anyway

python/jupyter/src/widget.ts
ceriottm: so, this will skip requests unless they've been already processed. this is intended, because otherwise they pile up and hang the browser. IDK about the unresolved promises, empirically this works perfectly fine... can you actually try, and if you see a better way to implement this, feel free.

python/chemiscope/jupyter.py
ceriottm: if structures are in external files, they remain in external files already now, there's no need to set this to false. do we need to mention about the python memory? if it's set to false, they will remain in the JS memory anyway

sofiia-chorna: I was wondering whether there are cases where a user might want (or need) to set `cache_structures=False`. If not, then caching could simply be enabled by default and the argument wouldn’t be necessary

python/jupyter/src/widget.ts
sofiia-chorna: ok ok, interested to see if it could be make better

python/chemiscope/jupyter.py
sofiia-chorna: Ok, not really a problem

python/chemiscope/jupyter.py
ceriottm: as far as I can tell, the main reason for setting cache_structures=False is if you want to a smooth-running trajectory, as there is some extra latency associated with data transfer. I think it is not a common scenario, which is why I make the default True, but seems a small enough feature overhead that we can leave the option there

python/jupyter/src/widget.ts
sofiia-chorna: did not manage to break it, seems to work fine. unless GF has some concerns, I'd leave it like this

sofiia-chorna:
I am adding a small refactor from my side if that's ok

GardevoirX:
> That's good. For my 3577 atoms trajectory, I saw a speed up in `chemiscope.show` from 13s to 8s (100 frames) and from 1m3s to 46s (500 frames)

Glad to see that now it takes only 0.9s to show this 500 frames trajectory, and we are able to show 2500 frames trajectory with 53650 atoms in 44.2s. I guess it's enough for most of the use cases

sofiia-chorna:
@GardevoirX thanks for checking and thanks @ceriottm for implementation

PR #453, opened 2025-11-21
Fix sorting of examples in documentation

sofiia-chorna:
This PR updates the docs configuration so that examples are sorted based on the leading number in their names (i.e. examples/10_* appears after examples/9_*)


ceriottm:
LGTM

PR #457, opened 2025-11-24
Make all loading be properly async

### LINKED ISSUE #456: The "Atom property coloring" example is broken
Currently unclear about which update leads to this error. See https://chemiscope.org/docs/examples/4-colors.html#atom-property-coloring
Image

ceriottm:
Fixes #456  and probably is saner.

### DIFF
--- a/src/structure/grid.ts
+++ b/src/structure/grid.ts
@@ -131,7 +131,7 @@ export class ViewersGrid {
     /// List of structures in the dataset
     private _structures: Structure[] | UserStructure[];
     /// List of per-atom properties in the dataset
-    public _properties: Record<string, number[]>;
+    private _properties: Record<string, number[]>;
     /// Widget display target (structure or atom)
     private _target: DisplayTarget;
     /// Cached string representation of structures


ceriottm:
> It worked. Can we try to capture these kinds of errors in the CI? It seems like we can only capture the errors at the python side but have no ideas about if the typescript works and how it would finally look like in the jupyter widget

I think we tried to have something at some point but it's very hard to test what the browser (or jupyter) renders

src/structure/grid.ts
```diff
@@ -131,7 +131,7 @@ export class ViewersGrid {
     /// List of structures in the dataset
     private _structures: Structure[] | UserStructure[];
     /// List of per-atom properties in the dataset
-    private _properties: Record<string, number[]>;
+    public _properties: Record<string, number[]>;
```
Luthaf:
```suggestion
    private _properties: Record<string, number[]>;
```

ceriottm: well spotted!

PR #458, opened 2025-11-26
Update version numbers for 0.8.8 release

ceriottm:
Bugfix release for #456


Luthaf:
This should be 0.8.8, not worth adding an extra number!

PR #459, opened 2025-11-26
Info panel and a bunch of papercut fixes

ceriottm:
* improve ergonomics of the property plotter (cf. amorphous silicon example
* added a reference to atomlabels in the examples
* scroll bars for mobile

### DIFF
--- a/src/static/chemiscope.css
+++ b/src/static/chemiscope.css
@@ -249,7 +249,7 @@
 .chsp-info-plot-holder {
     width: 80%;
     margin: 0 auto;
-    aspect-ratio: 4 / 3;
+    aspect-ratio: 16 / 9;
 }
 
 .chsp-settings-section-title {


ceriottm:
Ok maybe best to close this here. Ended up being a bunch of small fixes, some consequences of problems created by a previous change that I did (oops), some of small bugs I discovered while using CS

ceriottm:
Thanks @sofiia-chorna for the review. I think I addressed everything, if you are happy I think this can be merged, and maybe a new release can be made since it does fix a lot of annoying bugs, some new some old. 

> * When loading a saved chemiscope file where _opacity_ was previously changed, the opacity value does not seem to be applied on load

fixed

> * Property values extend beyond the panel boundaries. maybe we should use `text-overflow: ellipsis` to truncate long values... or at least have it working as in current main

fixed it and also limit the number of decimals, as it's really cringey to show 20 decimals 

> * (small) this panel appears slightly off-center on my side

I think I fixed this, and also made the panel resize automatically. pls check it works also on your side

> Also, is it OK that when changing color palette in structure viewer, atom labels color is not updated? 

I would say so. the text coloring still reflects the nature of the atoms, which is additional, and potentially useful, information

src/static/chemiscope.css
```diff
@@ -233,8 +233,11 @@
 
 .chsp-properties-table {
     height: 100%;
-    width: 80%;
+    width: 90%;
     margin: auto;
+    overflow-x: auto;
```
sofiia-chorna:
```suggestion
```

from my tests, I noticed that opening the info panel started to lag a bit, after this line was added (it is not instant as currently in main). since we now limit numeric values to 4 characters after the point, overflow is no longer really an issue. so I propose removing this line

src/utils/index.ts
```diff
@@ -171,3 +171,46 @@ export function binarySearch(array: number[], value: number): number {
 
     return -1;
 }
+
+/** Try to format a float to ensure it fits in a fixed characters limit */
+export function fixedWidthFloat(value: number, width: number = 4): string {
+    if (value === 0) {
+        return '0';
+    } else if (value > 0) {
+        // we can use the full 'width; characters for the number
```
sofiia-chorna:
```suggestion
        // we can use the full 'width' characters for the number
```

sofiia-chorna:
Fixes work on my side, thanks!

PR #464, opened 2025-12-03
Fix a corner case where string properties would be discarded from an …

ceriottm:
…ASE properties list

This was detected by a cookbook example

### DIFF
--- a/python/tests/ase_structures.py
+++ b/python/tests/ase_structures.py
@@ -154,6 +154,7 @@ class TestExtractProperties(unittest.TestCase):
 
         frame_2 = BASE_FRAME.copy()
         frame_2.info["valid_prop"] = [3, 4]
+        # this will be turned into a string because it is a string in the other frame
         frame_2.info["valid_string"] = 1345
         frame_2.info["invalid_prop"] = [1]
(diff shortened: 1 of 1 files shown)


sofiia-chorna:
Looks good but I think we also need to add it to `_ase_get_atom_properties`?

ceriottm:
added a test. the validation is only necessary for structure terms, if there are arrays they should be already homogeneous.

python/tests/ase_structures.py
```diff
@@ -148,11 +148,13 @@ def test_wrong_property_type(self):
     def test_different_lengths(self):
         frame_1 = BASE_FRAME.copy()
         frame_1.info["valid_prop"] = [1, 2]
+        frame_1.info["valid_string"] = "hello"
         frame_1.info["invalid_prop"] = [1, 2]
         frame_1.info["non_iter_prop"] = 3
 
         frame_2 = BASE_FRAME.copy()
         frame_2.info["valid_prop"] = [3, 4]
+        frame_2.info["valid_string"] = 1345
```
Luthaf, line 158:
```suggestion
        # this will be turned into a string because it is a string in the other frame
        frame_2.info["valid_string"] = 1345
```

Issue #460, opened 2025-12-01
Error when using lammpstrj in MDAnalysis


PicoCentauri:
I have a LAMMPS trajectory that I would like to visualise. However, I get an error that almost crashes my notebook because it is that long and not atoms will be displayed.

<img width="1124" height="501" alt="Image" src="https://github.com/user-attachments/assets/62527106-240e-4999-a454-219dd534e9b2" />

The code is fairly simple:

```python
import chemiscope
import MDAnalysis as mda

u = mda.Universe("system.data", "dump_NVT.lammpstrj", format="LAMMPSDUMP")
chemiscope.show(u.atoms, mode="structure")
```

[traj.zip](https://github.com/user-attachments/files/23853669/traj.zip)

I know that we can't deduce the correct atom colors because LAMMPS has integer elements, but in this case I would probably just use different colors based on the different elements.

ceriottm:
thanks @PicoCentauri . can you liaise with @GardevoirX ? this is probably a very easy fix, maybe it can make it into #459 , or into a simple one-purpose PR that can be merged quickly before we make a new 0.8.9 release

GardevoirX:
The error here means that chemiscope didn't collect enough information to deduce the secondary structure. I'll let it automatically fallback to the normal ball-and-stick representation and disable cartoon in that case

GardevoirX:
I took a look at this and I don't think it can be done easily. Though it might be able to color atoms based on different element names, the current bond-assigning function only works when the correct element names are given. ofc the bonding information is given in the topology file, but I may need some time for me to understand the data structure of bonding information in `3DMol`.

Though there's the error, I think we can release 0.8.9 without this fix, since one can still load this trajectory with `ase` and chemiscope can show it @ceriottm 

And below is the current status if I ignore the warning

<img width="548" height="532" alt="Image" src="https://github.com/user-attachments/assets/da2bba01-5baf-4f5a-80d4-6e8f23131f85" />

PicoCentauri:
> The error here means that chemiscope didn't collect enough information to deduce the secondary structure. I'll let it automatically fallback to the normal ball-and-stick representation and disable cartoon in that case

Ahh, yes I think this is a good idea.

> Though it might be able to color atoms based on different element names, the current bond-assigning function only works when the correct element names are given. ofc the bonding information is given in the topology file, but I may need some time for me to understand the data structure of bonding information in 3DMol.

[...]

PR #465, opened 2025-12-04
Utilizing the bonding information of mda

### LINKED ISSUE #461: Directly use the bonding information of `MDAnalysis`
We currently guess bonds instead of reading bond information. With `MDAnalysis` we are now entering a field where the topology of the system is pre-defined. Thus, we shall be also able to read the bond parsed by `MDAnalysis`, and use it whenever it's available

GardevoirX:
closes #461 , the system mentioned in #460 now looks like this
image

One thing that I am not very clear is this check:
https://github.com/GardevoirX/chemiscope/blob/004c911c1d6b79b0ae40f9754871935ff27e02d2/src/dataset.ts#L582-L584
This was introduced to check if the information from mda is enough for cartoon representation, but now we can also use mda to do non-bio things, what shall we do to this check?

### DIFF
--- a/python/chemiscope/structures/_mda.py
+++ b/python/chemiscope/structures/_mda.py
@@ -10,6 +10,8 @@ try:
 except ImportError:
     HAVE_MDA = False
 
+BIO_PROPERTIES = ["hetatom", "resids", "resnames", "chains"]
+
 
 def _mda_valid_structures(frames):
     if HAVE_MDA and isinstance(frames, mda.AtomGroup):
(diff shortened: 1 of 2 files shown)


Luthaf:
> One thing that I am not very clear is this check: https://github.com/GardevoirX/chemiscope/blob/004c911c1d6b79b0ae40f9754871935ff27e02d2/src/dataset.ts#L582-L584 This was introduced to check if the information from mda is enough for cartoon representation, but now we can also use mda to do non-bio things, what shall we do to this check?

We should keep the check around, because the bio molecules information could be set by other providers than MDA. We should also change the output of MDA to make sure it passes this check

GardevoirX:
so, we shall clean the data extracted from `AtomGroup` if the data are not enough to pass the check at the typescript side?

Luthaf:
I'd say so yes

PR #470, opened 2025-12-19
Fix the shape loader for structures in the jupyter widget

ceriottm:


Luthaf:
Closing in favor of #471

PR #425, opened 2025-07-10
Clean up docs

sofiia-chorna:
The idea is to check the examples in the documentation are actual & do clean up

PR #471, opened 2025-12-19
More fixes

Luthaf:
Replaces #470 and fix JSON reference after #425

### DIFF
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -127,8 +127,10 @@ export interface Structure {
      */
     bonds?: Array<[number, number, number]>;
     /**
+     * @private
+     *
      * possible shapes to display, multiple groups of shapes with different
-     * names are allowed
+     * names are allowed. This is set by `assignShapes` when loading the dataset.
      */
     shapes?: {
         /**
@@ -176,6 +178,15 @@ export interface UserStructure {
      * data.
      */
     data: unknown;
+
+    /**
+     * @private
+     * Dictionary containing shape data, set by `assignShapes` when loading the
+     * dataset.
+     */
+    shapes?: {
+        [name: string]: ShapeParameters;
+    };
 }
 
 /** Possible types of properties: full structure property, or atomic property */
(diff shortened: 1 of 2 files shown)


ceriottm:
OK, cleaner than chatty. It does have tendency to do unnecessary type gymnastics.
LGTM and seems to work but I'll leave Sofiia look at the documentation fixes

docs/src/getting-started/json-format.rst
```diff
@@ -189,31 +262,37 @@ Value types and shapes
...
      - Length must match ``n_atoms`` (for ``target="atom"``) or ``n_structures`` (for
        ``target="structure"``).
      - ``[0.1, 0.2, 0.3]``
    * - **Multi-dimensional**
-     - 2D ndarray (``[n_samples, n_components]``)
-     - Generates multiple properties (e.g., ``PCA[1]``, ``PCA[2]``)
+     - Array of arrays of numbers
+     - Used for parameterized properties; outer array length must match
+       ``n_atoms`` or ``n_structures``, inner arrays length must match linked
+       parameter length.
```
sofiia-chorna: Maybe we can get rid of ``n_atoms`` or ``n_structures``, something like:
```suggestion
     - Used for parameterized properties; outer array length must match
       the number of atoms (for atom-level properties) or the number of structures (for structure-level properties), inner arrays length must match linked
       parameter length.
```

docs/src/getting-started/json-format.rst
```diff
@@ -835,43 +743,98 @@ Each parameter is a dictionary with the following fields:
...
 
-.. _ase: https://wiki.fysik.dtu.dk/ase/index.html
+.. code-block:: json
+
+   "settings": {
+     "target": "structure",
+     "map": {"joinPoints": true},
+     "structure": [{
+       "keepOrientation": true,
+       "playbackDelay": 50
```
sofiia-chorna, line 817: It seems we did not document ``playbackDelay`` in the table above...

ceriottm:
@Luthaf  you need to enable "back-pushing" to your fork so we can synchronize this with main.

Luthaf:
I prefer to synchronize myself, to keep a nice git history!

sofiia-chorna (approved):
Nice! Thanks

PR #472, opened 2025-12-24
Adaptive resolution for plotly maps

ceriottm:
Implement level-of-detail sparsification of datasets to avoid performance degradation of plotly with > 100k points

PR #473, opened 2025-12-25
Update to plotly 3.0

ceriottm:
Update to plotly 3.0, that is supposedly faster and allowed removing a workaround for plotly within a shadow DOM.


sofiia-chorna (approved):
nice, it works fine too! thanks a lot for taking care of it

ceriottm:
This actually introduces a small bug with axes labels, so I'd just merge #472  that also includes a fix for that

PR #474, opened 2025-12-28
Allow to override visualization settings from show_input

ceriottm:
It's handy to be able to load an existing chemiscope from jupyter while changing the visualization settings

### DIFF
--- a/python/chemiscope/jupyter.py
+++ b/python/chemiscope/jupyter.py
@@ -287,7 +287,9 @@ def show_input(
 
     if settings is not None:
         if not isinstance(settings, dict):
-            raise ValueError(f"expected 'settings' to be a dict, got {type(settings)} instead")
+            raise ValueError(
+                f"expected 'settings' to be a dict, got {type(settings)} instead"
+            )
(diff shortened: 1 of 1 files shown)


python/chemiscope/jupyter.py
```diff
@@ -282,6 +285,9 @@ def show_input(path, *, mode="default", warning_timeout=10000, cache_structures=
     except KeyError:
         raise ValueError("missing metadata in chemiscope.load")
 
+    if settings is not None:
+        dict_input["settings"].update(settings)
```
sofiia-chorna, line 295: maybe we can add some validations like:
```suggestion
    if settings is not None:
        if not isinstance(settings, dict):
            raise ValueError(f"expected 'settings' to be a dict, got {type(settings)} instead")
        if "settings" not in dict_input:
            dict_input["settings"] = {}
        dict_input["settings"].update(settings)
```
(linter might fall if suggestion is applied)

sofiia-chorna (approved):
Looks good, thanks!

PR #475, opened 2025-12-28
Fix race condition when loading a dataset with structure color settings

ceriottm:
Loading a dataset with structure coloring or shapes can sometimes lead to crashes or to flickering back to default settings. This PR fixes the problem.

### DIFF
--- a/src/structure/viewer.ts
+++ b/src/structure/viewer.ts
@@ -383,11 +383,6 @@ export class MoleculeViewer {
      * @param envView display target, true if per environments
      * @param propertiesName property names used as the options in the modal
      */
-    /**
-     * Recreates structure options
-     * @param envView display target, true if per environments
-     * @param propertiesName property names used as the options in the modal
-     */
     public refreshOptions(envView: boolean, propertiesName?: string[]): void {
         // This prevents the "flicker" where settings revert to defaults after a refresh
         const previousSettings = this._options.saveSettings();


src/structure/viewer.ts
```diff
@@ -378,12 +378,20 @@ export class MoleculeViewer {
         this._resetView();
     }
 
+    /**
+     * Recreates structure options
+     * @param envView display target, true if per environments
+     * @param propertiesName property names used as the options in the modal
+     */
```
sofiia-chorna: seems to be a duplication
```suggestion
```

src/structure/viewer.ts
```diff
@@ -985,6 +995,10 @@ export class MoleculeViewer {
         // ======= color settings
         // setup state when the property changes
         const colorPropertyChanged = () => {
+            if (this._properties === undefined) {
```
sofiia-chorna, line 993: I don't understand in which case `this._properties` can be undefined, since it's only initialized in the load method. Do you have steps to reproduce this error? Anyways, for now I haven't been able to break it

ceriottm: This happens because applySettings is called before the structure data is loaded (and so, before this._properties is defined). When applySettings sets the color property, the change listener fires immediately, tries to access this._properties to calculate data ranges, and crashes because it is undefined

PR #480, opened 2026-01-06
fix(map): better handling of buttons in 3d mode

### LINKED ISSUE #477: Zoom selection button is not working in 3D mode (works in 2D)
In 3D mode, when zoom button is selected, zoom-in selection window is not displayed, so it is not possible to select a region to zoom into in 3D. I would expect it to work the same way as in the 2D mode.

Steps to reproduce:
1. Open any example
2. Select z-axis in the map settings
3. Click on the zoom button
4. Try to select a region on the map to zoom in

This might a bug on the plotly.js side. If this is true, we could probably disable or hide the zoom button in 3D mode

ceriottm:
Fix #477


sofiia-chorna (approved):
Looks good, thanks!

PR #482, opened 2026-01-07
Feature/jupyter selection

ceriottm:
We tried and failed a while ago to make structure selection accessible from jupyter with a traitlet interface. This is a new attempt, that works.

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -25,6 +25,49 @@ import * as styles from '../styles';
 import PNG_SVG from '../static/download-png.svg';
 import SVG_SVG from '../static/download-svg.svg';
 
+/**
+ * Export the plot as a PNG or SVG image, hiding the "selected" trace (index 1)
+ * during the export.
+ *
+ * @param gd Plotly plot element
+ * @param format format of the image
+ */
+function exportImage(gd: PlotlyScatterElement, format: 'png' | 'svg') {
+    const width = Math.max(gd._fullLayout.width, 600);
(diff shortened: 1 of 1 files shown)


ceriottm:
this is very similar to the implementation in the streamlit applet

sofiia-chorna:
I do this:
```
import ase.io
import chemiscope

frames = ase.io.read("showcase.xyz", ":")

indexes = [i for i in range(len(frames))]
cs = chemiscope.show(frames, properties={"test 1": indexes, "test 2": indexes})
cs
```
`print(cs.selected_ids) => {}`

change the index in the viewer

`print(cs.selected_ids) => still {}`

I don't understand why it does not work for me. On your side all good?

ceriottm:
> I do this:
> 
> ```
> import ase.io
> import chemiscope
> 
> frames = ase.io.read("showcase.xyz", ":")
> 
> indexes = [i for i in range(len(frames))]
> cs = chemiscope.show(frames, properties={"test 1": indexes, "test 2": indexes})
> cs
> ```
> 
> `print(cs.selected_ids) => {}`
> 
> change the index in the viewer
> 
> `print(cs.selected_ids) => still {}`
> 
> I don't understand why it does not work for me. On your side all good?

did you pip install and refresh the page in the browser?

ceriottm:
ok I'm closing this because I found another bug that needs to be fixed first. the setting sync does not work anymore, IDK when it broke.

PR #483, opened 2026-01-07
Switch to zero-based indexing in UI

ceriottm:
indexing  is internally zero-based but the UI displays it as one based. while it might be more intuitive, it gets confusing when working in jupyter, or in general when using this side-by-side with python tools that are zero based. 
given we are doing breaking changes with 1.0, this is a good time to move the UI to a more pythonic zero base


sofiia-chorna (approved):
Okay, seems everything is applied correctly, from my testing it works fine. Thanks!

PR #485, opened 2026-01-07
Sync structure selection between jupyter and viewers

ceriottm:
* Makes it possible to manipulate the selected structure ID from jupyter
* Also allows manipulation of multiple viewers grid.
* Fixes several small-but-nasty corner-case bugs triggered when switching from atom and structure view in hybrid datasets, and simplifies the logic for parameter synchronization, esp. for playbackdelay that was getting disconnected from the actual property value. 
* Adds documentation for the traitlets, that was already missing but is now even more important.

### DIFF
--- a/docs/src/python/jupyter.rst
+++ b/docs/src/python/jupyter.rst
@@ -20,9 +20,11 @@ Creating a chemiscope widget
 Widget properties and methods
 -----------------------------
 
-The widget object returned by :func:`chemiscope.show` and 
-:func:`chemiscope.show_input` provides traitlets 
-to interact with the visualization state programmatically.
+The widget object returned by :func:`chemiscope.show` provides traitlets to
+interact with the visualization state programmatically. For more information
+on using traitlets and widget events (like ``observe``), please refer to 
+the `ipywidgets documentation
+<https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Events.html>`_.
 
 .. py:attribute:: settings
 


python/jupyter/src/widget.ts
```diff
@@ -168,7 +398,7 @@ class ChemiscopeBaseView extends DOMWidgetView {
     ): Promise<Structure> {
         const requestId = this._nextRequestId++;
         return new Promise<Structure>((resolve, reject) => {
-            if (this._pendingStructureRequests.size > 0) {
+            if (this._pendingStructureRequests.size >= 10) {
```
sofiia-chorna: was it intentionally changed to 10?

python/chemiscope/jupyter.py
```diff
@@ -30,6 +30,10 @@ class ChemiscopeWidgetBase(ipywidgets.DOMWidget, ipywidgets.ValueWidget):
     settings = Dict().tag(sync=True)
     # timeout for warning messages
     warning_timeout = Int(10000).tag(sync=True)
+    # currently selected environment/structure/atom
```
sofiia-chorna, line 33: we don't display "environment id" (general global index), only structure and atom
```suggestion
    # currently selected structure/atom
```

python/jupyter/src/widget.ts
ceriottm: yes,  otherwise it triggers a lot of spurious events

ceriottm: good to test what happens if you have large structures and ask for short delay

python/chemiscope/jupyter.py
ceriottm: I have also enabled setting atom index and structure separately, and to keep the same atom selected when switching structure that might be useful when looking at trajectories with environments

python/jupyter/src/widget.ts
```diff
@@ -80,14 +103,43 @@ class ChemiscopeBaseView extends DOMWidgetView {
...
+                    Array.isArray(settings.pinned)
+                ) {
+                    const target = this.visualizer.saveSettings().target as DisplayTarget;
+                    const currentPinned = this.visualizer.structure
+                        .pinned()
+                        .map((value) => (target === 'atom' ? value.environment : value.structure));
+
+                    const pinned = settings.pinned as number[];
+                    const pinnedChanged =
+                        pinned.length !== currentPinned.length ||
```
sofiia-chorna: started to have this error on load in console:
Screenshot 2026-01-08 at 20 08 02

i suppose it comes from this.visualizer.structure.pinned() being undefined

src/info/info.ts
```diff
@@ -243,14 +239,31 @@ export class EnvironmentInfo {
...
+                        }
+
+                        /* Update the display with the details of the found structure
                         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
                         this.show(this._indexer.fromStructure(structure, this._target)!);
                         this.onchange(this._indexes());
 
                         // Recursively call startPlayback to continue playback
-                        slider.startPlayback(advance);
+                        slider.startPlayback(advance); */
```
sofiia-chorna: should we delete the comment?

sofiia-chorna (commented):
some small things i noticed:
- there is an error in the console for the Chemical Shielding example (but it seems to be working fine anyways)
[dataset.json](https://github.com/user-attachments/files/24506940/dataset.json)

Screenshot 2026-01-08 at 20 34 08

- when we set an id outside the valid range, the info panel is updated with this number. we can live with it but perhaps it is not perfect

python/jupyter/src/widget.ts
sofiia-chorna: tried with the examples from https://chemiscope.org/docs/examples/10-visualizing-biomolecules.html, seems to be working fine

docs/src/python/jupyter.rst
```diff
@@ -17,6 +17,55 @@ Creating a chemiscope widget
 .. autofunction:: chemiscope.show
 .. autofunction:: chemiscope.show_input
 
+Widget properties and methods
+-----------------------------
+
+The widget object returned by :func:`chemiscope.show` and 
+:func:`chemiscope.show_input` provides traitlets
```
Luthaf: should this link to the traitlets documentation in jupyter? What traitlets are might not be clear to everyone

python/jupyter/src/widget.ts
```diff
@@ -80,20 +103,268 @@ class ChemiscopeBaseView extends DOMWidgetView {
...
+                        atom = currentIndexes.atom;
+                    }
+
+                    // Validate atom index
+
+                    if (atom !== undefined) {
+                        const activeAtoms = this.visualizer.indexer.activeAtoms(structure);
+
+                        if (activeAtoms && activeAtoms.length > 0) {
+                            if (binarySearch(activeAtoms, atom) === -1) {
```
Luthaf, line 299: why is this a binary search, instead of some standard search on an array?

ceriottm: it is how all this was already implemented (binarysearch was introduced 4 years ago ^_^) I assumed it was because of the sparsity of the environment array.

docs/src/python/jupyter.rst
ceriottm: Done!

PR #487, opened 2026-01-09
Save images of the structures and map from jupyter

ceriottm:
Bit clunky because of javascript/jupyter limitations, but can be quite useful to generate snapshots for figures.

### DIFF
--- a/python/chemiscope/jupyter.py
+++ b/python/chemiscope/jupyter.py
@@ -314,6 +314,39 @@ class ChemiscopeWidgetBase(ipywidgets.DOMWidget, ipywidgets.ValueWidget):
 
     def _handle_js_msg(self, _, content, buffers):
         """Handle custom messages sent from the JS widget.
+        The pattern is that a message is sent from another
+        function, and the response is collected here, handling
+        a successful response (`msg_type="base-mst-result"`)
+        or and error (`msg_type="base-mst-result"`). See below
+        for a brief explanation of the different types of messages.
+
+        Currently implemented messages:
+
+        * "load-structure": Initiated on the JS side, used to
+            retrieve an atomic geometry from disk or from a
+            python-side cache.
+
+            {"type": "load-structure",
+             "requestId": int,
+             "data": "filename-or-id"}
+
+        * "save-image": Initiated on the Python side, used to get
+            a snapshot of the map or of the active structure
+
+            {"type": "save-image",
+             "requestId": int,
+             "target": "map" | "structure"}
+
+        * "save-structure-sequence": Initiated on the Python side, used to
+            generate snapshots for several atomic configurations.
+            Should provide indices (in the same format as for `selected_ids`)
+            and optionally structure settings as a list of dicts.
+
+            {"type": "save-structure-sequence",
+             "requestId": int,
+             "indices": [{"structure": id1[, "atom": id2]}, ...],
+             ["settings": [settings_1_dict, ...]]
+             }
 
         Expected messages:
 


docs/src/python/jupyter.rst
```diff
@@ -62,12 +62,86 @@ the `ipywidgets documentation
...
+There are two types of methods: ``save_*`` methods, which write the image data 
+directly to a file given the path, and ``get_*`` methods, which return the raw 
+image bytes (PNG formatted).
+
+.. note::
+    These methods return ``asyncio.Future`` objects. Depending on the environment,
+    environments, you may be able to ``await`` directly in a cell to wait for the 
+    result. If ``await`` hangs in your environment, you can use 
+    catch the ``get_*`` return value, and then in a separate cell get `.result()` 
+    to access the actual return value.
```
sofiia-chorna, line 92:
```suggestion
    These methods return ``asyncio.Future`` objects. Depending on the environment,
    you may be able to ``await`` directly in a cell to wait for the 
    result. If ``await`` hangs in your environment, you can use 
    catch the ``get_*`` return value, and then in a separate cell get `.result()` 
    to access the actual return value.
```

python/chemiscope/jupyter.py
```diff
@@ -106,10 +294,50 @@ def _handle_js_msg(self, _, content, buffers):
...
         """
         msg_type = content.get("type", None)
+        request_id = content.get("requestId")
+
+        if msg_type == "save-image-result":
+            pending = self._pending_requests.pop(request_id, None)
+            if pending and isinstance(pending, asyncio.Future):
+                data = content.get("data")
+                try:
+                    header, encoded = data.split(",", 1)
```
sofiia-chorna:
```suggestion
                    _header, encoded = data.split(",", 1)
```

python/chemiscope/jupyter.py
```diff
@@ -106,10 +294,50 @@ def _handle_js_msg(self, _, content, buffers):
...
+
+        elif msg_type == "save-structure-sequence-result":
+            pending = self._pending_requests.get(request_id)
+            if pending and isinstance(pending, dict):
+                step = content.get("step")
+                data = content.get("data")
+                results = pending["results"]
+
+                try:
+                    header, encoded = data.split(",", 1)
```
sofiia-chorna:
```suggestion
                    _header, encoded = data.split(",", 1)
```

python/jupyter/src/widget.ts
```diff
@@ -386,6 +400,159 @@ class ChemiscopeBaseView extends DOMWidgetView {
...
     }
 
+    /**
+     * Install a handler for custom messages coming from Python
+     */
+    protected _enableScreenshotMessages(): void {
+        this.model.on(
+            'msg:custom',
+            (content: ScreenshotRequest | StructureSequenceRequest, _buffers: unknown[]) => {
+                void _buffers;
```
sofiia-chorna, line 410: do we actually need to do it?
```suggestion
```

python/chemiscope/jupyter.py
```diff
@@ -73,6 +77,190 @@ def _cache_structures(self, structures):
...
+        async def _save_impl():
+            data_list = await self.get_structure_sequence(indices, settings)
+            for path, data in zip(paths, data_list, strict=True):
+                with open(path, "wb") as f:
+                    f.write(data)
+
+        return asyncio.ensure_future(_save_impl())
+
+    def _request_screenshot(self, target):
+        import time
```
sofiia-chorna: let's move import to top to other imports

python/chemiscope/jupyter.py
```diff
@@ -73,6 +77,190 @@ def _cache_structures(self, structures):
...
+        self.send({"type": "save-image", "target": target, "requestId": request_id})
+        return future
+
+    async def _save_image_to_file(self, path, target):
+        data = await self._request_screenshot(target)
+        with open(path, "wb") as f:
+            f.write(data)
+
+    async def _process_structure_sequence(self, indices, settings=None):
+        import time
```
sofiia-chorna: same comment about import

python/chemiscope/jupyter.py
```diff
@@ -73,6 +77,190 @@ def _cache_structures(self, structures):
...
+
+    def _request_screenshot(self, target):
+        import time
+
+        request_id = int(time.time() * 1000)
+        while request_id in self._pending_requests:
+            request_id += 1
+
+        loop = asyncio.get_running_loop()
+        future = loop.create_future()
```
sofiia-chorna: duplicated code from `_process_structure_sequence` function, we can move it to a separate function

python/chemiscope/jupyter.py
```diff
@@ -73,6 +77,190 @@ def _cache_structures(self, structures):
...
+
+        .. code-block:: python
+
+            data = await widget.get_structure_image()
+
+        :return: A Future that resolves to the image data as bytes.
+        """
+        return self._request_screenshot("structure")
+
+    def save_map_image(self, path):
```
sofiia-chorna, line 132: if widget mode is "structure", should we show a warning to the user? e.g. for `save_structure_image`, we throw a runtime error `"sendError('no structure in this visualizer');"`

python/chemiscope/jupyter.py
```diff
@@ -73,6 +77,190 @@ def _cache_structures(self, structures):
...
+    def _request_screenshot(self, target):
+        import time
+
+        request_id = int(time.time() * 1000)
+        while request_id in self._pending_requests:
+            request_id += 1
+
+        loop = asyncio.get_running_loop()
+        future = loop.create_future()
+        self._pending_requests[request_id] = future
```
sofiia-chorna, line 258: I am wondering if we should throw `TimeoutError` if js side never responses?

docs/src/python/jupyter.rst
```diff
@@ -62,12 +62,86 @@ the `ipywidgets documentation
     in the grid (0-based).
 
 
+Saving the widget state as a standalone dataset
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Luthaf:
```suggestion
Saving the dataset and settings as a standalone file
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

docs/src/python/jupyter.rst
```diff
@@ -62,12 +62,86 @@ the `ipywidgets documentation
...
+
 The "save" method allows exporting the current state of the widget 
 as a standalone JSON viewer that can be opened in the web app
```
Luthaf:
```suggestion
as a standalone JSON file that can be opened in the web app 
```

docs/src/python/jupyter.rst
```diff
@@ -62,12 +62,86 @@ the `ipywidgets documentation
...
+
+There are two types of methods: ``save_*`` methods, which write the image data 
+directly to a file given the path, and ``get_*`` methods, which return the raw 
+image bytes (PNG formatted).
+
+.. note::
+    These methods return ``asyncio.Future`` objects. Depending on the environment,
+    environments, you may be able to ``await`` directly in a cell to wait for the 
+    result. If ``await`` hangs in your environment, you can use 
+    catch the ``get_*`` return value, and then in a separate cell get `.result()`
```
Luthaf, line 91:
```suggestion
    result. If ``await`` hangs in your environment, you can use 
    the ``get_*`` return value, and then in a separate cell get `.result()` 
```

python/chemiscope/jupyter.py
```diff
@@ -73,6 +77,190 @@ def _cache_structures(self, structures):
...
+
+        msg = {
+            "type": "save-structure-sequence",
+            "indices": indices,
+            "requestId": request_id,
+        }
+        if settings is not None:
+            msg["settings"] = settings
+
+        self.send(msg)
```
Luthaf, line 288: Why is this a separate message instead of sending multiple `"type": "save-image"` messages and setting the indices in the middle?

ceriottm: same reason why you can't have a loop that sets the index with selected_ids and then save individual images with save_structure_image(). there is no guarantee that the widget will react to the incoming message.

python/chemiscope/jupyter.py
ceriottm: I simplified the request_id creation, it's three lines I don't think it's worth to make a function

python/chemiscope/jupyter.py
ceriottm: I fear that in the way this works we will never be able to detect this because we let it write (or check the return value of the future)

python/chemiscope/jupyter.py
ceriottm: I'm checking python side I think it's easier and better

python/jupyter/src/widget.ts
ceriottm: yes because linter.

python/chemiscope/jupyter.py
ceriottm: I already had to add error checking in separate functions because the core calls that go through futures would intercept the error

sofiia-chorna (approved):

python/chemiscope/jupyter.py
Luthaf: Even integrating with async io? Ok, fine for me. If we start growing the list of messages, it could be worth having some internal documentation about them, and what parameters are required in the message/answer.

PR #488, opened 2026-01-12
Save and load camera and range settings for 3D maps and the structure viewers

ceriottm:
This PR implements storing and restoring the 3D view of the map and structures.

### DIFF
--- a/src/structure/viewer.ts
+++ b/src/structure/viewer.ts
@@ -14,7 +14,7 @@ import { Environment, Settings, Structure } from '../dataset';
 
 import { Arrow, CustomShape, Cylinder, Ellipsoid, ShapeData, Sphere, mergeShapes } from './shapes';
 
-import { CameraState, ViewState, cameraToView, viewToCamera } from '../utils/camera';
+import { ViewState, cameraToView, viewToCamera } from '../utils/camera';
 
 import { StructureOptions } from './options';
 


src/map/map.ts
```diff
@@ -252,6 +281,8 @@ export class PropertiesMap {
     private _options: MapOptions;
     /// Button used to reset the range of color axis
     private _colorReset: HTMLButtonElement;
+    /// Saved camera state for 3D plot
+    private _cameraState: CameraState | undefined;
```
Luthaf:
```suggestion
    private _cameraState?: CameraState;
```

There is a special syntax in TypeScript for `T  | undefined` as a class/interface member

src/map/map.ts
```diff
@@ -328,15 +359,21 @@ export class PropertiesMap {
...
             (rect) => this.positionSettingsModal(rect),
-            settings,
+            optionsSettings,
             this.warnings
         );
         this._colorReset = this._options.getModalElement<HTMLButtonElement>('map-color-reset');
 
+        if (settings.camera) {
+            this._cameraState = settings.camera as unknown as CameraState;
+        }
```
Luthaf: This should do type validation instead of silently accepting user input. 

This is done by the MapOptions for all other settings, could we re-use the same code here instead of shoving it in the settings but not the options?

src/map/map.ts
```diff
@@ -1345,19 +1417,23 @@ export class PropertiesMap {
         });
 
         this._plot.on('plotly_afterplot', () => {
-            void this._afterplot();
+            void (async () => {
+                await this._afterplot();
+            })();
```
Luthaf: I don't love this (mixing async and sync code). 

Why is `_afterplot` an async function? If this is because other plotly functions are async, can we use something else than `plot.on()` that natively works with async?

src/map/map.ts
```diff
@@ -1422,7 +1541,7 @@ export class PropertiesMap {
     private _getAxisRange = (
         min: number,
         max: number,
-        axisName: string
+        _axisName: string
```
Luthaf: why?

src/map/map.ts
```diff
@@ -1433,13 +1552,28 @@ export class PropertiesMap {
...
             this.warnings.sendMessage(
-                `The inserted min and max values in ${axisName} are such that min > max!` +
+                `The inserted min and max values in ${_axisName} are such that min > max!` +
                     `The default values will be used.`
             );
         }
         return [minProvided ? min : undefined, maxProvided ? max : undefined];
     };
 
+    private _getAxisAutoRange = (min: number, max: number, _axisName: string): boolean => {
```
Luthaf:
```suggestion
    private _getAxisAutoRange = (min: number, max: number, axisName: string): boolean => {
```

src/map/map.ts
```diff
@@ -1433,13 +1552,28 @@ export class PropertiesMap {
...
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const __axisName = _axisName;
```
Luthaf: even more why?

src/map/map.ts
Luthaf: If we really need this pattern I would at least catch and log errors instead of silently casting everything to `void`

src/utils/camera.ts
```diff
@@ -0,0 +1,205 @@
...
+ * for Plotly.js and 3dmol.js viewers, including conversions between their
+ * respective camera representations.
+ *
+ * @packageDocumentation
+ * @module utils
+ */
+
+import { Vector3D, norm } from '../structure/linalg';
+
+export interface CameraState {
```
Luthaf, line 12: Since this is a new setting, it should be documented in https://chemiscope.org/docs/getting-started/json-format.html#settings-settings

src/map/map.ts
ceriottm: so, the problem is that there's no UI option for this (and I don't want to have one). are you suggesting of ignoring / checking in MapOptions.applySettings rather than handling the camera here?

src/map/map.ts
ceriottm: _Your discriminatory comments against underscores will be reported to HR.

sofiia-chorna:
"reset" button is not aligned on my side
Screenshot 2026-01-13 at 18 19 04

src/map/map.ts
ceriottm: yeah looks like there's really no need to run all of these asynchronously. afterplot needs to be asynchronous because it's calling an async plotly function though

src/map/map.ts
ceriottm: guess because it was used only for the error, but yeah can do without

src/map/map.ts
ceriottm: I looked a bit at different options and TBH this seems to me the cleanest.

Luthaf: I would at least like to have some code that validates that the correct types are passed. 

This also breaks the pattern that we had so far that every setting was handled by the options, so making this code also handled by the options would fix both. We could make it so that the values in the options are not linked to any specific HTML UI element, that would be fine with me.

If you think this is not possible, then we need to implement user input validation somewhere else!

sofiia-chorna (approved):
looks good & i did not manage to break it

ceriottm:
I'm merging as anyway we'll have to look at the input validation and option sync in one of the other PRs.

src/options.ts
```diff
@@ -283,11 +283,48 @@ export class HTMLOption<T extends OptionsType> {
     }
 }
 
+/**
+ * A setting option that is not bound to any HTML element, but can be saved and
+ * loaded with the rest of the settings.
+ */
+export class JSOption<T> {
+    /** Callback to validate the new value before propagating changes. */
+    public validate: (value: T) => void;
```
Luthaf, line 292: I see that this added the option to do validation, but is was not actually implemented. Maybe the constructor should take a `validate` callback, so users of the class are required to provide it?

PR #489, opened 2026-01-15
Adaptive resolution for 3D plots

ceriottm:
This PR completes the work on adaptive resolution making it work well also in 3D. More specifically, this PR:
* fixes some weird problems when saving camera state for plotly, and fixing parameter range detection in 3D
* implements binning in the projected space rather than in 3D (i.e. zooming in actually changes the resolution, even without changing the ranges)
* does a "coarse global binning plus local binning" to ensure that the point cloud is visible even when panning/rotating outside the range without having to wait for a redraw

### DIFF
--- a/src/structure/viewer.ts
+++ b/src/structure/viewer.ts
@@ -1593,18 +1593,43 @@ export class MoleculeViewer {
         return currentProperty;
     }
 
+    private _nonStandardElemColors = new Map<string, string>();
+
     /**
      * Get a function computing the atom color, that can be used as 3Dmol
      * `colorfunc`
      */
-    private _colorFunction(): ((atom: $3Dmol.AtomSpec) => number) | undefined {
-        if (this._properties === undefined) {
-            return undefined;
-        }
-
+    private _colorFunction(): ((atom: $3Dmol.AtomSpec) => $3Dmol.ColorSpec) | undefined {
         const property = this._options.color.property.value;
-        if (property === 'element') {
-            return undefined;
+
+        if (this._properties === undefined || property === 'element') {
+            const tab20Palette = COLOR_MAPS['tab20'];
+            return (atom: $3Dmol.AtomSpec) => {
+                if (atom.elem !== undefined) {
+                    const value = $3Dmol.elementColors.Jmol[atom.elem];
+                    if (value !== undefined) {
+                        // standard element names
+                        return value;
+                    } else {
+                        let value = this._nonStandardElemColors.get(atom.elem);
+                        if (value === undefined) {
+                            this._nonStandardElemColors.set(
+                                atom.elem,
+                                tab20Palette[
+                                    (this._nonStandardElemColors.size % (tab20Palette.length / 2)) *
+                                        2
+                                ][1]
+                            ); // each color is repeated twice
+                            value = this._nonStandardElemColors.get(atom.elem);
+                        }
+                        assert(value !== undefined);
+                        return value;
+                    }
+                } else {
+                    // missing values
+                    return 0xdddddd;
+                }
+            };
         }
 
         const transform = this._options.color.transform.value;


src/map/lod.ts
```diff
@@ -4,6 +4,79 @@
...
+ * @param yValues Array of Y coordinates
+ * @param zValues Array of Z coordinates
+ * @param camera  Current camera state (eye, center, up, zoom)
+ * @param threshLOD Maximum number of points to display (default: 50000)
+ * @returns Sorted array of indices to display
+ */
+export function computeScreenSpaceLOD(
+    xValues: number[],
+    yValues: number[],
+    zValues: number[],
```
sofiia-chorna, line 24: I am wondering if `zValues` should be optional here?

src/map/lod.ts
```diff
@@ -4,6 +4,79 @@
...
+    camera: CameraState,
+    bounds?: { x: [number, number]; y: [number, number]; z?: [number, number] },
```
sofiia-chorna, line 26: `bounds` is not annotated in the jsdoc of the function

src/map/options.ts
```diff
@@ -163,6 +163,13 @@ export class MapOptions extends OptionsGroup {
         // adaptive resolution for large datasets
         this.useLOD = new HTMLOption('boolean', true);
         this.camera = new JSOption<CameraState | undefined>(undefined);
+        this.camera.validate = (value: CameraState | undefined) => {
+            if (value) {
+                if (value.zoom <= 0) {
+                    throw Error(`zoom factor must be greater than zero, got ${value.zoom}`);
+                }
+            }
```
sofiia-chorna: we can get rid of the additional nesting
```suggestion
            if (value && value.zoom <= 0) {
                 throw Error(`zoom factor must be greater than zero, got ${value.zoom}`);
            }
```

src/map/plotly/plotly-scatter.d.ts
```diff
@@ -29,6 +36,7 @@ interface ScatterLayout {
 export interface PlotlyScatterElement extends PlotlyHTMLElement {
     _fullData: {
         type: 'scatter3d' | 'scattergl';
+        uid: string;
```
sofiia-chorna: seems to be not used anywhere
```suggestion
```

src/structure/options.ts
```diff
@@ -146,6 +146,13 @@ export class StructureOptions extends OptionsGroup {
 
         this.labelsProperty = new HTMLOption('string', 'element');
         this.camera = new JSOption<CameraState | undefined>(undefined);
+        this.camera.validate = (value: CameraState | undefined) => {
+            if (value) {
+                if (value.zoom <= 0) {
+                    throw Error(`zoom factor must be greater than zero, got ${value.zoom}`);
+                }
+            }
```
sofiia-chorna: same comment 
```suggestion
            if (value && value.zoom <= 0) {
                 throw Error(`zoom factor must be greater than zero, got ${value.zoom}`);
            }
```

src/map/lod.ts
```diff
@@ -4,6 +4,79 @@
...
+ * This is used for 3D plots where the visible density depends on camera rotation/zoom.
+ *
+ * @param xValues Array of X coordinates
+ * @param yValues Array of Y coordinates
+ * @param zValues Array of Z coordinates
+ * @param camera  Current camera state (eye, center, up, zoom)
+ * @param threshLOD Maximum number of points to display (default: 50000)
+ * @returns Sorted array of indices to display
+ */
+export function computeScreenSpaceLOD(
```
sofiia-chorna, line 21: should we check if length of values is smaller than threshLOD so maybe we can skip those computation entirely when dataset is small

src/map/map.ts
```diff
@@ -1460,6 +1477,22 @@ export class PropertiesMap {
...
+                this.setActive(this._active);
+                const data = this._selected.get(this._active);
+                if (data !== undefined) {
+                    this.activeChanged(
+                        this._active,
+                        this._indexer.fromEnvironment(data.current, this._target)
+                    );
+                }
+            }
+        }, 1000);
```
sofiia-chorna, line 1484: oh looks fragile but seems to work fine

src/map/lod.ts
ceriottm: no, because this only makes sense in 3D

src/map/lod.ts
ceriottm: this is checked in map.ts _before_ it even gets here.

src/map/map.ts
ceriottm: it is fragile and I'm not happy but I can't figure out why _fullLayout doesn't get updated otherwise, nor a more robust way to get it to work

ceriottm:
@sofiia-chorna  can you also pls check that setting the camera from jupyter works? it should go through the settings

sofiia-chorna:
for me, clicking on "reset view" make the widget blink back and forth one time
Screen Recording 2026-01-16 at 16 03 18

ceriottm:
> for me, clicking on "reset view" make the widget blink back and forth one time

Sigh. Where do you get this? I tried both chrome and firefox and I don't see this flicker-in/flicker back. Can you do a bit of console.log debugging to understand what's going on?

sofiia-chorna:
oh in browser it works fine, in vscode i have this flick, and seems even in main

ceriottm:
> oh in browser it works fine, in vscode i have this flick, and seems even in main

OK I was having it in jupyter. I think I fixed it now. We need to clean up the way relayout and afterplot interact, I think that is what adds a lot of fragility. Let's get to a state that _works_, and then maybe try separately to simplify the logic.

src/structure/options.ts
```diff
@@ -146,6 +146,10 @@ export class StructureOptions extends OptionsGroup {
 
         this.labelsProperty = new HTMLOption('string', 'element');
         this.camera = new JSOption<CameraState | undefined>(undefined);
+        // as long as the type matches, camera state is valid (zoom<is acceptable for 3dmol)
```
Luthaf, line 149: My point about input validation is that we don't know if the type matches in the first place, so we should also check that. Javascript does not care about types and will pass whatever around

src/utils/camera.ts
```diff
@@ -172,7 +174,119 @@ export function cameraToView(camera: CameraState): ViewState {
         }
     }
 
-    return [center.x, center.y, center.z, zLen, qx, qy, qz, qw];
+    return [center.x, center.y, center.z, camera.zoom, qx, qy, qz, qw];
+}
+
+/** 4x4 Matrix for 3D projection */
+export type Matrix4 = [
```
Luthaf, line 181: why not

```
Vector4 = [number, number, number, number];
Matrix4 = [Vector4, Vector4, Vector4, Vector4];
```

src/structure/options.ts
ceriottm: isn't this ensured by the typechecker? I've no idea of how to check this manually, if I check for members not being undefined eslint will complain the assert is useless no?

src/utils/camera.ts
ceriottm: the AI decided to do it this way and I decided it was not worth risking to introduce errors by manually changing it around. it's easy enough to follow. Feel strongly otherwise?

src/structure/options.ts
Luthaf: The typechecker is compile-time only, and does nothing at runtime. To bypass the `eslint` issue, the data should be passed to `validate` as something like `JsObject`: https://github.com/lab-cosmo/chemiscope/blob/0f02c674783f0cc8e1a8e3b9c3b36659ecb988a6/src/dataset.ts#L258

src/utils/camera.ts
Luthaf: I'm fine, I would at least add a comment if data is stored in row major or column major storage

src/structure/options.ts
Luthaf: Or even just `unknown`, and then the type will be refined:

```ts
function validate(value: unknown) {
    if (typeof value === 'object') {
        # value is an object here
    } else if (Array.isArrays(object) {
        # value is an array of unknown here
    }
}
```

ceriottm: ok, I added a checker for type structure. feels overkill but it should be general enough we can reuse it if needed

src/options.ts
```diff
@@ -87,6 +87,50 @@ export function optionValidator(valid: string[], name = ''): (value: string) =>
...
+            );
+        }
+
+        for (const key in template) {
+            if (Object.prototype.hasOwnProperty.call(template, key)) {
+                if (!Object.prototype.hasOwnProperty.call(value, key)) {
+                    throw Error(`missing key '${key}' in ${name}`);
+                }
+                validateObject(value[key], template[key], `${name}.${key}`);
+            }
```
Luthaf: This does not validate types of the properties, right? TBH I'm not sure this is worth doing in a generic way VS just having a `validateCameraState`, which will be easier to understand and follow by the next contributors

src/options.ts
```diff
@@ -87,6 +87,50 @@ export function optionValidator(valid: string[], name = ''): (value: string) =>
...
 
+/**
+ * Creates a validating function that checks that the value has the same
+ * properties and types as the template.
+ *
+ * @param  template  template object defining the expected structure
+ * @param  name      name of the setting for better error messages
+ * @return           a function that can be used to validate a new setting value
+ */
+export function objectValidator<T>(template: T, name = ''): (value: T) => void {
```
Luthaf: I don't think this will be able to handle optional properties? Maybe it's fine for camera stuff, but in general we have a lot of optional data

Luthaf (approved):
The validation code looks good, I trust you that the adaptative resolution changes where thoroughly tested!

PR #490, opened 2026-01-16
Allow for coloring atom with non-standard element names

GardevoirX:
Closes #446.

Previous: #465 

Now (loaded with `MDAnalysis`, non-standard element name):
image

Now (loaded with `ase`, standard element name):
image

### DIFF
--- a/docs/src/getting-started/json-format.rst
+++ b/docs/src/getting-started/json-format.rst
@@ -149,6 +149,13 @@ represented as an object with the following fields:
      - List of bonds as ``[[i, j, order], ...]`` (0-based atom indices).
      - No
      - ``[[0, 1, 1], [1, 2, 1]]``
+   * - ``data``
+     - string
+     - Path to an external structure file to load the full structure (relative to
+       the main JSON file). When present, the structure is read from that file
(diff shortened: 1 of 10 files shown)


ceriottm:
Looks good to me. Seems a sufficiently remote corner case that it does not need documentation. Just one concern to think about: if I understand correctly, colors are assigned sequentially based on how the elements appear in the atom list. If different frames have different elements, or the atom numbers change, the colors will flicker. it might make sense to make  it into a private class attribute, and maybe make it into a dictionary, say `_nonStandardElementColors`. Then here you'd just do
```
let value = $3Dmol.elementColors.Jmol[atom.elem];
if (value !== undefined) {
    // standard element names
    return value;
}
value = this._nonStandardElementColors.get(atom.elem);
if  (value !== undefined) {
    // standard element names
    return value;
} else {
this._nonStandardElementColors[this._nonStandardElementColors.length] = 
tab20Palette[this._nonStandardElementColors.length % tab20Palette.length]
}```
Or something like this. Simpler and stabler.

ceriottm (approved):
Looks good, works and we did a round of cleanup.

PR #494, opened 2026-01-20
Fix stack overflow

sofiia-chorna:
-  fixes error in 3D throwing "max stack overflow" with huge dataset
- refactor

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -270,8 +270,11 @@ export class PropertiesMap {
     private static readonly LOD_THRESHOLD = 50000;
     /// Stores the subset of point indices to display when LOD is active
     private _lodIndices: number[] | null = null;
-    /// Guard to prevent infinite recursion in afterplot loops
-    private _lodLocked = false;
+    /**
+     * Guard to skip concurrent LOD updates. When set, other callers simply return early
+     * instead of waiting
+     */
+    private _lodBusy = false;
     // Timeout id used to batch plotly afterplot events
     private _afterplotRequest: number | null = null;
 
@@ -978,10 +981,10 @@ export class PropertiesMap {
 
             // Perform async sequence while holding a LOD lock to avoid races
             void (async () => {
-                if (this._lodLocked) {
+                if (this._lodBusy) {
                     return;
                 }
-                this._lodLocked = true;
+                this._lodBusy = true;
 
                 try {
                     // Ensure traces are restyled according to current LOD
@@ -1004,7 +1007,7 @@ export class PropertiesMap {
                         this._setScaleStep(this._getBounds().z as number[], 'z');
                     }
                 } finally {
-                    this._lodLocked = false;
+                    this._lodBusy = false;
                 }
             })();
         });
@@ -1363,7 +1366,7 @@ export class PropertiesMap {
             }
 
             // don't intercept clicks while the subsampling is updating
-            if (this._lodLocked) {
+            if (this._lodBusy) {
                 return;
             }
             let environment = event.points[0].pointNumber;
@@ -1403,7 +1406,7 @@ export class PropertiesMap {
         });
 
         this._plot.on('plotly_afterplot', () => {
-            if (this._lodLocked) {
+            if (this._lodBusy) {
                 return;
             }
             if (this._afterplotRequest !== null) {
@@ -1418,7 +1421,7 @@ export class PropertiesMap {
 
         // 3D LOD: Listen to relayout to catch 3D camera changes (zoom/pan)
         this._plot.on('plotly_relayout', (event: Plotly.PlotRelayoutEvent) => {
-            if (this._lodLocked) {
+            if (this._lodBusy) {
                 return;
             }
             if (this._afterplotRequest !== null) {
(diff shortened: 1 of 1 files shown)


sofiia-chorna:
i tried to merge all features from https://github.com/lab-cosmo/chemiscope/pull/496 to this code

src/map/lod.ts
```diff
@@ -3,117 +3,111 @@
...
     }
 
-    const nPoints = xValues.length;
-    // Calculate View Matrix once
     const projections = projectPoints(xValues, yValues, zValues, camera, bounds);
 
-    // Grid resolution: we want roughly threshLOD points on screen.
-    const bins = Math.ceil(Math.sqrt(threshLOD));
-    const grid = new Map<string, number>();
+    // Bounds slightly larger to avoid popping at edges
```
Luthaf, line 41: bounds slightly larger than what?

src/map/lod.ts
```diff
@@ -3,117 +3,111 @@
...
+        if (grid[idx] === -1) {
+            grid[idx] = id;
+            result.push(id);
         }
     }
 
-    const indices = Array.from(grid.values());
-    indices.sort((a, b) => a - b);
-    return indices;
+    result.sort((a, b) => a - b);
```
Luthaf, line 77: What does the sorting accomplish here? I'm not sure I understand this bit of the code

src/map/lod.ts
```diff
@@ -3,117 +3,111 @@
...
- * @param camera  Current camera state (eye, center, up, zoom)
- * @param bounds  Optional boundaries to clip the data (Zoom level)
- * @param threshLOD Maximum number of points to display (default: 50000)
- * @returns Sorted array of indices to display
+ * @param camera Current camera state (eye, center, up, zoom)
+ * @param bounds Optional boundaries to clip the data
+ * @param maxPoints Maximum number of points to display
+ * @returns Array of indices to display
  */
 export function computeScreenSpaceLOD(
```
Luthaf, line 20: kinda unrelated question, but should we/do we ensure that selected structures from the viewer are always selected when doing LOD?

src/map/lod.ts
```diff
@@ -3,117 +3,111 @@
...
-    const indices = Array.from(grid.values());
-    indices.sort((a, b) => a - b);
-    return indices;
+    result.sort((a, b) => a - b);
+    return result;
 }
 
 /**
- * Computes the subset of points to display based on spatial grid binning.
+ * Computes LOD based on spatial grid binning. Used for 2D plots or 3D without camera.
```
Luthaf: What does 3D without camera means?

src/map/map.ts
```diff
@@ -1016,26 +976,35 @@ export class PropertiesMap {
             // LOD: Z changed, compute a first downsampling if necessary
             this._computeLOD();
 
-            // This has to run asynchronously to ensure that the z axis is scaled properly
+            // Perform async sequence while holding a LOD lock to avoid races
```
Luthaf, line 982: Is there a risk of [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issure here? If this is called twice in different contexts, could both of the calls progress through the if, and then be suspended, and then both try to acquire the lock?

I'm not sure if JS async functions can be suspended anywhere or only at explicit `await` points.

(let me know if this is not clear!)

src/map/lod.ts
ceriottm: to what would fill the view

src/map/lod.ts
ceriottm: it makes sure that points (esp. in 2D) are shown with the same z-order they'd show up without LOD. reduces "flickering" a bit and is very fast

src/map/lod.ts
ceriottm: not necessary because the selected points are shown as separate traces

src/map/lod.ts
ceriottm:
```suggestion
 * Computes LOD based on spatial grid binning. Used for 2D plots or to get a global
 * 3D subsampling that does not depend on the camera viewpoint
```

src/map/map.ts
ceriottm: Not clear to me ^_^

src/map/lod.ts
Luthaf: Worth adding a comment about this then!

src/map/lod.ts
Luthaf: How do we know 2.2 is the correct value? We assume everything is between -1 and 1?

src/map/map.ts
Luthaf: Basically, I'm wondering if we can have two threads of execution, both see `this._lodLocked == false`, continue working, and both set `this._lodLocked=true` at the same time. 

AFAIK, JS is single-threaded, so this is fine. But async throws a wrench in there, because an async function can be suspended, and there is a risk a first async call gets suspended in between the `if (this._lodLocked)` and setting the value below; then a second async call goes through and set `_lodLocked=true`, and then the first one resume: we now have two calls/threads of execution "holding" the same lock.

Luthaf: Or is this not a lock (and thus not something to help with races) and more a "we are doing this now, please skip other work"?

sofiia-chorna: there is no `await` between check and set, so i assume no other call can sneak in with the single thread?
https://github.com/lab-cosmo/chemiscope/blob/6a054563b5aecfe41fdb49878665aa31015be0d9/src/map/map.ts#L979-L986

sofiia-chorna: upon my investigation we are safe, as you wrote, it is not a blocking lock but rather a "skip-if-busy" guard. thanks for bringing this

i like this one though :))

Screenshot 2026-01-27 at 13 55 18

Luthaf: Ok, can you update the naming/comments around this, so what's happening is clear next time we look at it?

PR #496, opened 2026-01-21
Cleaned-up LOD and relayout handling

ceriottm:
This tries to minimize the amount of heavy processing that happens when multiple events are triggered by user interaction.


ceriottm:
Closed in favor of #494

PR #497, opened 2026-01-26
Handle huge files loading

sofiia-chorna:
- added DecompressionStream api
- added error handling

I started this branch from `fix-stack-overflow` (https://github.com/lab-cosmo/chemiscope/pull/494) and tested with `huge.json.gz` from the chemiscope chat (we had `RangeError: Invalid string length error`, fixed in this PR).

I tried to do with webworkers but did not manage to work for now (browser tab still crashes). So now i put a warning when the size of the file is >500 mb.

### DIFF
--- a/docs/src/getting-started/json-format.rst
+++ b/docs/src/getting-started/json-format.rst
@@ -698,13 +698,18 @@ Configures the scatter plot. Sub-keys for axes (x/y/z), color, size, etc.
      - ``{"property": "PCA[1]", "scale": "linear"}``
    * - ``color``
      - object
-     - Color config: like axis, plus ``"palette"`` option for colormap selection, and
-       ``"opacity"`` to make traces transparent.
+     - Color config: like axis, plus ``"palette"`` option for colormap selection,
+       ``"opacity"`` to make traces transparent, and ``"select"`` to filter points.
+       Unselected points are shown in gray.
        Supported palettes: ``"inferno"``, ``"magma"``, ``"plasma"``, ``"viridis"``,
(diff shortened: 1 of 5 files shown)


Luthaf:
I changed the base to #494, this will make reviewing easier! We'll have to merge the other first, but then everything should just work

app/app.ts
```diff
@@ -316,13 +351,22 @@ function stopLoading() {
...
 
     let text;
     // '1f 8b' is the magic constant starting gzip files
-    if (magic[0] == 0x1f && magic[1] == 0x8b) {
-        text = inflate(new Uint8Array(buffer), { to: 'string' });
+    if (magic[0] === 0x1f && magic[1] === 0x8b) {
+        if (typeof DecompressionStream !== 'undefined') {
+            // use DecompressionStream API (for huge files)
+            const stream = new Response(buffer).body!;
+            const decompressedStream = stream.pipeThrough(new DecompressionStream('gzip'));
```
Luthaf: how much does this help? Would it make sense to also read the initial gzip using streams?

sofiia-chorna: not magically rapid but for big files it works a bit faster. tested with datasets from chemiscope chat (e.g. `large.json.gz`, 500k structures, and `env.json.gz` ~2M point) and in the branch it is usually like 3 seconds faster

(tested with the full streaming as well)

ceriottm:
on my system at 400MB it crashes with chrome, and with firefox it doesn't crash but it keep loading forever (I think the loader crashes in the background somehow, the console doesn't show anything). I would reduce the threshold to 200MB as it's already impractical to circulate a file that big IMO.

Luthaf (approved):
Code looks good, if you are happy with the behavior we can merge!

PR #499, opened 2026-01-28
Add options to highlight a subset of the points, either by property value range, or by category.

### LINKED ISSUE #79: Enable Selective Opacity

ceriottm:
Fixes #79 

I promise this is the last feature I want to include in 1.0. It has been on the wishlist for a while and it is very useful to analyze datasets with multiple subsets, classifiers, etc.

### DIFF
--- a/src/map/options.ts
+++ b/src/map/options.ts
@@ -179,6 +179,30 @@ export class MapOptions extends OptionsGroup {
         // adaptive resolution for large datasets
         this.useLOD = new HTMLOption('boolean', true);
         this.camera = new JSOption<CameraState | undefined>(undefined);
+        // custom equality check to avoid triggering unnecessary updates
+        this.camera.equals = (a, b) => {
+            if (a === undefined && b === undefined) {
+                return true;
+            }
+
+            if (a === undefined || b === undefined) {
+                return false;
+            }
+
+            const vecEquals = (
+                v1: { x: number; y: number; z: number },
+                v2: { x: number; y: number; z: number }
+            ) => {
+                return v1.x === v2.x && v1.y === v2.y && v1.z === v2.z;
+            };
+
+            return (
+                vecEquals(a.eye, b.eye) &&
+                vecEquals(a.center, b.center) &&
+                vecEquals(a.up, b.up) &&
+                a.zoom === b.zoom
+            );
+        };
         this.camera.validate = (value: CameraState | undefined) => {
             if (value !== undefined) {
                 validateCamera(value);
--- a/src/options.ts
+++ b/src/options.ts
@@ -292,6 +292,8 @@ export class JSOption<T> {
     public validate: (value: T) => void;
     /** Additional callbacks to run whenever the setting value changes */
     public onchange: Array<(value: T, origin: OptionModificationOrigin) => void>;
+    /** Check if two values are equal. Default to strict equality */
+    public equals: (a: T, b: T) => boolean;
 
     private _value: T;
 
@@ -299,6 +301,7 @@ export class JSOption<T> {
         this._value = value;
         this.validate = () => {};
         this.onchange = [];
+        this.equals = (a, b) => a === b;
         Object.preventExtensions(this);
     }
 
@@ -313,6 +316,9 @@ export class JSOption<T> {
     public setValue(v: T, origin: OptionModificationOrigin) {
         // We can not check for equality here since T can be an object/array
         this.validate(v);
+        if (this.equals(this._value, v)) {
+            return;
+        }
         this._value = v;
         for (const callback of this.onchange) {
             callback(this._value, origin);


ceriottm:
I still have to clean up the code, but @sofiia-chorna it'd be great if you could start testing that functionally there are no problems

sofiia-chorna:
nice idea! some feedback:
`select: range`:
1) i'd pre-select the min and max values based on the data instead of leaving the fields empty
2) i'd also prevent setting max < min or min > max and raise a warning in that case (same behavior as the color property)
3) reset range button does not work for me

`select: category`: i did not manage to make it work, it is always empty for me

ceriottm:
> nice idea! some feedback: `select: range`:
> 
> 1. i'd pre-select the min and max values based on the data instead of leaving the fields empty
> 2. i'd also prevent setting max < min or min > max and raise a warning in that case (same behavior as the color property)
> 3. reset range button does not work for me
> 
> `select: category`: i did not manage to make it work, it is always empty for me

when it's empty it's automatic, as for the other axes. for categories you need a property with categorical data - try the chemical shieldings. true that reset ranges does not act on that

sofiia-chorna:
> > nice idea! some feedback: `select: range`:
> > 
> > 1. i'd pre-select the min and max values based on the data instead of leaving the fields empty
> > 2. i'd also prevent setting max < min or min > max and raise a warning in that case (same behavior as the color property)
> > 3. reset range button does not work for me
> > 
> > `select: category`: i did not manage to make it work, it is always empty for me
> 
> when it's empty it's automatic, as for the other axes. for categories you need a property with categorical data - try the chemical shieldings. true that reset ranges does not act on that

with chemical shieldings, categorical filter works fine. btw, "reset ranges" button seems to be related to the color settings above the opacity settings.

ceriottm:
ok, @sofiia-chorna I have done a lot of cleaning up, and I have made sure that unavailable options are grayed out. can you do another round of testing, and start also looking at the code quality?

src/map/map.ts
```diff
@@ -1945,22 +2131,45 @@ export class PropertiesMap {
...
+    private _restyleFull(layoutUpdate?: Partial<Layout>): void {
         const fullUpdate: Record<string, unknown> = {
             x: this._coordinates(this._options.x),
             y: this._coordinates(this._options.y),
             z: this._coordinates(this._options.z),
             'marker.color': this._colors(),
             'marker.size': this._sizes(),
             'marker.symbol': this._symbols(),
+            'marker.line.color': this._lineColors(),
+            visible: this._selectTrace(true, true, this._options.hasColors()),
```
sofiia-chorna, line 2151: should we add `marker.opacity` here too? and `customdata: this._colorValues(0)[0]` ?

docs/src/getting-started/json-format.rst
```diff
@@ -698,12 +698,18 @@ Configures the scatter plot. Sub-keys for axes (x/y/z), color, size, etc.
...
+       Unselected points are shown in gray.
        Supported palettes: ``"inferno"``, ``"magma"``, ``"plasma"``, ``"viridis"``,
        ``"cividis"``, ``"seismic"``, ``"brg"``, ``"bwr"``, ``"rwg"``,  ``"twilight
        (periodic)"``, ``"twilight dark (periodic)"``, ``"hsv (periodic)"``, ``"tab10"``,
        ``"tab20"``, ``"tab20b"``, ``"tab20bc"``. Opacity range: 1-100.
+
+       The ``select`` object supports:
+       - ``mode``: ``"all"`` (default), ``"range"``, or ``"category"``.
+       - ``min``/``max``: range values (for ``mode="range"``).
+       - ``category``: ``property/value`` string (for ``mode="category"``).
```
sofiia-chorna, line 711: fyi, it is not displayed as a bullet list, but it passes for me

Image

docs/src/getting-started/json-format.rst
```diff
@@ -698,12 +698,18 @@ Configures the scatter plot. Sub-keys for axes (x/y/z), color, size, etc.
...
      - ``{"property": "energy", "palette": "viridis", "opacity": 100}``
```
sofiia-chorna: let's add an example with filtering
```suggestion
     - ``{"property": "energy", "palette": "viridis", "opacity": 100, "select": {"mode": "range", "min": -5, "max": 0}}``
```

src/map/map.ts
```diff
@@ -1709,6 +1727,53 @@ export class PropertiesMap {
...
+        const n = this._property(this._options.x.property.value).values.length;
+        if (this._options.color.select.mode.value === 'all') {
+            return new Array<boolean>(n).fill(true);
+        }
+
+        let colors;
+        if (this._options.hasColors()) {
+            colors = this._property(this._options.color.property.value).values;
+        } else {
+            colors = new Array<number>(n).fill(0.5);
```
sofiia-chorna, line 1762: why with 0.5?

src/map/map.ts
```diff
@@ -1761,9 +1908,43 @@ export class PropertiesMap {
...
+        const mask = this._getMask();
+        if (mask.some((v) => !v)) {
+            values = values.map((v, i) => (mask[i] ? v : 'rgba(0,0,0,0)'));
+        }
+
+        // LOD: Apply filter to main trace values
+        const mainValues = trace === 0 || trace === undefined ? this._applyLOD(values) : values;
+
+        // Assume trace 1 line color is black.
+        const selected = new Array<string>(this._selected.size).fill('black');
```
sofiia-chorna, line 1941:
```suggestion
        const selected = new Array<string>(this._selected.size).fill(globalColor);
```

sofiia-chorna (commented):
i did not observe any issues after the opacity filtering change. however when I switch from 2D to 3D with LOD (eg in the chemical shielding example), I get this warning:

`Error setting state: TypeError: Cannot read properties of undefined (reading 'camera')
`

also, in this case, if I click on "reset global", it starts behaving strangely

checking on the workstation rn

src/map/map.ts
ceriottm: it has always been this way. for fixed colors, it takes the 0.5 point of the selected palette.

src/map/map.ts
ceriottm: I removed `globalColor` instead, was not really useful `'black'` is shorter and more informative

sofiia-chorna:
oh i still have this error (can double check from workstation tomorrow or can try to fix it :) )

Screenshot 2026-01-29 at 23 16 59


> however when I switch from 2D to 3D with LOD (eg in the chemical shielding example), I get this warning:
> Error setting state: TypeError: Cannot read properties of undefined (reading 'camera') 
> also, in this case, if I click on "reset global", it starts behaving strangely

ceriottm:
> oh i still have this error (can double check from workstation tomorrow or can try to fix it :) )
> 
> Screenshot 2026-01-29 at 23 16 59
> > however when I switch from 2D to 3D with LOD (eg in the chemical shielding example), I get this warning:
> > Error setting state: TypeError: Cannot read properties of undefined (reading 'camera')
> > also, in this case, if I click on "reset global", it starts behaving strangely

This is in jupyter?

sofiia-chorna:
yes, in jupyter

ceriottm:
> Error setting state: TypeError: Cannot read properties of undefined (reading 'camera')

ok try now

sofiia-chorna:
this error has disappeared! however still there is an issue after clicking "Reset global", it starts flicking, in the terminal there is a repeating warning:
N times `[Violation] 'setTimeout' handler took <N>ms`

ceriottm:
Try now.  There was a deep problem connected with the fact it could not detect when the same camera settings were set multiple times.

sofiia-chorna:
works very well! also tested with large datasets, seems good!

minor thing, there is a warning from plotly in console
`WARN: Variable 'customdata' in hovertemplate could not be found! Please verify that the template is correct. Using value: '-'.`

sofiia-chorna (approved):

ceriottm:
> works very well! also tested with large datasets, seems good!
> 
> minor thing, there is a warning from plotly in console `WARN: Variable 'customdata' in hovertemplate could not be found! Please verify that the template is correct. Using value: '-'.`

yeah, this has been around for some time, not sure I understand where it comes from.... I'll merge it, we can fix it as we polish the release

PR #505, opened 2026-01-30
Choose between grayed-out and hidden background points.

ceriottm:
Allow to hide as well as gray out the non-selected points in the map

### DIFF
--- a/src/map/options.ts
+++ b/src/map/options.ts
@@ -153,13 +153,7 @@ export class MapOptions extends OptionsGroup {
             }
         };
         this.color.select.mode.validate = optionValidator(
-            [
-                'all',
-                'range-gray',
-                'category-gray',
-                'range-hide',
-                'category-hide',
-            ],
+            ['all', 'range-gray', 'category-gray', 'range-hide', 'category-hide'],
             'selection mode'
         );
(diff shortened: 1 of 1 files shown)


src/map/options.ts
```diff
@@ -153,7 +153,15 @@ export class MapOptions extends OptionsGroup {
...
         this.color.select.mode.validate = optionValidator(
-            ['all', 'range', 'category'],
+            [
+                'all',
+                'range',
+                'range-gray',
+                'category-gray',
+                'range-hide',
+                'category-hide',
+                'category',
```
sofiia-chorna: i think we can delete `'range'` and `'category'` as they no longer used
```suggestion
                'all',
                'range-gray',
                'category-gray',
                'range-hide',
                'category-hide',
```

ceriottm:
> regarding "link points", what behavior do we want? currently, it links all points, including greyed-out ones and in "hide" mode, edges are shown only if both nodes are visible. on my side, it is OK like this
> 
> apart from that, it seems to be working fine!

Yeah we'd have to create a new trace to do that. seems really overkill I think this is good enough

sofiia-chorna (approved):

PR #508, opened 2026-02-01
Clean up large map and structure viewer files

ceriottm:
map.ts and viewer.ts had become huge AND hard to read. I tried to move out as much as I could into a utils.ts module, and reordered the private members in a more logical order. this is a noisy PR because of the many stuff moved around, but should be harmless and feature-neutral, and IMO makes these huge files a bit more readable.

### DIFF
--- a/src/structure/utils.ts
+++ b/src/structure/utils.ts
@@ -1,0 +1,178 @@
+/**
+ * @packageDocumentation
+ * @module structure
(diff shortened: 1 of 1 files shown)


sofiia-chorna (approved):
checked those reorganizations, looks good! thanks a lot @ceriottm

PR #506, opened 2026-01-31
New chemiscope JOSS paper

ceriottm:
This branch contains the material for a new JOSS submission to mark the 1.0 release of chemiscope.

PR #512, opened 2026-02-03
Allow using chemfiles structures as input

Luthaf:
The plot in https://github.com/lab-cosmo/chemiscope/pull/506 pushed me over the edge to implement this =)

~Do not merge yet, I need to add a tutorial/docs/tests/…~

### DIFF
--- a/python/chemiscope/structures/_chemfiles.py
+++ b/python/chemiscope/structures/_chemfiles.py
@@ -1,0 +1,292 @@
+import warnings
+from typing import Sequence
+
+import numpy as np
+
+from ._ase import _remove_invalid_properties
+
+
+try:
+    import chemfiles
+
+    if chemfiles.__version__ < "0.10.0" or chemfiles.__version__ > "0.11.0":
+        print(
+            "chemiscope requires chemfiles version >=0.10,<0.11; "
+            f"but version {chemfiles.__version__} is installed."
+        )
+        HAVE_CHEMFILES = False
+    else:
+        HAVE_CHEMFILES = True
+
+    from chemfiles import Frame, MemoryTrajectory, Trajectory
+
+except ImportError:
+    HAVE_CHEMFILES = False
+
+
+def _chemfiles_valid_structures(structures):
+    if not HAVE_CHEMFILES:
+        return structures, False
+
+    if isinstance(structures, (Trajectory, MemoryTrajectory)):
+        return [structures.read_step(i) for i in range(structures.nsteps)], True
+
+    elif isinstance(structures, Frame):
+        return [structures], True
+
+    elif (
+        isinstance(structures, Sequence)
+        and len(structures) > 0
+        and isinstance(structures[0], Frame)
+    ):
+        for structure in structures:
+            assert isinstance(structure, Frame)
+        return structures, True
+
+    else:
+        return structures, False
+
+
+# fmt: off
+ELEMENTS = [
+    "X", "H", "He", "Li", "Be", "B", "C", "N", "O", "F", "Ne", "Na", "Mg", "Al", "Si",
+    "P", "S", "Cl", "Ar", "K", "Ca", "Sc", "Ti", "V", "Cr", "Mn", "Fe", "Co", "Ni",
+    "Cu", "Zn", "Ga", "Ge", "As", "Se", "Br", "Kr", "Rb", "Sr", "Y", "Zr", "Nb", "Mo",
+    "Tc", "Ru", "Rh", "Pd", "Ag", "Cd", "In", "Sn", "Sb", "Te", "I", "Xe", "Cs", "Ba",
+    "La", "Ce", "Pr", "Nd", "Pm", "Sm", "Eu", "Gd", "Tb", "Dy", "Ho", "Er", "Tm", "Yb",
+    "Lu", "Hf", "Ta", "W", "Re", "Os", "Ir", "Pt", "Au", "Hg", "Tl", "Pb", "Bi", "Po",
+    "At", "Rn", "Fr", "Ra", "Ac", "Th", "Pa", "U", "Np", "Pu", "Am", "Cm", "Bk", "Cf",
+    "Es", "Fm", "Md", "No", "Lr", "Rf", "Db", "Sg", "Bh", "Hs", "Mt", "Ds", "Rg", "Cn",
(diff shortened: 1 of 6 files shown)


sofiia-chorna:
very cool 😁

Luthaf:
Do not merge yet, I found some bugs with local tests. I'll fix them later today!

Luthaf:
This is still missing tutorial and docs, but should at least work and be tested!

python/chemiscope/structures/_chemfiles.py
```diff
@@ -0,0 +1,280 @@
...
+        for name in only:
+            if name in atom_properties.keys():
+                selected[name] = atom_properties[name]
+
+    # create property in the format expected by create_input
+    atom_properties = {
+        name: {"target": "atom", "values": np.stack(value, axis=0)}
+        for name, value in selected.items()
+    }
+    _remove_invalid_properties(properties, "chemfiles")
```
sofiia-chorna: i am wondering if there we should check `atomic_properties` not `properties`? it seems we check `properties` before in L248

python/chemiscope/structures/_chemfiles.py
```diff
@@ -0,0 +1,280 @@
...
+
+    return environments
+
+
+def _chemfiles_get_structure_properties(frames):
+    # extract the set of common properties between all frames
+    all_properties = {}
+    extra = set()
+
+    for name in frames[0].list_properties():
```
sofiia-chorna, line 170: we could live with it but what if there is `len(frames) = 0`?

python/chemiscope/structures/_chemfiles.py
```diff
@@ -0,0 +1,280 @@
...
+
+import numpy as np
+
+from ._ase import _remove_invalid_properties
+
+
+try:
+    import chemfiles
+
+    if chemfiles.__version__ < "0.10.0" or chemfiles.__version__ > "0.11.0":
```
sofiia-chorna, line 12: what if no chemfiles installed at all 😁

python/chemiscope/structures/_chemfiles.py
```diff
@@ -0,0 +1,280 @@
...
+    return all_properties
+
+
+def _chemfiles_get_atom_properties(frames, environments):
+    assert environments is not None
+    # extract the set of common properties between all atoms in all frames
+    all_properties = {}
+    extra = set()
+
+    frame = frames[environments[0][0]]
```
sofiia-chorna, line 212: same here, what if environments is [] we will have out of range error. up to you if we need to handle this error ^^

python/chemiscope/structures/_chemfiles.py
```diff
@@ -0,0 +1,280 @@
...
+    data["z"] = [float(positions[i][2]) for i in range(data["size"])]
+
+    if frame.cell.shape != chemfiles.CellShape.Infinite:
+        data["cell"] = frame.cell.matrix.T.flatten().tolist()
+
+    # bonds
+    topology = frame.topology
+    if len(topology.bonds) > 0:
+        data["bonds"] = [
+            (int(bond[0]), int(bond[1]), BOND_ORDERS_TO_NUMERIC[order])
```
sofiia-chorna: from the json-reference.rst it should be a list not a tuple...

python/chemiscope/structures/_chemfiles.py
```diff
@@ -0,0 +1,280 @@
...
+    # biomolecule-specific information
+    chains = []
+    resnames = []
+    resids = []
+    hetatom = []
+    has_biomol_info = False
+    for atom_i in range(data["size"]):
+        residue = frame.topology.residue_for_atom(atom_i)
+        if residue is None:
+            continue
```
sofiia-chorna, line 127: oh i am afraid if this could lead to some inconsistencies if we skip. i mean data["size"] will not correspond to the size of `data["resids"]`

python/chemiscope/structures/_chemfiles.py
```diff
@@ -0,0 +1,280 @@
...
+        selected = atom_properties
+    else:
+        selected = {}
+        for name in only:
+            if name in atom_properties.keys():
+                selected[name] = atom_properties[name]
+
+    # create property in the format expected by create_input
+    atom_properties = {
+        name: {"target": "atom", "values": np.stack(value, axis=0)}
```
sofiia-chorna, line 277: i see in ASE we use `np.concatenate`, why `np.stack` here? also in ASE we normalize the values in `_ase_normalize_value`, just wondering if chemfiles already handles it internally?

sofiia-chorna (commented):
i like a lot the code, it is so clean and readable! i left some questions:

python/chemiscope/structures/_chemfiles.py
Luthaf: Yes, good catch!

python/chemiscope/structures/_chemfiles.py
Luthaf: This goes to the `except ImportError:` below, and set `HAVE_CHEMFILES = False`. Although this test should also lead to `HAVE_CHEMFILES=False`!

python/chemiscope/structures/_chemfiles.py
Luthaf: I think both turn into the same JSON, but I can change it!

python/chemiscope/structures/_chemfiles.py
Luthaf: chemfiles properties are a lot more constrained than ASE, they can only be number, string, boolean or 3D vector. All should be checked in the tests.

python/chemiscope/structures/_chemfiles.py
Luthaf: added a check

python/chemiscope/structures/_chemfiles.py
Luthaf: handled!

python/chemiscope/structures/_chemfiles.py
Luthaf: good point, I'll see what I can do

sofiia-chorna (approved):
thanks a lot, looks good and works well! a doc with an example would be useful but we can add it separately if that's good for you

Issue #516, opened 2026-02-10
bug: highlighted points in map do not move in the widget during panning


jakublala:
[Works on the Website — Video](https://drive.google.com/file/d/1wHjeRwwnTihwX0Z2GQQNTItpziWnNNHB/view?usp=sharing)

[Does not Work in the Widget — Video](https://drive.google.com/file/d/1kwnVW-9yCnIhCaGUREdaqpcQtHP68dXz/view?usp=sharing)

Observe the coloured/highlighted points in the map, and how they do/don't move.

ceriottm:
I think the difference is not web vs widget, but 3D vs 2D. This is because active point traces in 2D have to be made as a separate trace. It is not the prettiest, but I think it's not easy to fix. Worth looking at it though.

jakublala:
That is totally possible, you are right.

sofiia-chorna:
@jakublala can you please check the issue is fixed on your side so we can close it? 😊

jakublala:
@sofiia-chorna I tried doing it on a 2D map and still got the error. I am trying to debug it locally but am slow, sorry.

sofiia-chorna:
no worries, thanks a lot for checking!
perhaps, you have not rebuild the jupyter widget with the new changes. I usually do this to be sure i have updated code:
```
npm run build
pip install .
jupyter lab
```

ceriottm:
just `pip install -U chemiscope` it should really work now

jakublala:
Yes, you are right! It works! Thanks!

PR #517, opened 2026-02-11
Fix a bug where dragging a 3D plot and mouse-upping outside the plot area would not trigger an update

ceriottm:
Bugfix for a small annoyance, that seems connected to a plotly limitation that does not trigger a relayout event when the mousedrag is concluded outside the plot area.


ceriottm:
Also fixed #516 since it was a one-liner on top of the existing changes.

PR #520, opened 2026-02-13
Better settings for the layout of the webapp

ceriottm:
Another small papercut fix. The layout looked terrible when the window is narrow or thin, which usually never happens but I think that with this css setup it degrades more gently in those corner cases.

### DIFF
--- a/app/app.css
+++ b/app/app.css
@@ -66,12 +66,12 @@ main.container-fluid {
 
 @media screen and (min-width: 900px) {
     main.container-fluid {
-        margin: 2em 2em calc(2em + var(--header-height)) 2em;
-        width: calc(100% - 4em);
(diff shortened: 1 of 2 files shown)


sofiia-chorna (commented):
question: for me the content is now full-width, previously it was 90% with margins. is this intentional or a side effect of aspect-ratio fixes?

now
Image

before
Image

ceriottm:
> question: for me the content is now full-width, previously it was 90% with margins. is this intentional or a side effect of aspect-ratio fixes?
> 

You are right, I wanted to make the panels as large as possible, but it was a bit too much. I have added back some (small) margins

sofiia-chorna (approved):
i like it, thanks a lot

PR #521, opened 2026-02-14
Modified bio example to also demonstrate usage of chemfiles

ceriottm:
Also modified the names->element detection heuristics as chemfiles is doing some very wild guesses (e.g. CD is unlikely to be cadmium in a biomolecule)

### DIFF
--- a/python/examples/10-visualizing-biomolecules.py
+++ b/python/examples/10-visualizing-biomolecules.py
@@ -98,8 +98,8 @@ chemiscope.show(
 # Selecting atoms of interest
 # ---------------------------
 #
-# We can also use `chemfiles selections <chfl-select>`_ to filter atoms. The selection returns
-# the indices of the atoms matching the query. We then remove the atoms we
+# We can also use `chemfiles selections <chfl-select>`_ to filter atoms. The selection
+# returns the indices of the atoms matching the query. We then remove the atoms we
 # don't want from the frame.
 #
 # .. _chfl-select: https://chemfiles.org/chemfiles/latest/selections.html


python/chemiscope/structures/_chemfiles.py
```diff
@@ -91,10 +135,28 @@ def _chemfiles_to_json(frame):
...
 
-    if all_have_element:
-        data["elements"] = elements
+    if not all_have_element:
+        # If there are unknown elements we assume the element detection failed,
+        # and try a more conservative approach assuming the elements is stored
+        # in atom type names. Chemfiles uses a similar heuristic, but maps CA
+        # and CD to Ca and Cd, while in all likelihood they should be carbon atoms.
+        elements = []
+        for atom in frame.atoms:
```
GardevoirX: Why this instead of `data["names"]`?

python/chemiscope/structures/_chemfiles.py
```diff
@@ -91,10 +135,28 @@ def _chemfiles_to_json(frame):
...
+            name = atom.name
+            if not name:
+                elements.append("X")
+                continue
+            if name[0] in "HBCNOPS":
```
GardevoirX: Hmm what if the name is `Cl`?

python/chemiscope/structures/_chemfiles.py
```diff
@@ -91,10 +135,28 @@ def _chemfiles_to_json(frame):
...
+                elements.append(name[0])
+            elif len(name) >= 2 and name[0:2].capitalize() in ELEMENTS:
+                elements.append(name[0:2].capitalize())
+            elif name[0] in "FIUKV":
```
GardevoirX: Maybe better if there are some comments explaining why this is checked here but not with `if name[0] in "HBCNOPS":`

GardevoirX (changes requested):

python/chemiscope/structures/_chemfiles.py
ceriottm: Then it will guess it is carbon. If chemfiles manages to assign all species, we trust it, but if it doesn't it's a difficult guessing game (CA could be both calcium and Calpha, and usually is the latter)

python/chemiscope/structures/_chemfiles.py
ceriottm: Was to leave out Fe mostly

python/chemiscope/structures/_chemfiles.py
GardevoirX: okay maybe throw a warning for the case `all_have_element == True` would be better, just to warn the user that the element names may be wrong

ceriottm: I assumed you meant when it's false. If it finds all elements, I'd be quite optimistic that it's correct.

python/chemiscope/structures/_chemfiles.py
ceriottm: did all the one-letters

python/chemiscope/structures/_chemfiles.py
GardevoirX: Yes sorry I meant false

python/examples/10-visualizing-biomolecules.py
```diff
@@ -75,6 +77,45 @@
...
+# +++++++++++++++
+#
+# Alternatively, we can use `chemfiles <https://chemfiles.org/>`_ to read the PDB file.
+
+trajectory = chemfiles.Trajectory(f"./{pdb_id}.pdb")
+frame = trajectory.read()
+
+# Just like with MDAnalysis, we can pass the chemfiles frame directly to chemiscope.
+chemiscope.show(
+    structures=frame,
```
Luthaf: not sure if relevant, but we can also do 
```suggestion
    structures=trajectory,
```

directly here

Luthaf: Although the current example also works fine!

python/examples/10-visualizing-biomolecules.py
```diff
@@ -75,6 +77,45 @@
...
+)
+
+# %%
+#
+# Selecting atoms of interest
+# ---------------------------
+#
+# We can also use chemfiles selections to filter atoms. The selection returns
+# the indices of the atoms matching the query. We then remove the atoms we
+# don't want from the frame.
```
Luthaf:
```suggestion
# We can also use `chemfiles selections <chfl-select>`_ to filter atoms. The selection returns
# the indices of the atoms matching the query. We then remove the atoms we
# don't want from the frame.
#
# .. _chfl-select: https://chemfiles.org/chemfiles/latest/selections.html
```

PR #522, opened 2026-02-16
Add 2026 to the dates

sofiia-chorna:


docs/src/conf.py
```diff
@@ -14,7 +14,7 @@
 # -- Project information -----------------------------------------------------
 
 project = "chemiscope"
-copyright = "2021, Guillaume Fraux"
+copyright = "2020-2026, Guillaume Fraux"
```
ceriottm: Can we modify this to say "chemiscope developers"? @Luthaf ?

Luthaf: Yes, go for it! Or explicitly list people, as you prefer

sofiia-chorna: modified to "chemiscope developers"

PR #523, opened 2026-02-20
Fix some annoying bugs around enabling and disabling centering

ceriottm:
Different interactions with the viewer would leave an inconsistent state, where environments are marked as disabled but are actually enabled. This PR simplifies the logic and hopefully just make it work

### DIFF
--- a/src/structure/viewer.ts
+++ b/src/structure/viewer.ts
@@ -886,14 +886,11 @@ export class MoleculeViewer {
                 }
             } else {
                 if (this._highlighted !== undefined) {
-                    // remove the 3DMol model for highlighted environment
-                    this._viewer.removeModel(this._highlighted.model);
                     // keep information about the last highlighted point around
                     this._lastHighlighted = {
                         center: this._highlighted.center,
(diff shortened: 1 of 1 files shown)


sofiia-chorna (commented):
click on "more options" near the env settings -> disable the environments -> "center" and "more options" buttons are no longer visible even though the panel remains open. isn't that a bit strange?

Image

i'd keep the logic consistent with the other settings, just disable the buttons instead of hiding them completely, but as you wish

ceriottm:
> click on "more options" near the env settings -> disable the environments -> "center" and "more options" buttons are no longer visible even though the panel remains open. isn't that a bit strange?

ok this was a bug, should now be fixed. my, you're good at finding corner cases!

> i'd keep the logic consistent with the other settings, just disable the buttons instead of hiding them completely, but as you wish

so the issue here is that the "center"  toggle remains visible and toggled on. we want to keep the info that it's toggled, but I think it's confusing to the user who sees it's enabled but does nothing. that's why I thought it was better to just hide everything, but I'm open to change my mind if you think that this is not a good argument

PR #524, opened 2026-02-20
Headless chemiscope widget

ceriottm:
This PR introduces a new chemiscope.headless() mode that allows to run and generate figures outside of a jupyter notebook, using playwright to run a headless browser process. Finally allows to make CS-powered figures from a python script

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,4 +80,4 @@ explore = [
     "vesin >=0.3.5,<0.6",
 ]
 streamlit = ["streamlit>=1.25"]
-headless = ["playwright>=1.52", "nest_asyncio"]
+headless = ["playwright>=1.52"]


python/chemiscope/headless.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def save(self, path):
+        """
+        Save the dataset displayed by this widget as JSON to the given ``path``.
+        If ``path`` ends with ``.gz``, the file is written as gzip compressed
+        JSON string.
+
+        :param str path: where to save the dataset.
+        """
+        import gzip
```
sofiia-chorna: let's move import to top of the page

python/chemiscope/headless.py
```diff
@@ -0,0 +1,361 @@
...
+            Path(os.path.dirname(__file__)) / "sphinx" / "static" / "chemiscope.min.js"
+        )
+
+        content = None
+        if package_static.exists():
+            with open(package_static, "r") as f:
+                content = f.read()
+        else:
+            raise ImportError(
+                "Cannot locate `chemiscope.min.js`. Please the chemiscope package"
```
sofiia-chorna:
```suggestion
                "Cannot locate `chemiscope.min.js`. Please check the chemiscope package"
```

python/chemiscope/headless.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def save_map_image(self, path):
+        """
+        Save a snapshot of the map to a file.
+        """
+        data = self.get_map_image()
+        with open(path, "wb") as f:
+            f.write(data)
+
+    def save_structure_sequence(self, indices, paths, settings=None):
```
sofiia-chorna: maybe we can also then have `get_structure_sequence`? to be consistent with jupyter widget API

python/examples/11-headless.py
```diff
@@ -0,0 +1,308 @@
...
+        boxcoords="offset points",
+        arrowprops=dict(arrowstyle="->", color="black", alpha=0.5),
+        bboxprops=dict(facecolor="white", alpha=0.8, boxstyle="round,pad=0.2"),
+        frameon=True,
+        pad=0.0,
+    )
+    ax_main.add_artist(ab)
+
+plt.tight_layout()
+plt.savefig("headless_curated_plot.png", dpi=150)
```
sofiia-chorna, line 305: let's add `*.png` to the `python/examples/.gitignore`

ceriottm:
ok @sofiia-chorna   this should be ready for another round of review

python/chemiscope/headless.py
```diff
@@ -0,0 +1,418 @@
...
+            Path(os.path.dirname(__file__)) / "sphinx" / "static" / "chemiscope.min.js"
+        )
+
+        content = None
+        if package_static.exists():
+            with open(package_static, "r") as f:
+                content = f.read()
+        else:
+            raise ImportError(
+                "Cannot locate `chemiscope.min.js`. Please chech that the chemiscope"
```
sofiia-chorna:
```suggestion
                "Cannot locate `chemiscope.min.js`. Please check that the chemiscope"
```

python/chemiscope/headless.py
```diff
@@ -0,0 +1,418 @@
...
+        as ``indices``. Each element is a dictionary of structure
+        settings (e.g. ``{"spaceFilling": True}``) to apply for
+        that specific frame.
+        """
+        if settings is not None and len(indices) != len(settings):
+            raise ValueError("indices and settings must have the same length")
+
+        if self._mode == "map":
+            raise RuntimeError(
+                "Cannot save structure image: this widget is a map-only viewer."
```
sofiia-chorna: it is get method so
```suggestion
                "Cannot retrieve structure image: this widget is a map-only viewer."
```

python/tests/headless.py
```diff
@@ -0,0 +1,163 @@
...
+
+            with self.assertRaisesRegex(
+                RuntimeError,
+                "Cannot save structure image: this widget is a map-only viewer.",
+            ):
+                widget.save_structure_image("dummy.png")
+
+            with self.assertRaisesRegex(
+                RuntimeError,
+                "Cannot save structure image: this widget is a map-only viewer.",
```
sofiia-chorna:
```suggestion
                "Cannot retrieve structure image: this widget is a map-only viewer.",
```

python/chemiscope/headless.py
```diff
@@ -0,0 +1,418 @@
...
+            raise ImportError(
+                "chemiscope.headless requires the 'playwright' package. "
+                "Please install it with `pip install playwright` and "
+                "run `playwright install`."
+            )
+
+        # apply nest_asyncio if available to allow re-entering the event loop
+        # in environments like Jupyter/IPython
+        try:
+            import nest_asyncio
```
sofiia-chorna: any specific scenario were you trying to support with `nest_asyncio`? ^^

sofiia-chorna (commented):
thanks a lot for the updates after review! just wanted to clarify `nest_asyncio` and we good to go i think ^^

python/chemiscope/headless.py
ceriottm: I noticed it didn't work when running interactively from an ipython shell. the error was indicating you needed nest_asyncio to be active.

ceriottm: ok that actuallly didn't work on jupyter. seems the only way to make it run there (that also works from python) is to use the async api of playwright.

pyproject.toml
```diff
@@ -77,6 +77,7 @@ docstring-code-format = true
 explore = [
     "pet_mad>=1.4,<2.0",
     "metatomic-torch >=0.1.0,<0.2",
-    "vesin >=0.3.5,<0.4",
+    "vesin >=0.3.5,<0.6",
 ]
 streamlit = ["streamlit>=1.25"]
+headless = ["playwright>=1.52", "nest_asyncio"]
```
sofiia-chorna: we don't use it anymore
```suggestion
headless = ["playwright>=1.52"]
```

ceriottm:
Incidentally, maybe worth noting that this opens up the possibility of CI-testing regressions in the appearance of the widget, e.g. we could store reference pngs and check that they are unchanged by every commit. only concern is that this might be very noisy if small changes upstream mean that the outputs are not pixel-perfect

PR #528, opened 2026-03-02
Papercut fixes

ceriottm:
Starting a PR to fix a few annoying papercut bugs. Everyone feel free to contribute

### DIFF
--- a/python/chemiscope/headless.py
+++ b/python/chemiscope/headless.py
@@ -4,6 +4,7 @@ import base64
 import gzip
 import json
 import os
+import tempfile
 import threading
 from pathlib import Path
 
@@ -217,8 +218,6 @@ class ChemiscopeHeadless(HasTraits):
         # magnitude faster than page.evaluate(script, data) for large datasets,
         # because evaluate() serializes data through CDP which is extremely slow
         # (e.g. 300s for 50MB vs 15s with script tag).
-        import tempfile
-
         with tempfile.NamedTemporaryFile(
             mode="w", suffix=".js", delete=False, prefix="chemiscope_"
         ) as f:
--- a/python/chemiscope/input.py
+++ b/python/chemiscope/input.py
@@ -1209,7 +1209,14 @@ _VALID_KEYS = {
 }
 
 
-def _check_vector(value, name, expected_length):
+def _check_scalar(value, name, kind=""):
+    """Check that a value is a numeric scalar (int, float, or numpy numeric)."""
+    if not np.issubdtype(type(value), np.number):
+        prefix = f"{kind} shape " if kind else ""
+        raise TypeError(f"{prefix}'{name}' must be a numeric value, got {type(value)}")
+
+
+def _check_vector(value, name, expected_length, kind=""):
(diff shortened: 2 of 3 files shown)


python/chemiscope/headless.py
```diff
@@ -212,15 +212,37 @@ def _initialize_visualizer(self, data, mode):
 
         config = {key: f"chemiscope-{key}" for key in config_keys}
 
+        # Write dataset to a temp JS file and load via add_script_tag(path=...)
+        # which reads the file directly into the browser. This is orders of
+        # magnitude faster than page.evaluate(script, data) for large datasets,
+        # because evaluate() serializes data through CDP which is extremely slow
+        # (e.g. 300s for 50MB vs 15s with script tag).
+        import tempfile
```
sofiia-chorna: we can move import to the top

python/chemiscope/input.py
```diff
@@ -1177,44 +1235,32 @@ def _check_valid_shape(shape):
...
+        raise ValueError(f"unknown shape kind '{kind}'")
 
+    for parameter in parameters:
+        if parameter not in valid_keys:
+            raise ValueError(
+                f"unknown shape parameter '{parameter}' for '{kind}' shape kind"
+            )
+
+    if kind == "sphere":
         if not isinstance(parameters["radius"], float):
```
sofiia-chorna: in `_check_vector` we accept both float and int but here only float. maybe we can add support for int values too, so` {'radius': 1}` wont fail?

python/chemiscope/input.py
```diff
@@ -1162,13 +1191,42 @@ def _validate_shapes(structures, shapes):
...
+def _check_vector(value, name, expected_length):
+    """Check that a value is a numeric sequence with the expected length."""
+    if isinstance(value, list):
+        if len(value) == expected_length and all(
+            isinstance(v, (int, float)) for v in value
+        ):
+            return
+    arr = np.asarray(value).astype(np.float64, casting="safe", subok=False, copy=False)
+    if arr.shape != (expected_length,):
+        raise ValueError(f"'{name}' must be an array with {expected_length} values")
```
sofiia-chorna: we can live like this but i see that before we also included the shape kind to the error message, e.g. `"'semiaxes' must be an array with 3 values for 'ellipsoid' shape kind"`

python/chemiscope/headless.py
```diff
@@ -212,15 +212,37 @@ def _initialize_visualizer(self, data, mode):
...
+
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".js", delete=False, prefix="chemiscope_"
+        ) as f:
+            f.write("window.__chsp_data = ")
```
sofiia-chorna, line 224: i am wondering if it is safe to use just `__chsp_data` as a global variable name

ceriottm: I think it's ok, it's created and immediately destroyed

python/chemiscope/input.py
ceriottm: ok changed it to use a generic check for numerical values

PR #532, opened 2026-03-10
New shapes to make more convenient to define advanced visualizations in the structure box

ceriottm:
Add multiple cylinders and multiple spheres at once, combine many shapes in one

### DIFF
--- a/python/chemiscope/input.py
+++ b/python/chemiscope/input.py
@@ -1125,7 +1125,7 @@ def _validate_single_shape(shapes_for_key, structures):
     atom_parameters = shapes_for_key["parameters"].get("atom", None)
 
     if structure_parameters is not None:
-        if len(structure_parameters) < len(structures):
+        if len(structure_parameters) != len(structures):
             raise TypeError(
                 f"structure_parameters must be a list with {len(structures)} "
                 f"elements, got {len(structure_parameters)}"
@@ -1133,7 +1133,7 @@ def _validate_single_shape(shapes_for_key, structures):
 
     if atom_parameters is not None:
         total_atoms = sum(s["size"] for s in structures)
-        if len(atom_parameters) < total_atoms:
+        if len(atom_parameters) != total_atoms:
             raise TypeError(
                 f"atom_parameters must be a list coinciding to the atomic "
                 f"environments, got {len(atom_parameters)} elements"
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -440,25 +440,25 @@ function checkSingleShape(
     const parameters = shape.parameters;
 
     if ('structure' in parameters) {
-        const s_parameters = parameters.structure;
-        if (!Array.isArray(s_parameters)) {
+        const structureParameters = parameters.structure;
+        if (!Array.isArray(structureParameters)) {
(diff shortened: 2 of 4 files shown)


src/dataset.ts
```diff
@@ -482,57 +525,110 @@ function validateShape(kind: string, parameters: Record<string, unknown>): strin
...
+    structureSize: number,
+    atomsCount: number
+): { result: ShapeParameters; error: string } {
+    const parameters = {
+        global: shape.parameters.global,
+        structure: shape.parameters.structure,
+        atom: shape.parameters.atom,
+    };
+
+    let full_parameters = shape.parameters.global;
```
sofiia-chorna: let's have camel style variable name for consistency in js everywhere

```suggestion
    let fullParameters = shape.parameters.global;
```
same for `atom_parameters`

src/structure/shapes.ts
```diff
@@ -718,6 +761,274 @@ export class Cylinder extends Shape {
...
+                return '"bases" must have the same length as "vectors" for "cylinders" shapes';
+            }
+            for (const p of parameters.bases as unknown[]) {
+                if (!Array.isArray(p) || p.length !== 3) {
+                    return 'each entry in "bases" must be an array with 3 elements for "cylinders" shapes';
+                }
+            }
+        }
+
+        if ('radii' in parameters) {
```
sofiia-chorna, line 848: i might mistake but it seems we don't validate `colors` at all? e.g. to make sure its lengths is the same as N of "cylinders" shapes

src/structure/viewer.ts
```diff
@@ -1261,158 +1272,265 @@ export class MoleculeViewer {
...
+                                    }
 
-                                // obey explicit color specification if given,
-                                // otherwise color as the corresponding atom
-                                if (!atomShapeData.color) {
-                                    if (colorFunc) {
-                                        atomSpec.serial = i;
-                                        atomSpec.elem = elem;
-                                        atomShapeData.color = colorFunc(atomSpec);
+                                    if (currentShape.kind === 'sphere') {
```
sofiia-chorna: all the the logic (if kind == Sphere => create new Sphere(...) etc) seems almost identical to atom shape handling with only different data variable name, so maybe we can extract it all to one function?

python/chemiscope/input.py
```diff
@@ -1078,104 +1118,115 @@ def _typetransform(data, name):
...
+    """Validate a single (non-combined) shape definition against structures."""
+    base_shape = shapes_for_key["parameters"].get("global", {})
+    structure_parameters = shapes_for_key["parameters"].get("structure", None)
+    atom_parameters = shapes_for_key["parameters"].get("atom", None)
 
-    # validate type and number of element for each entries in the shapes
-    for key, shapes_for_key in shapes.items():
-        if not isinstance(key, str):
+    if structure_parameters is not None:
+        if len(structure_parameters) < len(structures):
```
sofiia-chorna: why not != here?
same for `len(atom_parameters) < total_atoms:`

ceriottm:
Thanks, @sofiia-chorna , the duplicate code in viewer.ts was a nice catch!

PR #540, opened 2026-03-20
Fix some bugs related to parameterized properties

ceriottm:
There were a bunch of bugs related to parameterized properties:
* chemiscope.show didn't have a "parameters" option
* structure and map viewer don't work 
* there was no example


sofiia-chorna (approved):
thanks!

PR #550, opened 2026-03-22
Make headless mode resilient to browser backend failure

ceriottm:
I noticed that headless chemiscope would break completely when the background browser process hangs. this makes it recover nicely (or at least try...)

### DIFF
--- a/python/chemiscope/headless.py
+++ b/python/chemiscope/headless.py
@@ -6,6 +6,7 @@ import json
 import os
 import tempfile
 import threading
+import warnings
 from pathlib import Path
 
 from traitlets import Dict, HasTraits, observe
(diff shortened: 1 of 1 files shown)


python/chemiscope/headless.py
```diff
@@ -53,17 +54,28 @@ def _run_loop(self):
...
+        if not self._running:
+            return
+        self._running = False
+        # browser/playwright may already be dead, so ignore errors
+        for coro in [self._browser.close(), self._playwright.stop()]:
+            try:
+                future = asyncio.run_coroutine_threadsafe(coro, self._loop)
+                future.result(timeout=5)
+            except Exception:
+                pass
```
sofiia-chorna: currently the exceptions in `stop()` and `close()` fail silently, i am wondering if we should probably add some `warnings.warn` ?

python/chemiscope/headless.py
```diff
@@ -113,9 +126,15 @@ def __init__(self, data, mode="default", width=None, height=None, **kwargs):
...
+        # Get a shared server, or create one. If the browser has crashed
+        # (common in long-running Jupyter sessions), retry with a fresh server.
         self._server = _get_server()
-        self._page = self._server.new_page(device_scale_factor=1)
+        try:
+            self._page = self._server.new_page(device_scale_factor=1)
+        except Exception:
+            _close_server()
+            self._server = _get_server()
+            self._page = self._server.new_page(device_scale_factor=1)
```
sofiia-chorna, line 145: what if this second retry fails as well? maybe we can wrap it in try except here as well 😁

sofiia-chorna (commented):
works fine! some comments:

python/chemiscope/headless.py
ceriottm: yeah and raise an error then I guess

ceriottm: ok I warn at the first failure, but if it fails again we let it raise its own error.

python/chemiscope/headless.py
ceriottm: done

sofiia-chorna (approved):
thank you very much!

Issue #556, opened 2026-04-27
Improve the plot rendering ability of the stand-alone online viewer


e-deutsch:
When using the online viewer, if the number of points >~50k on the lefthand side then not all points are loaded or rendered, reducing the utility of the tool and requiring subsampling which can be a hassle. In 2D, many points above this regime only render when zooming in. In 3D, some points are extremely hard to render and view due as well to the unintuitive way that Chemiscope zooms in 3D.

I was wondering if this was something that could be improved or if it is truly just a limitation of the backend. An improvement on this front would greatly increase the utility of the web version and make it comparable in utility to that of something like the Jupyter widget which does not have this problem. Currently I lean towards using the widget because it has no rendering issues, but the online viewer is significantly more convenient.

sofiia-chorna:
hello @e-deutsch! thank you very much for opening the issue.

what you are seeing is the work of Level-of-Detail feature. by design, when a dataset has 50k+ points, chemiscope subsamples them on a grid so that the map stays interactive when loading a huge dataset.

if you'd like to render every point at once and accept the rendering performance cost, you can disable it under the map settings => toggle `adaptive resolution` off:

<img width="442" height="418" alt="Image" src="https://github.com/user-attachments/assets/900d53c9-8dcd-4d05-a9c4-8b1485280576" />


regarding the widget, it uses the same LoD code path, so it should behave the same way above 50k points. could you confirm which chemiscope version you're using in the notebook? if it's an older release, that would explain the difference

e-deutsch:
Thank you greatly for the reply, I was not aware of that toggle on the web version and it works well. I was using an older version in Python for the Jupyter notebook widget (Chemiscope 0.9.0) that did not have that toggle so I never noticed it... but it also did not have this issue and worked very well with the entire dataset. The Python documentation for Chemiscope does not mention anything about adaptive resolution or setting the default adaptive resolution to off, which may be part of why I missed it.

After upgrading to the newest version, the 50k+ dataset Chemiscope looks the same as the web version but has very noticeable performance issues in the widget compared to 0.9.0. Could just be my setup that is making it perform poorly in the widget, but I thought I would mention it.

sofiia-chorna:
interesting, we would need to investigate that. thanks for reporting!
you're right that our docs should cover LOD better, the update is coming 😊

PR #557, opened 2026-04-28
Add LOD to docs

sofiia-chorna:
added mentioning of the LOD in the docs following https://github.com/lab-cosmo/chemiscope/issues/556. please let me know if you have better idea where we should mention it


sofiia-chorna:
the LOD performance investigation will potentially be in the separate PR

ceriottm (approved):

PR #558, opened 2026-05-18
Use best-min-hash strategy to pick structures in the LOD routine

ceriottm:
Found it, fixed it.


sofiia-chorna (approved):
works good for me, thanks for the fix!

PR #559, opened 2026-05-21
Make chemiscope shapes obey alpha specifications

ceriottm:
There are limitations due to upstream limitations, but this allows to an extent support for transparent shapes in CS structure viewer

### DIFF
--- a/python/examples/12-parameterized-properties.py
+++ b/python/examples/12-parameterized-properties.py
@@ -33,8 +33,7 @@ rmax = 3.5
 nbins = 50
 rdf_values = []
 for structure in structures:
-    ana = Analysis(structure)
-    rdf = ana.get_rdf(rmax=rmax, nbins=nbins, return_dists=False)[0]
+    rdf = get_rdf(structure, rmax=rmax, nbins=nbins, no_dists=True)
     rdf_values.append(rdf)
 
 # The r grid corresponding to the RDF bins
--- a/python/examples/12-parameterized-properties.py
+++ b/python/examples/12-parameterized-properties.py
@@ -1,0 +1,72 @@
+"""
+Parameterized properties
+========================
+
+This example demonstrates the use of parameterized (multidimensional) properties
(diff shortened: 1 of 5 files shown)


python/examples/12-parameterized-properties.py
```diff
@@ -0,0 +1,72 @@
...
+We compute the radial distribution function (RDF) for each structure and display
+it as a parameterized property alongside the structure viewer.
+"""
+
+# %%
+#
+
+import ase.io
+import numpy as np
+from ase.geometry.analysis import Analysis
```
sofiia-chorna: there is a small warning from the ase in the docs
```
/Users/sofiia/cosmo/chemiscope/python/examples/12-parameterized-properties.py:37: FutureWarning: Use `ase.geometry.rdf.get_rdf` instead.
  rdf = ana.get_rdf(rmax=rmax, nbins=nbins, return_dists=False)[0]
```
we can clean up it with `get_rdf`
```suggestion
from ase.geometry.rdf import get_rdf
```

python/examples/12-parameterized-properties.py
```diff
@@ -0,0 +1,72 @@
...
+# %%
+#
+# Compute the RDF for each structure using ASE
+
+rmax = 3.5
+nbins = 50
+rdf_values = []
+for structure in structures:
+    ana = Analysis(structure)
+    rdf = ana.get_rdf(rmax=rmax, nbins=nbins, return_dists=False)[0]
```
sofiia-chorna:
```suggestion
    rdf = get_rdf(structure, rmax=rmax, nbins=nbins, no_dists=True)
```

sofiia-chorna (commented):
thanks a lot for the implementation & parameterized properties docs.

we could add a small example (like, just one line change in the `05_shapes` example to use an `alpha` for one of the shapes, maybe as an object for `color` property) if you wish! otherwise good for me!

sofiia-chorna (approved):

PR #561, opened 2026-05-28
implement streaming dataset loading in the app

sofiia-chorna:
it allows to load the datasets with more than 1M points by adding a streaming loader that parses the top-level json straight to the decompression stream without ever holding the whole text


sofiia-chorna:
should be working without "hang" at the end and faster. i could add more comments if necessary

ceriottm (approved):
I did a simplification pass, moving as much code as possible into streaming.ts. 
It works well and is pretty fast, so I'll approve and merge!

sofiia-chorna:
thank you 😊

PR #562, opened 2026-06-03
Prioritize selected points when doing LOD to avoid artifacts

ceriottm:
When the selection is very small, most of the selected points will be dropped by LOD, creating a visual artifact when using range or class based masking.

### DIFF
--- a/src/map/lod.ts
+++ b/src/map/lod.ts
@@ -89,9 +89,11 @@ export function computeScreenSpaceLOD(
         const idx = ui + vi * binsX;
         // Foreground ids (priorityMask[id] === true) get the high bit cleared,
         // background ids get it set, so any foreground id always beats any
-        // background id competing for the same bin.
+        // background id competing for the same bin. The `>>> 0` keeps the
+        // result an unsigned 32-bit value so it compares correctly against
+        // bestHash, which is read from a Uint32Array.
         const baseH = Math.imul(id, 2654435761) >>> 1;
-        const h = priorityMask && !priorityMask[id] ? baseH | 0x80000000 : baseH;
+        const h = priorityMask && !priorityMask[id] ? (baseH | 0x80000000) >>> 0 : baseH;
         if (h < bestHash[idx]) {
             bestHash[idx] = h;
             grid[idx] = id;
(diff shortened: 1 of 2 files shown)


src/map/lod.ts
```diff
@@ -206,7 +218,11 @@ export function computeLODIndices(
             idx += zi * bins * bins;
         }
 
-        const h = Math.imul(id, 2654435761) >>> 0;
+        // Foreground ids (priorityMask[id] === true) get the high bit cleared,
+        // background ids get it set, so any foreground id always beats any
+        // background id competing for the same bin.
+        const baseH = Math.imul(id, 2654435761) >>> 1;
+        const h = priorityMask && !priorityMask[id] ? baseH | 0x80000000 : baseH;
```
sofiia-chorna: as i understand you want to have the huge ids for background points so they are not selected but `baseH | 0x80000000` will actually give the smallest number (eg see `100 | 0x80000000` in browser js console, it will give -2147483548). so background points can end up in the bin (not easy to catch it though, maybe iteration order dependent). to be sure, i would do:
```suggestion
        const h = priorityMask && !priorityMask[id] ? (baseH | 0x80000000) >>> 0 : baseH;
```

src/map/lod.ts
```diff
@@ -83,7 +87,11 @@ export function computeScreenSpaceLOD(
         }
 
         const idx = ui + vi * binsX;
-        const h = Math.imul(id, 2654435761) >>> 0;
+        // Foreground ids (priorityMask[id] === true) get the high bit cleared,
+        // background ids get it set, so any foreground id always beats any
+        // background id competing for the same bin.
+        const baseH = Math.imul(id, 2654435761) >>> 1;
+        const h = priorityMask && !priorityMask[id] ? baseH | 0x80000000 : baseH;
```
sofiia-chorna: same suggestion

```suggestion
        const h = priorityMask && !priorityMask[id] ? (baseH | 0x80000000) >>> 0 : baseH;
```

sofiia-chorna (commented):
thanks a lot for fixing, i added a suggestion

sofiia-chorna (approved):

PR #563, opened 2026-06-06
Fix a bunch of problems related to the partial updates of the selction

ceriottm:
Color scale and LOD changes were not reflected in the visualization following changes of the subselection

### DIFF
--- a/src/map/map.ts
+++ b/src/map/map.ts
@@ -888,6 +888,15 @@ export class PropertiesMap {
             this._options.color.min.value = min;
             this._options.color.max.value = max;
             this._setScaleStep([min, max], 'color');
+
+            // fill the selection range with the data bounds, unless it was
+            // restored from saved settings
+            if (
+                isNaN(this._options.color.select.min.value) &&
+                isNaN(this._options.color.select.max.value)
+            ) {
+                this._resetSelectionRange();
+            }
         } else {
             this._options.color.min.value = 0;
             this._options.color.max.value = 0;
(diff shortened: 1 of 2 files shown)


src/map/map.ts
```diff
@@ -1084,8 +1103,18 @@ export class PropertiesMap {
...
-        const updateColors = () => {
-            this._computeLOD();
+        const updateColors = (_value: number | string, origin: OptionModificationOrigin) => {
+            // Only react to user edits. Programmatic ('JS') changes come from
+            // repopulating the range bounds (_resetSelectionRange) or from
+            // applySettings, which render on their own; this mirrors the axis
+            // `rangeChange` / camera handlers and avoids redundant recomputes.
+            if (origin === 'JS') {
+                return;
+            }
```
sofiia-chorna: this seems to break changing the selection in jupyter for an already displayed map, eg:
```

widget = chemiscope.show_input("dataset.json", mode="map")
widget

widget.settings = {"map": {"color": {"select": {"mode": "range-gray", "min": 85, "max": 90}}}} 
<= the map above is not updated
```

affects the "select" settings

sofiia-chorna (commented):
thanks for spotting the issue

src/map/map.ts
ceriottm: Thanks for the careful check. Turns out that also the x, y, z axes updates had been broken since forever. Tried to fix all of them.

sofiia-chorna (approved):
works well for me, thanks a lot!

PR #579, opened 2026-06-19
Anywidget rewrite of the chemiscope widget

ceriottm:
TBD whether this should be cleaned and merged, or just used for a more manual rewrite inspired by it


ceriottm:
OK I think this is ready for review. It seems pretty clean to me and also fixes a couple of long-standing annoying quirks, like the map not resizing when you open the side-bar in jupyter-lab

python/chemiscope/widget.py
```diff
@@ -0,0 +1,829 @@
...
+        return structure_refs
+
+    def get_map_image(self):
+        """
+        Request a snapshot of the map. Returns a Future that resolves to the
+        image data (PNG formatted).
+
+        This method is asynchronous. In a notebook, you should ``await``
+        it to get the data - typically in a different cell after the widget
+        is displayed, to let the JS side time to process the
```
sofiia-chorna: truncated description

python/chemiscope/widget.py
```diff
@@ -0,0 +1,829 @@
...
+             "target": "map" | "structure"}
+
+        * "save-structure-sequence": Initiated on the Python side, used to
+            generate snapshots for several atomic configurations.
+            Should provide indices (in the same format as for `selected_ids`)
+            and optionally structure settings as a list of dicts.
+
+            {"type": "save-structure-sequence",
+             "requestId": int,
+             "indices": [{"structure": id1[, "atom": id2]}, ...],
```
sofiia-chorna:
```suggestion
             "indices": [int | {"structure": id1[, "atom": id2]}, ...],
```

sofiia-chorna (commented):
thanks for implementation!

checked in colab, mariso, jupyter and vscode

noticed that in vscode creating and displaying a widget in the same cell appears as a blank cell for me:
```
widget = chemiscope.show(frames, properties=properties)
widget
```
in the console:
```
  SyntaxError: Unexpected end of JSON input
      at parseJsonWithNaN
      at ChemiscopeView.render
  render() does parseJsonWithNaN(this.model.get('value')) as a one-shot read
```

the rest seems to work well!

sofiia-chorna (approved):
thanks a lot! works fine as far as i checked, merging

PR #582, opened 2026-06-28
Add an example in google colab

ceriottm:
One-click demo of the interactive notebook widget


sofiia-chorna (approved):
nice, thanks for adding!

PR #583, opened 2026-07-06
add trunctation of long legend name

sofiia-chorna:
truncates long names in the legend names like: 
long-legend-screenshot

+ adds a tooltip to see the full name

dataset to test: [test-legend-truncation.json](https://github.com/user-attachments/files/29716983/test-legend-truncation.json)


ceriottm (approved):
Seems to work, and I agree this is good sanity check. We could also take the opportunity to set the threshold for categorical data to 64 rather than 50

ceriottm (approved):

PR #586, opened 2026-07-07
Raise an error if a metatomic model request custom inputs 

Luthaf:
These are not supported in chemiscope at this point.


sofiia-chorna (approved):
thanks a lot!

PR #588, opened 2026-07-13
fix draggable windows from going outside of the screen

### LINKED ISSUE #587: Visualiser setting window can be dragged off-screen and cannot be recovered
I've noticed that the visualiser setting window can be dragged outside the browser page. The same behaviour was observed for Jupyter widgets as well.

The dialogue could only be dragged by its title bar, and re-opening it would keep its previous position. Therefore, once the title bar is accidentally moved outside the visible viewport, I see no simple way to move it back other than refreshing the page (web) or reloading the widget (Jupyter). 

https://github.com/user-attachments/assets/227622ea-9323-4734-babe-1e681c5af745

sofiia-chorna:
closes #587

### DIFF
--- a/src/map/options.ts
+++ b/src/map/options.ts
@@ -10,7 +10,14 @@ import Modal from '../modal';
 import { Settings } from '../dataset';
 import { HTMLOption, JSOption, OptionsGroup } from '../options';
 import { optionValidator } from '../options';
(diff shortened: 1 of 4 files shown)


ceriottm (approved):
Looks good and seems to work

PR #597, opened 2026-09-22
Added a checker and an error message when WebGL is disabled

ceriottm:
Otherwise the widget crashes with an obscure error

Also fixes an annoying bug with recent webpack


sofiia-chorna (approved):
thanks a lot for the quick fix

PR #599, opened 2026-09-23
fix the symbols overlapping the colorbar

sofiia-chorna:
when we have many labels in symbols, the legend runs over the colorbar (screenshot). this PR limits the height of the symbol legend

from materials cloud:

Screenshot 2026-09-23 at 13 50 14

with the fix:

Screenshot 2026-09-23 at 13 50 32


src/map/map.ts
```diff
@@ -2059,12 +2070,26 @@ export class PropertiesMap {
...
+        const margins = DEFAULT_LAYOUT.margin.t + DEFAULT_LAYOUT.margin.b;
+        const plotHeight = this._plot.clientHeight - margins;
+
+        // no height yet, the resize observer will recompute this
+        if (count === 0 || plotHeight <= 0) {
             return 1;
         }
-        const LEGEND_ITEM_HEIGH = 0.045;
+
+        const LEGEND_ITEM_HEIGHT_PX = 19;
```
EricBoittier: I would usually watch out for magic numbers, but i'll check on mobile after merge

EricBoittier (approved):
Overall makes sense, and it's tested so I can merge. There are some "magic numbers" that might matter if the plots become very large or small, but in normal browsers this seems to work.

PR #600, opened 2026-09-23
remove mace-torch from the docs requirements

sofiia-chorna:
the proposition is the remove mace-torch from dependencies because we don't run the example in ci anyways
but the dependency can fail ci jobs, see https://github.com/lab-cosmo/chemiscope/actions/runs/35849801263/job/107144512983


docs/requirements.txt
```diff
@@ -13,7 +13,6 @@ MDAnalysis
 MDAnalysisTests
 
 # required for the examples of chemiscope.explore
-mace-torch
 scikit-learn
```
Luthaf, line 16: do we still need this one?

sofiia-chorna: we use `from sklearn.decomposition import PCA` in explore example

Luthaf (approved):

PR #601, opened 2026-09-23
fix usage of string property with the colorbar + symbols

sofiia-chorna:
when coloring by a string property, the colorbar is label with numerical values, which is fixed in this PR. when the same property is used for the symbols and the colors, the legend markers get the colors instead of showing colorbar

before
Screenshot 2026-09-23 at 16 00 52

after 
Screenshot 2026-09-23 at 16 00 34

to be merged after https://github.com/lab-cosmo/chemiscope/pull/599


EricBoittier (approved):
LGTM, also experienced similar things in Matplotlib

PR #602, opened 2026-09-23
update scatter3d to accept additional markers

sofiia-chorna:
in 3d we had an issue that some markers were displayed as circles:
Screenshot 2026-09-23 at 17 06 47

after the fix:
Screenshot 2026-09-23 at 17 07 03


EricBoittier (approved):
Nice

